PR #1, opened 2021-03-10
Basic tutorial notebook extended

leriomaggio:
## Description
This PR extends and complements the basic tutorial on **Apache Beam** by integrating very thorough explanations and references to the preparatory execution framework for the exercises, as well as to Apache Beam programming model and main design components. 

The notebook has been optimised for Google Colab, making extensive use of **forms** to hide tedious and irrelevant details (mostly related to the setup, and to hide _hints_ and _solutions_ for the exercise parts.

Conversely, all the coding bits which are considered relevant to the learning journey into `apache_beam` features have been intentionally left visible.

A few list of the most relevant features that have been included in the notebook:

- The `Netflix-Prize` dataset is now directly downloaded from **Kaggle** using `kaggle` official Python API
- The default size of the dataset is fixed to `10K` lines, but it could be easily customised
- Data objects and custom `PTransformers` have been slightly optimised (in terms of Python code) as well as made compliant with Python `multiprocessing` execution environment
- This particularly affects the use of _custom_ `dataclass` objects (saved into a separate Python module) and corresponding `beam.coders.Coder` implementation
- Execution of the `beam.Pipeline` is configure to exploit (as **default**) the `multiprocessing` Python environment to leverage on the multiple cores - if running this notebook on a larger version of the dataset and/or in an environment with more than `2 CPUs` (as in default colab).
- A few changes in the formatting in the text of exercises has been applied;
- The code in the provided solutions for the exercises has been adapted to the new _execution framework_ setup in the first part of the notebook.
- Very thorough documentation reference and quite detailed explanation of the analysed dataset, and of all features and component used throughout the notebook.
- 

## Affected Dependencies
None - all the required packages are installed automatically in the notebook. This includes `apache_beam` and `kaggle` (which is already available in Colab, but needs configuring).

## How has this been tested?

- The notebook has been tested on Google Colab, as well as on local Jupyter notebook server on a MacBook Pro laptop with 8 cores. 
- A few performance tests have been also carried out (on both the laptop and in colab) to compare the amount of time required by the multiple supported execution modes on increasing dataset sizes (from `10K` to `10M` lines). Reproducing those tests is quite easy considering the fast dataset size selection as dropdown list.


dvadym:
Wow, thanks for contributing Valerio! I've quickly checked, but I don't have much time now to make thorough review. I'll return back to the review earlier next week.

leriomaggio:
> Wow, thanks for contributing Valerio! I've quickly checked, but I don't have much time now to make thorough review. I'll return back to the review earlier next week.

Hi @dvadym thanks a lot for the feedback: glad you appreciated. 

Following up from a quick chat that I had on slack with @chinmayshah99 I am also going to share here a few more details about some performance tests I ran on my laptop (_MacBook Pro_ ,`8` cores) on increasing sizes of the datasets (and different  `direct_running_mode`).

On Colab, on the other hand, the different running mode do not make any difference, as the default VM is offering **only** `2` cores (so no _real_ gain by parallelisation of tasks). 

However, I worked on the tutorial assuming that the notebook could be executed either on Colab or in a Jupyter session.
Besides, I personally think it is a good bit to include in the tutorial for a framework like Apache Beam (at least, it was fun for me to explore a little bit the internals and the execution framework, while working on the PR) :)

Looking forward to receiving your feedbacks 🙌

leriomaggio:
As promised, attached to this PR, I am also sharing the results of a few experiments I tried on my laptop by running the notebook with the three `direct_running_mode` configurations, and multiple dataset sizes. 

##### Premise 

Those experiments were originally motivated by my intention to find the best combination of those parameters that led to _reasonable_ running time on Colab. However, as already pointed out in my previous comment, including the support for different `direct_running_mode` on Colab is quite pointless as the (_very limited_) `2 cores` on the default VM instance don't make appreciate any difference, nor gain in performance. 

However, the scenario is completely different when running the notebook in a (_slightly_)  better-equipped computing environment.  Gathered results will allow to derive a quite interesting pattern IMHO, as they also **motivate** my re-implementation of the `run_pipeline` function to allow for customisations.

##### Experiments

**Case 1**: `1M lines` Dataset (`~1%` of total size)

In the following picture, the different running times for the first `count_all_views` function executed with the three `direct_running_mode` configurations

1M lines

This very first experiment shows some preliminary gain in performance of the `multiprocessing`-based approach vs the `single-process` (in memory) approach.   

Interestingly, the `multi_threading` mode takes even more than `in_memory` - and that makes total sense as there is presumably extra over-head in communication of the multiple threads.

 **Case 2**: `10M lines` Dataset (`~10%` of total size)

With a dataset size of `10M` lines the scenario changed a bit, observing another interesting effect of the multiple execution modes. 

10M lines

The leap in performance of the two execution mode becomes substantially more evident with a dataset size that is `10x` bigger. 

However, in the case of `multi-threaded` case: 
multi-threading error

The execution did not complete successfully, with an error message in the `grpc` backend as "too many pings".
 I tried to dig a little bit in the issue and the only thing I could find was a (_closed_) issue on `grpc` [#2444](https://github.com/dgraph-io/dgraph/issues/2444) specifically mentioning python client and `grpc 1.8`. In my environment, I am currently using `grpc 1.36.1`

To my understanding, the issue is generated by some processing _not yet completed_ which makes the `worker threads` going in `timeout`. 

In more details:  for what I understood about the general execution framework in Apache Beam, the main worker process always spawns auxiliary threads (i.e. `pthreads` at C-level via `grpc`) to handle parallel (I/O?) operations on the parallel collection(s). And this is still true, regardless of the selected `direct_running_mode`.

Digging a little bit into the `PipelineOptions`, I've discovered a `job-server-timeout` which (supposedly) controls the timeout of those `job` worker threads - set to `60 secs` by default. 
To further corroborate this hypothesis, it has to be said that the same issue also happens with the `multi_processing` running mode. In fact, I had to change the default of this parameter to `65,536`, to circumvent this issue from happening  with `multi_processing` in the former `1M` lines dataset. 

In this particular case (`10M-lines` dataset) that value was still enough for `multi_processing` and `in_memory` but not for `multi_threading` for which I had to increase that value by `2/4x` to complete the execution (see below)

10M lines multi_threading

**AGAIN** (as expected) `multi_threading` was slightly worse than `in_memory` since Python doesn't like CPU-bound threads, and extra time is wasted in communication.

##### Considerations and Take-Away Messages

After these two experiments, and looking at the performance gain with the multiple configurations, I think it would be fair to conclude that execution time scales up linearly w.r.t. the dataset sizes. So in a `multi_processing` fashion:
```
  1M ==> 11 s
 10M ==> 1  m
100M ==> 10 m
```
whereas with `multi-threading` the whole completion time would summ up to `~1 h` in the `100M` lines (almost full dataset).

dvadym:
Sorry, I haven't had yet time

leriomaggio:
> Sorry, I haven't had yet time

No problem at all, I totally understand. 

I have been meaning to issue another PR (as a follow up on this one) to align the second notebook of the tutorial, but I could not finish that either.

Maybe it would be also useful to get the feedbacks on this one first, and then finishing the other notebook - anyway :)

dvadym:
Yeah, I think it's definetely worth to review the 1st Colab first and then to make changes in 2nd Colab. 

Thanks again for contributing!

docs/tutorial_1/1_beam_introduction.ipynb
```diff
@@ -1,876 +1,1681 @@
...
+    "---"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## The Execution Framework\n",
+    "\n",
+    "In this section we will define the core main components that will be used throughout the exercises. These components will be based on **Apache Beam**, which consitutes the reference computational framework, and provides the building blocks to define our data workflows.  "
```
dvadym, line 462: typo

docs/tutorial_1/1_beam_introduction.ipynb
```diff
@@ -1,876 +1,1681 @@
...
+    "5. Open this file, and include this data into the form below:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#@title Setup `kaggle` API credentials { display-mode: \"form\" }\n",
```
dvadym, line 190: Please combine this cell with cell "Authenticate on Kaggle via APIs" (it's better to have less cells for the user to run)

docs/tutorial_1/1_beam_introduction.ipynb
```diff
@@ -1,876 +1,1681 @@
...
+    "kaggle.api.authenticate()"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#@title **Search** on Kaggle for available Netflix-Prize datasets { display-mode: \"form\" }\n",
```
dvadym, line 254: Maybe put this cell and the following explanation in hidden section, with name something like "Searching dataset on Kaggle (optional)"

docs/tutorial_1/1_beam_introduction.ipynb
```diff
@@ -1,876 +1,1681 @@
...
+   "metadata": {},
+   "source": [
+    "---"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## The Execution Framework\n",
```
dvadym, line 460: Thanks, I like this section!

docs/tutorial_1/1_beam_introduction.ipynb
```diff
@@ -1,876 +1,1681 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "From these statistics, we can immediately notice that each data file roughly contains the same number of lines, which means that the amout of time required to process each one of them will be more or less the same.\n",
+    "\n",
+    "However, even just limiting our analysis to one single file (i.e. `>20M` lines) might be prohibitive, and indeed beyond of the scope of this tutorial. \n",
+    "\n",
+    "The **main goal** of the tutorial is to demostrate the ability of `apache_beam` to scale up the computation on large datasets. For real scalability however, we might also need decent computation capacity (_and we generally don't, if running this notebook on our laptop_).\n",
```
dvadym, line 369: It's hard to demonstrate how to scale up computations on 1 machine :). I'd recommend to remove this sentence

docs/tutorial_1/1_beam_introduction.ipynb
```diff
@@ -1,876 +1,1681 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "##### Python Type Safety and Custom objects\n",
+    "\n",
+    "The Apache Beam SDK for Python uses type hints during pipeline construction and runtime to try to emulate the correctness guarantees achieved by true static typing. Additionally, using type hints lays some groundwork that allows the backend service to perform efficient type deduction and registration of `Coder` objects.\n",
+    "\n",
+    "Since in this tutorial we want to handle our own custom **data object**, we also need to implement our custom (_yet simple_) `beam.coders.Coder` to allow for a correct object _serialisation_.\n",
+    "\n",
+    "**Note**: This step is particularly crucial during the execution (_in parallel_) of the Pipeline, to avoid `apache_beam` to fallback to the default `Serializer` (i.e. `Coder`) based in `pickle`.\n",
```
dvadym, line 575: just wondering, does Pickle less effective or unsafe?

dvadym, line 575: just wondering, does Pickle less effective or unsafe?

docs/tutorial_1/1_beam_introduction.ipynb
```diff
@@ -1,876 +1,1681 @@
...
+    "```\n",
+    "\n",
+    "A more thorough description of each of those files is available in the [`README`](netflix_prize_dataset/data/README) file. "
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "#### Preparing our dataset \n",
```
dvadym, line 334: My feeling there is too much information for the user who sees this dataset for the first time. What do you think about put it in a hidden by default section with name something like:
"Details on preparing our dataset (optional)"
and to have 1-2 cell to run outside of this cell in order that the user has shortcut ?

docs/tutorial_1/1_beam_introduction.ipynb
```diff
@@ -1,876 +1,1681 @@
...
+    "## The Execution Framework\n",
+    "\n",
+    "In this section we will define the core main components that will be used throughout the exercises. These components will be based on **Apache Beam**, which consitutes the reference computational framework, and provides the building blocks to define our data workflows.  "
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### `apache_beam` in a Nutshell\n",
```
dvadym, line 469: Thanks, nice explanation!

docs/tutorial_1/1_beam_introduction.ipynb
```diff
@@ -1,876 +1,1681 @@
...
+    "More information on **Python Type Safety** in Apache Beam pipelines is available in the official [documentation](https://beam.apache.org/documentation/sdks/python-type-safety/)."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "class MovieViewCoder(beam.coders.Coder):\n",
```
dvadym, line 585: Could you please add
#@title ....
and hide the code by default (it migth be too much information for the user)

docs/tutorial_1/1_beam_introduction.ipynb
```diff
@@ -1,876 +1,1681 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "@beam.typehints.with_output_types(MovieView)  # type-hint annotation for the output\n",
+    "class ParseMovieViews(beam.DoFn):\n",
```
dvadym, line 726: Could you please add
#@title ....
and hide the code by default (it might be too much information for the user)

docs/tutorial_1/1_beam_introduction.ipynb
```diff
@@ -1,876 +1,1681 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "@beam.typehints.with_output_types(MovieTitle)  # type-hint annotation for the output\n",
+    "class ParseMovieTitles(beam.DoFn):\n",
```
dvadym, line 768: Could you please add
#@title ....
and hide the code by default (it might be too much information for the user)

docs/tutorial_1/1_beam_introduction.ipynb
```diff
@@ -1,876 +1,1681 @@
...
+    "The first step will be to create a function that returns the reference `PCollection` we want to deal with. Having two separate **data objects** (i.e. `MovieView` and `MovieTitle`), we might assume we will be needing two separate functions to gather instances of either type."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def netflix_movie_views_collection(p: beam.Pipeline, data_file: str = DATA_FILE) -> beam.PCollection[MovieView]:\n",
```
dvadym, line 808: Please combine both functions in one cell (it's better for the user to run less cells) and

could you please add
#@title ....
and hide the code by default (it might be too much information for the user)

docs/tutorial_1/1_beam_introduction.ipynb
```diff
@@ -1,876 +1,1681 @@
...
+    "# Once defined, we can register the new Coders\n",
+    "beam.coders.registry.register_coder(MovieView, MovieViewCoder)\n",
+    "beam.coders.registry.register_coder(MovieTitle, MovieTitleCoder)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "##### Workaround `multiprocessing` issues with namespace"
```
dvadym, line 619: Please add optional, and that it's not effective for Google Colab to the name of this subsection and hide the content of this subsection by default

docs/tutorial_1/1_beam_introduction.ipynb
```diff
@@ -1,876 +1,1681 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import time\n",
+    "from apache_beam.runners import DirectRunner\n",
+    "from apache_beam.options.pipeline_options import PipelineOptions\n",
+    "\n",
+    "def run_pipeline(pipeline_fn: PipelineModule, \n",
```
dvadym, line 917: could you please add
#@title ....
and hide the code by default (it might be too much information for the user) ?

dvadym (changes requested):
Thanks a lot for improvements! Sry for late reply, I've been finishing other projects. Further I'll try to keep latency on review not more than 1-2 working days.

I've left comments please check. I like this PR improvements, most my comments are about hiding some cells by default and merging some code cells. It might be too much information for the user if everything is open by default :)

docs/tutorial_1/1_beam_introduction.ipynb
leriomaggio: Sorry @dvadym which one? Review of notebooks via Github is indeed very painful! :) 
Trying to recollect all your useful comments.. thanks!

docs/tutorial_1/1_beam_introduction.ipynb
leriomaggio: Done

docs/tutorial_1/1_beam_introduction.ipynb
leriomaggio: done

docs/tutorial_1/1_beam_introduction.ipynb
leriomaggio: Yep, you're right. Rephrased the paragraph with: 

```
However, even just limiting our analysis to one single file (i.e. >20M lines) might be prohibitive, and indeed beyond of the scope of this tutorial.
For real scalability experiments, we would need better computation capacity (which generally goes by far beyond laptops specs).
```

docs/tutorial_1/1_beam_introduction.ipynb
leriomaggio: Sure - I guess you're right! :)

docs/tutorial_1/1_beam_introduction.ipynb
dvadym: Sry, I probably incorrectly located a position in .ipynb file when mapped from Colab. I can't find it now. Never mind about this comment.

 Yeah, reviewing notebooks in Github is painful :(

PR #14, opened 2021-05-04
SparkOperations implementation

### LINKED ISSUE #6: Spark adapter
## Context

**PipelineOperations** interface provides interface with generic  pipeline operations and it allows us to encapsulate a pipeline framework (Apache Beam or Apache Spark or other).

## Goal 

Implement **SparkOperations** class, which is inherited from **PipelineOperations** and implements all abstract methods using Apache Spark API

For example 

```
def map(self, col, fn, stage_name: str):
	return col.map(fn)
```

**Note**: stage_name is needed for Beam, and it might be ignored for **SparkPipelineOperations** 

**Note**: it is required to investigate how well the current PipelineOperation model is consistent with Spark APIs.

**Note**: it is better to split an implementation of different methods in different pull requests.

## Links
TODO: add a link to PipelineOperations sources when [pull request](https://github.com/OpenMined/PipelineDP/pull/2) is submitted

boyaryn:
## Description
Implemented SparkOperations class.
Resolves #6.

## Affected Dependencies
Python package `pyspark` should be installed if this implementation is going to be used.

## How has this been tested?
The implemented methods were checked in Jupyter Notebook using python 3.7 and pyspark 2.4.5.

## Checklist
- [x] My changes are covered by tests. (Tests are going to be added later)


pipeline_dp/pipeline_operations.py
```diff
@@ -74,3 +86,48 @@ def sample_fixed_per_key(self, col, n: int, stage_name: str):
...
+
+  def keys(self, rdd, stage_name: str = None):
+    return rdd.keys()
+
+  def values(self, rdd, stage_name: str = None):
+    return rdd.values()
+
+  def sample_fixed_per_key(self, rdd, n: int, stage_name: str = None):
+    return rdd.groupByKey()\
+      .mapValues(lambda it: list(it))\
```
dvadym, line 128: I'm thinking about case when there is a lot of elements per key. Is it possible that pyspark.resultiterable.ResultIterable represents a large lazy collection which doesn't fit in memory?
Is there some way not to materialize it to list?
For example Beam sample_per_key does in the following way:
1.There is a combiner for getting top_n elements per key
2.sample_per_key = GroupByKey().mapValues(lambda it: ((rand(), x) for x in it)).reduceValues(top_n)
i.e. using rand() to ensure a random order and choosing top n elements in this order guarantees uniformly random sampling. 

Would something like this possible in PySpark?

pipeline_dp/pipeline_operations.py
```diff
@@ -74,3 +86,48 @@ def sample_fixed_per_key(self, col, n: int, stage_name: str):
...
+  def filter(self, rdd, fn, stage_name: str = None):
+    return rdd.filter(fn)
+
+  def keys(self, rdd, stage_name: str = None):
+    return rdd.keys()
+
+  def values(self, rdd, stage_name: str = None):
+    return rdd.values()
+
+  def sample_fixed_per_key(self, rdd, n: int, stage_name: str = None):
```
dvadym, line 126: This function is not just a wrapper of PySpark function, could you please write some simple test to it?

dvadym (commented):
Thanks for PR, it looks great. I left some questions.

pipeline_dp/pipeline_operations.py
dvadym: Sry, in the description how Beam works for sampling, not GroupByKey but CombinePerKey is used, which means it's fine to have a large number of elements per key

pipeline_dp/pipeline_operations.py
```diff
@@ -74,3 +86,48 @@ def sample_fixed_per_key(self, col, n: int, stage_name: str):
 
   def count_per_element(self, col, stage_name: str):
     return col | stage_name >> combiners.Count.PerElement()
+
+
+class SparkOperations(PipelineOperations):
+  """Apache Spark Core adapter."""
```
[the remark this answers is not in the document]
boyaryn: I borrowed this terminology from [here](https://spark.apache.org/docs/latest/api/python/reference/pyspark.html). In addition to older Spark RDD, which is similar to Beam PCollection and called Spark Core at that link, there is also a more performance-optimized alternative, which resembles a relational table and is called Spark SQL there.

Is it better to keep here _Spark Core_ (or _Spark RDD_)? Or just _Spark_, and rename it to a more specific name when/if we add another Spark Operation implementation in future?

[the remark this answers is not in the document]
dvadym: Thanks for explanation! Probably it makes sense to rename SparkOperations to SparkRDDOperations or SparkCoreOperations, to emphasize which exactly Spark API is used. WDYT?

pipeline_dp/pipeline_operations.py
boyaryn: You're spot on. I hesitated implementing it this way because of potential issues in case of imbalanced distribution of values between keys. My excuse was that people dealing with huge datasets will be more likely to use the Scala/Java version of this library and Dataframe/Dataset based implementation.

However, on second thought, it's possible to prevent Out of Memory errors by using a simple implementation which I'll push with the next commit.

Spark's _top(…)_ doesn't fit well here because it would gather everything to the driver. _RDD.sampleByKey(…)_ could be used here but it required proving fractions, not _n_, so preliminary collecting of sample data to the driver (like one done in Spark's RangePartitioner) would be necessary, which is complicated.

pipeline_dp/pipeline_operations.py
boyaryn: Done.

pipeline_dp/pipeline_operations.py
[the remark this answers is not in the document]
boyaryn: Good idea! I've renamed it.

tests/pipeline_operations_test.py
```diff
@@ -1,12 +1,47 @@
...
+
 class PipelineOperationsTest(unittest.TestCase):
   pass
 
+
+class SparkRDDOperationsTest(unittest.TestCase):
+  @classmethod
+  def setUpClass(cls):
+    conf = pyspark.SparkConf()
+    SparkRDDOperationsTest.sc = pyspark.SparkContext(conf=conf)
```
dvadym, line 15: Nit (Nit = optional improvement): SparkRDDOperationsTest -> cls

tests/pipeline_operations_test.py
```diff
@@ -1,12 +1,47 @@
...
+
+  def test_sample_fixed_per_key(self):
+    spark_operations = pipeline_dp.SparkRDDOperations()
+    data = [(1, 11), (2, 22), (3, 33), (1, 14), (2, 25), (1, 16)]
+    dist_data = SparkRDDOperationsTest.sc.parallelize(data)
```
dvadym, line 20: would self.sc work?

pipeline_dp/pipeline_operations.py
```diff
@@ -75,6 +87,51 @@ def sample_fixed_per_key(self, col, n: int, stage_name: str):
...
+    return rdd.filter(fn)
+
+  def keys(self, rdd, stage_name: str = None):
+    return rdd.keys()
+
+  def values(self, rdd, stage_name: str = None):
+    return rdd.values()
+
+  def sample_fixed_per_key(self, rdd, n: int, stage_name: str = None):
+    return rdd.mapValues(lambda x: [x])\
```
dvadym, line 127: thanks! I like this implementation, it's simple, scalable and provides uniform sampling

dvadym (commented):
Some minor comments

dvadym:
I realized that it would be great to have E2E example on Spark and since you're Spark expert it would be easier for you to add it. Could you please add it?

Could you please add to https://github.com/OpenMined/PipelineDP/blob/main/examples/movie_view_ratings.py a possibility to run on Spark. Atm this example simply outputs its input, but with implementing other DP aggregations it will output DP aggregations results.

Ideally it would be that running on Beam and Spark would only differ in some setup work (like setting SparkContext, reading input, writing output) and [creating different](https://github.com/OpenMined/PipelineDP/blob/6694a76063e7260c521b2740ae0b27ff10895a8d/examples/movie_view_ratings.py#L71) PipelineOperations.

For choosing running on Beam or Spark, you can define flags.DEFINE_enum as [here](https://abseil.io/docs/python/guides/flags#example-usage).

pipeline_dp/pipeline_operations.py
```diff
@@ -75,6 +87,51 @@ def sample_fixed_per_key(self, col, n: int, stage_name: str):
...
+      .reduceByKey(lambda x, y: random.sample(x+y, min(len(x+y), n)))
```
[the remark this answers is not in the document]
boyaryn: It makes sense. Fixed.

tests/pipeline_operations_test.py
boyaryn: Yeah, it'll look better. Changed.

tests/pipeline_operations_test.py
boyaryn: No, it wouldn't because `setUpClass(cls)` is annotated in the parent class as @classmethod so we can't access and initialize `self.sc` in `setUpClass(cls)`.

pipeline_dp/pipeline_operations.py
boyaryn: Thanks for drawing attention to uniformity! In fact, in general `reduceByKey` won't give us uniform sampling. An RDD in Spark consists of partitions, and [`reduceByKey` firstly reduces each partition before sending the results for final reductions](https://spark.apache.org/docs/latest/api/scala/org/apache/spark/rdd/PairRDDFunctions.html#reduceByKey(func:(V,V)=%3EV):org.apache.spark.rdd.RDD[(K,V)]). It's implemented this way to reduce shuffling (i.e. traffic between executors/workers). So for a given key the less values reside in a specific partition, the more likely (when comparing with values for this key from other partitions) these values will get to the final sample.

Should we keep this implementation or use the previous one for now?

Ideally, we'd have to collect statistics about cardinality for each key to the driver in the first phase and based on this to sample the corresponding number of values from each partition. But it's complicated and more time is needed for this implementation.

boyaryn:
Yeah, it's a good idea to have a Spark example as well. I'll add it.

pipeline_dp/pipeline_operations.py
dvadym: Ah, correct, this is not uniform sampling. The uniformness is not a strict requirement. 

Could you please add a comment that this sampling is not uniform? It's important to inform about such caveats :). I think scalability is more important that uniformness here.

boyaryn: Added the comment :)

dvadym (approved):
Let's merge this PR w/o an example and make the example in the next PR.

PR #16, opened 2021-05-04
Add local map operation

replomancer:
## Description
Adds a method for local map operation. See comment about a file change below.

## Affected Dependencies
Nope.

## How has this been tested?
Tests added :test_tube: 
The empty `__init__.py` file has been added because without it pytest (and probably other testing frameworks?) cannot find `pipeline_dp` for importing (Python 3.8.5). I can imagine PyCharm users do not see the issue, but it will be needed for CI.

## Checklist

### DIFF
--- a/pipeline_dp/pipeline_operations.py
+++ b/pipeline_dp/pipeline_operations.py
@@ -78,8 +78,8 @@ class BeamOperations(PipelineOperations):
 class LocalPipelineOperations(PipelineOperations):
   """Local Pipeline adapter."""
   
-  def map(self, col, fn, stage_name: str):
-    pass
+  def map(self, col, fn, stage_name: str = None):
+    return map(fn, col)
 
   def map_tuple(self, col, fn, stage_name: str):
     pass
--- a/tests/pipeline_operations_test.py
+++ b/tests/pipeline_operations_test.py
@@ -1,12 +1,25 @@
 import unittest
 
-import pipeline_dp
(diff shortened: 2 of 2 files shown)


pipeline_dp/pipeline_operations.py
```diff
@@ -78,8 +78,8 @@ def count_per_element(self, col, stage_name: str):
 class LocalPipelineOperations(PipelineOperations):
   """Local Pipeline adapter."""
   
-  def map(self, col, fn, stage_name: str):
-    pass
+  def map(self, col, fn, stage_name: str = None):
+    return map(fn, col)
```
replomancer, line 82: Here the order is reversed: `fn` is passed before `col`. That's how `map` works in most languages, so current order of the arguments in `PipelineOperations.map` may be confusing to people. Can we reverse it or should `col` stay as the first argument? Same thing with filter: function is passed before iterable to Python's filter.

dvadym: That's a good question! I think this is a common thing that different things are different in different languages/frameworks :)

For example in Spark and Beam collection is before fn in statement:

Beam:  col | beam.Map(fn)
Spark: col.Map(fn)

Another thing is that for all PipelineOperations method col is always an argument, fn isn't (eg. in Keys()), so it's more uniform to have col always as the 1st argument.

dvadym (approved):
Thanks!

PR #17, opened 2021-05-05
Add Budget accounting

AlexandraImbrisca:
## Description
Adds the budget accountant that manages the privacy budget.

## Affected Dependencies
None.

## How has this been tested?
The budget_accounting_test.py file contains some simple additional tests.

## Checklist

### DIFF
--- a/pipeline_dp/budget_accounting.py
+++ b/pipeline_dp/budget_accounting.py
@@ -5,98 +5,98 @@ from dataclasses import dataclass
 
 @dataclass
 class Budget:
-  """Manages the budget allocated for an operation.
+    """Manages the budget allocated for an operation.
 
-  The values for eps and delta will be computed only after the pipeline was
-  constructed and the method compute_budgets was called.
-  """
-  _eps: float = None
-  _delta: float = None
-
-  @property
-  def eps(self):
-    """Parameter of (eps, delta)-differential privacy.
-
-    Raises:
-      AssertionError: The privacy budget is not calculated yet.
+    The values for eps and delta will be computed only after the pipeline was
+    constructed and the method compute_budgets was called.
     """
-    if self._eps is None:
-      raise AssertionError("Privacy budget is not calculated yet.")
-    return self._eps
+    _eps: float = None
+    _delta: float = None
 
-  @property
-  def delta(self):
-    """Parameter of (eps, delta)-differential privacy.
+    @property
+    def eps(self):
+        """Parameter of (eps, delta)-differential privacy.
 
-    Raises:
-      AssertionError: The privacy budget is not calculated yet.
-    """
-    if self._delta is None:
-      raise AssertionError("Privacy budget is not calculated yet.")
-    return self._delta
+        Raises:
+            AssertionError: The privacy budget is not calculated yet.
+        """
+        if self._eps is None:
+            raise AssertionError("Privacy budget is not calculated yet.")
+        return self._eps
 
-  def set_eps_delta(self, eps, delta):
-    self._eps = eps
-    self._delta = delta
+    @property
+    def delta(self):
+        """Parameter of (eps, delta)-differential privacy.
+
+        Raises:
+            AssertionError: The privacy budget is not calculated yet.
+        """
+        if self._delta is None:
+            raise AssertionError("Privacy budget is not calculated yet.")
+        return self._delta
+
+    def set_eps_delta(self, eps, delta):
+        self._eps = eps
+        self._delta = delta
 
 @dataclass
 class RequestedBudget:
-  """Manages the budget requested for an operation."""
-  budget: Budget
-  weight: float
-  use_eps: bool
-  use_delta: bool
+    """Manages the budget requested for an operation."""
+    budget: Budget
+    weight: float
+    use_eps: bool
+    use_delta: bool
 
 class BudgetAccountant:
-  """Manages the privacy budget."""
+    """Manages the privacy budget."""
 
-  def __init__(self, epsilon, delta):
-    """Constructs a BudgetAccountant.
+    def __init__(self, epsilon, delta):
+        """Constructs a BudgetAccountant.
 
-    Args:
-      epsilon, delta: Parameters of (epsilon, delta)-differential privacy.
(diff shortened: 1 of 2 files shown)


pipeline_dp/budget_accounting.py
```diff
@@ -1,20 +1,91 @@
...
+      raise ValueError(f"Epsilon must be positive, not {epsilon}.")
     if delta < 0:
-      raise ValueError(f"Delta must non-negative, not {delta}")
+      raise ValueError(f"Delta must be non-negative, not {delta}.")
+
     self._eps = epsilon
     self._delta = delta
+    self._requested_budgets = []
+
+  def request_budget(self, weight, use_eps, use_delta):
```
dvadym, line 61: Looking at tests, I realized that it's might be hard to distinguish use_eps/use_delta if they are used as positional arguments, please make them keyword arguments (i.e. enforce to use them to be keyword arguments), namely:
def request_budget(self, weight, *, use_eps, use_delta):

pipeline_dp/budget_accounting.py
```diff
@@ -1,20 +1,91 @@
...
+      print("No budgets were requested.")
+      return
+
+    total_weight = sum([budget.weight for budget in self._requested_budgets])
 
-  # TODO: implement BudgetAccountant functionality.
+    for budget in self._requested_budgets:
+      eps = 0
+      delta = 0
+      if total_weight != 0:
```
dvadym, line 88: since some operations might not use eps or delta. Then computing total_weight independenly of use_eps/use_delta, might lead to non-allocating eps/delta as in test_compute_budgets. It's better to allocate whole budget, since it will increase accuracy.

Could you please compute separately weights by eps and delta to ensure that the whole budget is allocated?

pipeline_dp/budget_accounting.py
```diff
@@ -1,20 +1,91 @@
 """Privacy budget accounting for DP pipelines."""
 
 
+class Budget:
+  """Manages the budget allocated for an operation."""
+
+  def __init__(self, weight, use_eps, use_delta, eps=None, delta=None):
+    """Constructs a Budget.
```
dvadym, line 8: No need to have eps/delta as arguments of the constructor (explanation: the prototype used eps/delta because it was an option of eager mode, let's for now go only with lazy mode, maybe we will not eager mode :) )

pipeline_dp/budget_accounting.py
```diff
@@ -1,20 +1,91 @@
...
+    constructed and the method compute_budgets was called.
+
+    Args:
+      weight: The weight is used to get more accurate results.
+      use_eps: A boolean that is False when the operation doesn't need epsilon.
+      use_delta: A boolean that is False when the operation doesn't need delta.
+    """
+    self._eps = eps
+    self._delta = delta
+    self.weight = weight
```
dvadym, line 20: Thought again, I think from the API point of view it's better to avoid having weight/use_eps/use_delta in public API of this class. These fields are mostly implementation details.

My suggestions is the following, introduce @ dataclass RequestedBudget (or suggest another name)
with fields
budge:Budget
weight
use_eps
use_delta

And to keep in BudgetAccountant a list of RequestedBudget.

WDYT?

tests/budget_accounting_test.py
```diff
@@ -6,17 +6,37 @@
 class BudgetAccountant(unittest.TestCase):
```
dvadym, line 6: Sry, my fault, could you please rename to BudgetAccountantTest?

pipeline_dp/budget_accounting.py
```diff
@@ -1,20 +1,91 @@
...
+    self.use_eps = use_eps
+    self.use_delta = use_delta
+
+  @property
+  def eps(self):
+    """eps: Optional; Parameter of (eps, delta)-differential privacy."""
```
dvadym, line 26: Could you please remove eps: Optional?

Please add: Raise an error if the budget is not calculated yet. 


(the same with delta).

pipeline_dp/budget_accounting.py
```diff
@@ -1,20 +1,91 @@
...
+      method compute_budgets is called.
+    """
+    budget = Budget(weight=weight, use_eps=use_eps, use_delta=use_delta)
+    self._requested_budgets.append(budget)
+    return budget
+
+  def compute_budgets(self):
+    """Computes the budgets after constructing the pipeline."""
+    if not self._requested_budgets:
+      print("No budgets were requested.")
```
dvadym, line 80: Please use logging (sry in the prototype it wasn't used):
import logging

print -> logging.warning

(here is more information on logging https://docs.python.org/3/howto/logging.html )

dvadym (changes requested):
Thanks a lot for contributing! It looks great in general. Thanks for updating comments!

I left some comments. If you have any question please don't hesitate to ask (in the review, in slack)

pipeline_dp/budget_accounting.py
AlexandraImbrisca: It makes a lot of sense since the fields: weight, use_eps, use_delta are not useful outside the methods request_budget and compute_budgets. I will modify the classes. Thanks for pointing it out!

pipeline_dp/budget_accounting.py
```diff
@@ -1,20 +1,102 @@
...
+            A "lazy" budget object that doesn't contain epsilon/delta until the
+            method compute_budgets is called.
+        """
+        budget = Budget()
+        requested_budget = RequestedBudget(budget, weight, use_eps, use_delta)
+        self._requested_budgets.append(requested_budget)
+        return budget
+
+    def compute_budgets(self):
+        """Computes the budgets after constructing the pipeline."""
```
dvadym, line 86: Please add to the function comment something like:

All previously requested Budget objects are updated with corresponding budget values.

dvadym (changes requested):
Thanks it looks great! I left some minor comments

AlexandraImbrisca:
Thank you very much for the feedback!

dvadym (approved):
Thanks for implementing Budget Accountant!

PR #18, opened 2021-05-05
Implement local map_tuple method

zachferr:
## Description
Added the local map_tuple implementation and the tests.

I also updated the indentation of the files to be 4 spaces as per the weekly sync

## Affected Dependencies
N/A

## How has this been tested?
- I added unit tests for my change
- In order to get the test directory to see the pipeline_dp directory I had to run ```export PYTHONPATH="${PYTHONPATH}:/Users/zachferr/Documents/github/PipelineDP```, I'm not sure if this is intended or my project is configured incorrectly.

## Checklist


pipeline_dp/pipeline_operations.py
```diff
@@ -6,101 +6,102 @@
...
-    return map(fn, col)
+    """Local Pipeline adapter."""
+
+    def map(self, col, fn, stage_name: str = None):
+        return map(fn, col)
 
-  def map_tuple(self, col, fn, stage_name: str):
-    pass
+    def map_tuple(self, col, fn, stage_name: str = None):
+        return [fn(k, v) for k, v in col]
```
dvadym, line 86: We need to have all operations lazy, that's a cool thing that Python often helps with that :). Please use generator expressions https://www.python.org/dev/peps/pep-0289/ (basically replace [...] -> (...) ]

An explanation why async is convenient for us:
1.Having all operations to be async allows to build a pipeline and then to execute it
2.After the pipeline built we can optimize the privacy budget usage (compute the right amount of noise to add) and then to use this noise when the pipeline runs
3.All pipeline frameworks (Beam/Spark) are inherently async - a pipeline is built on the main machine and then executed on workers - which is convenient for us.

dvadym (changes requested):
Thanks! I left some comments

pipeline_dp/pipeline_operations.py
zachferr: I forgot about that requirement, thanks for reminding me! 

Should be fixed now

dvadym (approved):
Thank you!

PR #22, opened 2021-05-14
Add local group_by_key operation

zachferr:
## Description
Adding local group_by_key and associated tests

## Affected Dependencies
N/A

## How has this been tested?
- Unit test

## Checklist

### DIFF
--- a/pipeline_dp/pipeline_operations.py
+++ b/pipeline_dp/pipeline_operations.py
@@ -5,7 +5,8 @@ import random
 import abc
 import apache_beam as beam
 import apache_beam.transforms.combiners as combiners
-from collections import defaultdict
+import collections
+import typing
 
 
 class PipelineOperations(abc.ABC):
@@ -151,15 +152,15 @@ class LocalPipelineOperations(PipelineOperations):
     def map(self, col, fn, stage_name: str = None):
         return map(fn, col)
 
-    def map_tuple(self, col, fn, stage_name: str = None):
+    def map_tuple(self, col, fn, stage_name: typing.Optional[str] = None):
         return (fn(k, v) for k, v in col)
 
     def map_values(self, col, fn, stage_name: str):
         pass
 
-    def group_by_key(self, col, stage_name: str = None):
+    def group_by_key(self, col, stage_name: typing.Optional[str] = None):
         def group_by_key_generator():
-            d = defaultdict(list)
+            d = collections.defaultdict(list)
             for key, value in col:
                 d[key].append(value)
             for item in d.items():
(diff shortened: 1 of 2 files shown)


pipeline_dp/pipeline_operations.py
```diff
@@ -156,8 +157,11 @@ def map_tuple(self, col, fn, stage_name: str = None):
     def map_values(self, col, fn, stage_name: str):
         pass
 
-    def group_by_key(self, col, stage_name: str):
-        pass
+    def group_by_key(self, col, stage_name: str = None):
```
dvadym, line 160: We need to have it lazy, d.items() is generator-like object, but for-loop is executed eagerly. What do you think about the following approach

```
def group_by_key(self, ...):
   def group_by_key_generator(...):
         d = defaultdict(list)
         for key, value in col:
             d[key].append(value)
         for item in d.items():
             yield item
   return group_by_key_generator
```

dvadym (changes requested):
Thanks for contributing! I left a comment

pipeline_dp/pipeline_operations.py
zachferr: I updated it, let me know what you think.
Thanks for catching this I didn't consider that

pipeline_dp/pipeline_operations.py
```diff
@@ -5,6 +5,7 @@
 import abc
 import apache_beam as beam
 import apache_beam.transforms.combiners as combiners
+from collections import defaultdict
```
dvadym, line 8: I recalled that by Google style guide, packages should be imported not classes. 
https://google.github.io/styleguide/pyguide.html#224-decision

Please update to 
import collections

pipeline_dp/pipeline_operations.py
dvadym: Ah, yeah, thanks, () was missing in "return group_by_key_generator"

dvadym (changes requested):
Thanks, I left a minor comment

replomancer:
LGTM  (I guess only thing left is the style fix)

replomancer (approved):

pipeline_dp/pipeline_operations.py
```diff
@@ -156,8 +157,14 @@ def map_tuple(self, col, fn, stage_name: str = None):
     def map_values(self, col, fn, stage_name: str):
         pass
 
-    def group_by_key(self, col, stage_name: str):
-        pass
+    def group_by_key(self, col, stage_name: str = None):
```
chinmayshah99, line 160: change this to `stage_name: Optional[str]` to confor to mypy standards.

chinmayshah99 (changes requested):

pipeline_dp/pipeline_operations.py
zachferr: Added it

dvadym (approved):
Thanks!

PR #24, opened 2021-05-18
Add more local pipeline operations

replomancer:
## Description
Adds every other remaining method to `LocalPipelineOperations` (we split work with Zachery).

## Affected Dependencies
None.

## How has this been tested?
Unit tests added :broom:

## Checklist

### DIFF
--- a/pipeline_dp/pipeline_operations.py
+++ b/pipeline_dp/pipeline_operations.py
@@ -149,14 +149,14 @@ class SparkRDDOperations(PipelineOperations):
 class LocalPipelineOperations(PipelineOperations):
     """Local Pipeline adapter."""
 
-    def map(self, col, fn, stage_name: str = None):
+    def map(self, col, fn, stage_name: typing.Optional[str] = None):
         return map(fn, col)
 
     def map_tuple(self, col, fn, stage_name: typing.Optional[str] = None):
         return (fn(k, v) for k, v in col)
 
-    def map_values(self, col, fn, stage_name: str):
-        pass
+    def map_values(self, col, fn, stage_name: typing.Optional[str] = None):
+        return ((k, fn(v)) for k, v in col)
 
     def group_by_key(self, col, stage_name: typing.Optional[str] = None):
         def group_by_key_generator():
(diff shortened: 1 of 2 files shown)


pipeline_dp/pipeline_operations.py
```diff
@@ -153,23 +154,23 @@ def map(self, col, fn, stage_name: str = None):
     def map_tuple(self, col, fn, stage_name: str = None):
         return (fn(k, v) for k, v in col)
 
-    def map_values(self, col, fn, stage_name: str):
-        pass
+    def map_values(self, col, fn, stage_name: str = None):
```
chinmayshah99, line 157: can you change this to `stage_name: Optional[str]`

chinmayshah99 (changes requested):

pipeline_dp/pipeline_operations.py
replomancer: Has it been decided that we're adding mypy to the project? Why add the annotation for only one argument then?

replomancer: Done.

pipeline_dp/pipeline_operations.py
```diff
@@ -147,29 +150,29 @@ def count_per_element(self, rdd, stage_name: str = None):
...
+    def values(self, col, stage_name: Optional[str] = None):
+        return (v for k, v in col)
 
     def sample_fixed_per_key(self, col, n: int, stage_name: str):
         pass
 
-    def count_per_element(self, col, stage_name: str):
-        pass
+    def count_per_element(self, col, stage_name: Optional[str] = None):
+        yield from collections.Counter(col).items()
```
dvadym, line 178: I like idea with Counter!

One question, in this line, is 'col' accessed the first time only when iterations over count_per_element() started?

What do you think about adding the 'lazyness' test:
Let 'exception_generator' is a generator which raise an exception on the first access to it

and then
def test_lazyness():
  self.ops.map(exception_generator) # should be no exception
  self.ops.map_tuple(exception_generator)
   ....

on implementing a new method it would be easy to add to the test and to ensure that the method is lazy

dvadym (commented):
Thanks, it look great. I left a question and a suggestion

chinmayshah99:
@replomancer there are some conflicts. Can you please resolve them?

pipeline_dp/pipeline_operations.py
replomancer: Cool idea for testing!
I added a test like this and yes, `count_per_element` works as expected.

tests/pipeline_operations_test.py
```diff
@@ -72,6 +82,57 @@ def test_local_group_by_key(self):
...
+
+    def test_laziness(self):
+        def exceptions_generator_function():
+            yield 1 / 0
+
+        def assert_laziness(operator, *args):
+            try:
+                operator(exceptions_generator_function(), *args)
+            except ZeroDivisionError:
+               self.fail(f"local {operator.__name__} is not lazy")
```
replomancer, line 122: This function exists for tests to fail properly with `AssertionError` (`assertNotRaises` does not exists in `unittest`). 

e.g. when filter implementation is broken:
```python
        return list(filter(fn, col))
```
the fail looks like this:
```
tests/pipeline_operations_test.py:122: in assert_laziness
    self.fail(f"local {operator.__name__} is not lazy")
E   AssertionError: local filter is not lazy
```

dvadym: Thanks for explanation. I like that approach!

dvadym (approved):

Issue #35, opened 2021-06-02
PLD based Budget Accountant


dvadym:
## Context 

The privacy budget accounting is an important feature of the DP aggregation system. On the [issue](https://github.com/OpenMined/PipelineDP/issues/5) BudgetAccountant was implemented. The BudgetAccountant uses the naive composition, namely the total budget is the sum of budgets:

(𝜺_1, 𝛿_1) + (𝜺_2, 𝛿_2) = (𝜺_1+𝜺_2,𝛿_1+𝛿_2).

The downside of using the naive composition is that the total budget is growing linearly with the number of DP aggregations. There are others, more advanced ways of composing budgets, such that the total budget is growing slower than linearly in the number of DP aggregations. 

One of those ways is PLD (privacy loss distributions, [math details](https://eprint.iacr.org/2018/820.pdf)), which is implemented in [dp_accounting](https://github.com/google/differential-privacy/tree/main/python/dp_accounting) library.

## Goals
Implementation class PLDBudgetAccountant, which uses  [dp_accounting](https://github.com/google/differential-privacy/tree/main/python/dp_accounting) library.

```
class MechanismType(Enum):
   LAPLACE = 1
   GAUSSIAN = 2
   # more mechanisms will be added later

class PLDBudgetAccountant:
   def __init__(self, total_eps : double, total_delta : double):
   def request_budget(self, weight : double, mechanism_type : MechanismType, sensitivity:float) -> Budget
   def compute_budgets(self):

class PLDBudget:
   @property def noise_parameter()
```

The idea is that **PLDBudgetAccountant** collects all requests on budget usage from different DP operations (the same way as **BudgetAccountant**).  **request_budget()** returns a **PLDBudget** object which does not have yet **noise_parameter**. And then when the pipeline is constructed, (i.e. all DP operations are known) **compute_budget** is called and it computes the minimum noise for each DP aggregation, such that the total budget is not more than (total_eps, total_delta)

**Note**: for some DP operations it might be more important to get more accurate result, that is why **weight** parameter is used.

# Notes on implementation

**Note:** In this task we can consider PLD as a blackbox, with the following properties
- a PLD object corresponds to a DP mechanism
- there is a way to compose PLD object and the result is a PLD object, which corresponds to composition of corresponding DP mechanisms
- PLD allows to compute **eps** for given **delta**.

## Using PLD for composing of DP mechanisms

Let input is 0<= delta < 1 and DP mechanisms (M_1, ..., M_n) 

[...]

PR #38, opened 2021-06-05
PLD budget accountant

jspacek:
## Description

Functionality that gives the user the ability to input epsilon and delta values, compute a pipeline budget where the noise is adjusted for the given epsilon.

_Input_: Epsilon and Delta, _Output_: Minimized noise

Related to #35 

## Affected Dependencies

N/A

## How has this been tested?

Unit tests where the calculated noise value is compared to the noise value returned by the PLD library for the same epsilon and delta. 

eg. 
```
compare_pld = accountant._compose_distributions(expected_noise)
self.assertAlmostEqual(epsilon, compare_pld.get_epsilon_for_delta(delta), 3)
```
The `expected_noise` should match the calculated noise from PipelineDP because the same epsilon and delta are input.

```
accountant = PLDBudgetAccountant(epsilon, delta)
accountant.compute_budgets()
self.assertEqual(expected_noise, accountant.minimum_noise)
```

## Checklist

### DIFF
--- a/pipeline_dp/budget_accounting.py
+++ b/pipeline_dp/budget_accounting.py
@@ -1,7 +1,10 @@
 """Privacy budget accounting for DP pipelines."""
 
 import logging
+import math
 from dataclasses import dataclass
+from pipeline_dp.aggregate_params import NoiseKind
+from dp_accounting import privacy_loss_distribution as pldlib
 
 
 @dataclass
@@ -59,10 +62,8 @@ class BudgetAccountant:
         Args:
             epsilon, delta: Parameters of (epsilon, delta)-differential privacy.
         """
-        if epsilon <= 0:
-            raise ValueError(f"Epsilon must be positive, not {epsilon}.")
-        if delta < 0:
-            raise ValueError(f"Delta must be non-negative, not {delta}.")
+
+        _validate_epsilon_delta(epsilon, delta)
 
         self._eps = epsilon
         self._delta = delta
@@ -74,8 +75,8 @@ class BudgetAccountant:
 
         Args:
             weight: The weight used to compute epsilon and delta for the budget.
-            use_eps: A boolean that is False when the operation doesn't need epsilon.
-            use_delta: A boolean that is False when the operation doesn't need delta.
+            use_eps: False when the operation doesn't need epsilon.
+            use_delta: False when the operation doesn't need delta.
 
         Returns:
             A "lazy" budget object that doesn't contain epsilon/delta until the
@@ -87,7 +88,7 @@ class BudgetAccountant:
         return budget
 
     def compute_budgets(self):
-        """All previously requested Budget objects are updated with corresponding budget values."""
+        """Updates all previously requested Budget objects with corresponding budget values."""
         if not self._requested_budgets:
             logging.warning("No budgets were requested.")
             return
@@ -100,7 +101,205 @@ class BudgetAccountant:
         for requested_budget in self._requested_budgets:
             eps = delta = 0
             if total_weight_eps:
-                eps = requested_budget.use_eps * self._eps * requested_budget.weight / total_weight_eps
+                numerator = requested_budget.use_eps * self._eps * requested_budget.weight
+                eps = numerator / total_weight_eps
             if total_weight_delta:
-                delta = requested_budget.use_delta * self._delta * requested_budget.weight / total_weight_delta
+                numerator = requested_budget.use_delta * self._delta * requested_budget.weight
+                delta = numerator / total_weight_delta
             requested_budget.budget.set_eps_delta(eps, delta)
+
+
+@dataclass
+class MechanismSpec:
+    """Specifies the parameters for a mechanism.
+
+    NoiseKind defines the kind of noise distribution.
+    noise is the minimized noise standard deviation.
+    """
+    noise_kind: NoiseKind
+    _noise_standard_deviation: float = None
+
+    @property
+    def noise_standard_deviation(self):
+        """Noise value for the mechanism.
+
+        Raises:
+            AssertionError: The noise value is not calculated yet.
+        """
+        if self._noise_standard_deviation is None:
+            raise AssertionError(
+                "Noise standard deviation is not calculated yet.")
+        return self._noise_standard_deviation
+
+
+@dataclass
+class MechanismSpecInternal:
+    """Stores sensitivity and weight not exposed in MechanismSpec."""
+    sensitivity: float
+    weight: float
+    mechanism_spec: MechanismSpec
+
+
+class PLDBudgetAccountant:
+    """Manages the privacy budget for privacy loss distributions.
+
+    It manages the privacy budget for the pipeline using the
+    Privacy Loss Distribution (PLD) implementation from Google's
+    dp_accounting library.
+    """
+
+    def __init__(self,
+                 total_epsilon: float,
+                 total_delta: float,
+                 pld_discretization: float = 1e-4):
+        """Constructs a PLDBudgetAccountant.
+
+        Args:
+            total_epsilon: epsilon for the entire pipeline.
+            total_delta: delta for the entire pipeline.
+            pld_discretization: `value_discretization_interval` in PLD library.
+                Smaller interval results in better accuracy, but increases running time.
+
+        Raises:
+            ValueError: Arguments are missing or out of range.
+        """
+
+        _validate_epsilon_delta(total_epsilon, total_delta)
+
+        self._total_epsilon = total_epsilon
+        self._total_delta = total_delta
+        self._mechanisms = []
+        self.minimum_noise_std = None
+        self._pld_discretization = pld_discretization
+
+    def request_budget(self,
+                       noise_kind: NoiseKind,
+                       sensitivity: float = 1,
+                       weight: float = 1) -> MechanismSpec:
+        """Request a budget.
+
+        Constructs a mechanism spec based on the parameters.
+        Adds the mechanism to the pipeline for future calculation.
+
+        Args:
+            noise_kind: The kind of noise distribution for the mechanism.
+            sensitivity: The sensitivity for the mechanism.
+            weight: The weight for the mechanism.
+
+        Returns:
+            A "lazy" mechanism spec object that doesn't contain the noise
+            standard deviation until compute_budgets is called.
+        """
+        if noise_kind == NoiseKind.GAUSSIAN and self._total_delta == 0:
+            raise AssertionError("Gaussian delta value must be greater than 0")
+        mechanism_spec = MechanismSpec(noise_kind=noise_kind)
+        mechanism_spec_internal = MechanismSpecInternal(
+            mechanism_spec=mechanism_spec,
+            sensitivity=sensitivity,
+            weight=weight)
+        self._mechanisms.append(mechanism_spec_internal)
+        return mechanism_spec
+
+    def compute_budgets(self):
+        """Computes the budget for the pipeline.
+
+        Composes the mechanisms and adjusts the amount of
+        noise based on given epsilon. Sets the noise for the
+        entire pipeline.
+        """
+        if not self._mechanisms:
+            return
+        if self._total_delta == 0:
+            standard_noise = len(self._mechanisms) / self._total_epsilon
+            self.minimum_noise_std = 0
+            for mechanism in self._mechanisms:
+                mechanism_noise_std = mechanism.sensitivity * standard_noise / mechanism.weight
+                mechanism.mechanism_spec._noise_standard_deviation = mechanism_noise_std
+                self.minimum_noise_std = self.minimum_noise_std + mechanism_noise_std
+        else:
+            minimum_noise_std = self._find_minimum_noise_std()
+            self.minimum_noise_std = minimum_noise_std
+            for mechanism in self._mechanisms:
+                mechanism_noise_std = mechanism.sensitivity * minimum_noise_std / mechanism.weight
+                mechanism.mechanism_spec._noise_standard_deviation = mechanism_noise_std
+
+    def _find_minimum_noise_std(self) -> float:
+        """Finds the minimum noise which satisfies the total budget.
+
+        Use binary search to find a minimum noise value that gives a
+        new epsilon close to the given epsilon (within a threshold).
+        By increasing the noise we can decrease the epsilon.
+
+        Returns:
+            The noise value adjusted for the given epsilon.
+        """
+        threshold = 1e-4
+        maximum_noise_std = self._calculate_max_noise_std()
+        low, high = 0, maximum_noise_std
+        while low + threshold < high:
+            mid = (high - low) / 2 + low
+            pld = self._compose_distributions(mid)
+            pld_epsilon = pld.get_epsilon_for_delta(self._total_delta)
+            if pld_epsilon <= self._total_epsilon:
+                high = mid
+            elif pld_epsilon > self._total_epsilon:
+                low = mid
+
+        return high
+
+    def _calculate_max_noise_std(self) -> float:
+        """Calculates an upper bound for the noise to satisfy the budget."""
+        max_noise_std = 1
+        pld_epsilon = self._total_epsilon + 1
+        while pld_epsilon > self._total_epsilon:
+            max_noise_std *= 2
+            pld = self._compose_distributions(max_noise_std)
+            pld_epsilon = pld.get_epsilon_for_delta(self._total_delta)
+        return max_noise_std
+
+    def _compose_distributions(
+            self,
+            noise_standard_deviation: float) -> pldlib.PrivacyLossDistribution:
+        """Uses the Privacy Loss Distribution library to compose distributions.
+
+        Args:
+            noise_standard_deviation: The noise of the distributions to construct.
+
+        Returns:
+            A PrivacyLossDistribution object for the pipeline.
+        """
+        composed, pld = None, None
+
+        for mechanism_spec_internal in self._mechanisms:
+            if mechanism_spec_internal.mechanism_spec.noise_kind == NoiseKind.LAPLACE:
+                # The Laplace distribution parameter = std/sqrt(2).
+                pld = pldlib.PrivacyLossDistribution.from_laplace_mechanism(
+                    mechanism_spec_internal.sensitivity *
+                    noise_standard_deviation / math.sqrt(2) /
+                    mechanism_spec_internal.weight,
+                    value_discretization_interval=self._pld_discretization)
+            elif mechanism_spec_internal.mechanism_spec.noise_kind == NoiseKind.GAUSSIAN:
+                pld = pldlib.PrivacyLossDistribution.from_gaussian_mechanism(
+                    mechanism_spec_internal.sensitivity *
+                    noise_standard_deviation / mechanism_spec_internal.weight,
+                    value_discretization_interval=self._pld_discretization)
+
+            composed = pld if composed is None else composed.compose(pld)
+
+        return composed
+
+
+def _validate_epsilon_delta(epsilon: float, delta: float):
+    """Helper function to validate the epsilon and delta parameters.
+
+    Args:
+        epsilon: The epsilon value to validate.
+        delta: The delta value to validate.
+
+    Raises:
+        A ValueError if either epsilon or delta are out of range.
+    """
+    if epsilon <= 0:
+        raise ValueError(f"Epsilon must be positive, not {epsilon}.")
+    if delta < 0:
+        raise ValueError(f"Delta must be non-negative, not {delta}.")
--- a/pipeline_dp/budget_accounting.py
+++ b/pipeline_dp/budget_accounting.py
@@ -3,8 +3,10 @@
 import logging
 import math
 from dataclasses import dataclass
+from dp_accounting.common import DifferentialPrivacyParameters
 from pipeline_dp.aggregate_params import NoiseKind
 from dp_accounting import privacy_loss_distribution as pldlib
+from dp_accounting import accountant
 
 
 @dataclass
@@ -191,7 +193,9 @@ class PLDBudgetAccountant:
             standard deviation until compute_budgets is called.
         """
         if noise_kind == NoiseKind.GAUSSIAN and self._total_delta == 0:
-            raise AssertionError("Gaussian delta value must be greater than 0")
+            raise AssertionError(
+                "The Gaussian mechanism requires that the pipeline delta is greater than 0"
+            )
         mechanism_spec = MechanismSpec(noise_kind=noise_kind)
         mechanism_spec_internal = MechanismSpecInternal(
             mechanism_spec=mechanism_spec,
@@ -210,10 +214,12 @@ class PLDBudgetAccountant:
         if not self._mechanisms:
             return
         if self._total_delta == 0:
-            standard_noise = len(self._mechanisms) / self._total_epsilon
+            dp_params = DifferentialPrivacyParameters(self._total_epsilon, 0)
             self.minimum_noise_std = 0
             for mechanism in self._mechanisms:
-                mechanism_noise_std = mechanism.sensitivity * standard_noise / mechanism.weight
+                min_laplace_noise = accountant.get_smallest_laplace_noise(
+                    dp_params, 1, mechanism.sensitivity)
+                mechanism_noise_std = min_laplace_noise / mechanism.weight
                 mechanism.mechanism_spec._noise_standard_deviation = mechanism_noise_std
                 self.minimum_noise_std = self.minimum_noise_std + mechanism_noise_std
         else:
--- a/pipeline_dp/budget_accounting.py
+++ b/pipeline_dp/budget_accounting.py
@@ -3,10 +3,8 @@
 import logging
 import math
 from dataclasses import dataclass
-from dp_accounting.common import DifferentialPrivacyParameters
 from pipeline_dp.aggregate_params import NoiseKind
 from dp_accounting import privacy_loss_distribution as pldlib
-from dp_accounting import accountant
 
 
 @dataclass
@@ -214,20 +212,17 @@ class PLDBudgetAccountant:
         if not self._mechanisms:
             return
         if self._total_delta == 0:
-            dp_params = DifferentialPrivacyParameters(self._total_epsilon, 0)
-            self.minimum_noise_std = 0
+            sum_weights = 0
             for mechanism in self._mechanisms:
-                min_laplace_noise = accountant.get_smallest_laplace_noise(
-                    dp_params, 1, mechanism.sensitivity)
-                mechanism_noise_std = min_laplace_noise / mechanism.weight
-                mechanism.mechanism_spec._noise_standard_deviation = mechanism_noise_std
-                self.minimum_noise_std = self.minimum_noise_std + mechanism_noise_std
+                sum_weights += mechanism.weight
+            minimum_noise_std = sum_weights / self._total_epsilon * math.sqrt(2)
         else:
             minimum_noise_std = self._find_minimum_noise_std()
-            self.minimum_noise_std = minimum_noise_std
-            for mechanism in self._mechanisms:
-                mechanism_noise_std = mechanism.sensitivity * minimum_noise_std / mechanism.weight
-                mechanism.mechanism_spec._noise_standard_deviation = mechanism_noise_std
+
+        self.minimum_noise_std = minimum_noise_std
+        for mechanism in self._mechanisms:
+            mechanism_noise_std = mechanism.sensitivity * minimum_noise_std / mechanism.weight
+            mechanism.mechanism_spec._noise_standard_deviation = mechanism_noise_std
 
     def _find_minimum_noise_std(self) -> float:
         """Finds the minimum noise which satisfies the total budget.
--- a/requirements.dev.txt
+++ b/requirements.dev.txt
@@ -5,3 +5,4 @@ pyspark
 pytest
 pylint
 yapf
+git+https://github.com/google/differential-privacy.git@327972c1ae710e8cd0a4754fffdd78c3500272ee#subdirectory=python
--- a/tests/budget_accounting_test.py
+++ b/tests/budget_accounting_test.py
@@ -1,8 +1,13 @@
+"""Budget Accounting Test"""
+
 import unittest
-
+from dataclasses import dataclass
 import pipeline_dp
+from pipeline_dp.aggregate_params import NoiseKind
+from pipeline_dp.budget_accounting import MechanismSpec, PLDBudgetAccountant
 
 
+# pylint: disable=protected-access
 class BudgetAccountantTest(unittest.TestCase):
 
     def test_validation(self):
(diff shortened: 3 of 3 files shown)


tests/pld_budget_accounting_test.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def test_optimize_noise_laplace(self):
+        accountant = PLDBudgetAccountant(noise=20,
+                                         total_epsilon=1,
+                                         total_delta=1e-5)
+        for _ in range(0, 4):
+            accountant.request_budget(NoiseKind.LAPLACE)
+            accountant.request_budget(NoiseKind.GAUSSIAN)
+
+        optimal_noise = accountant.optimize_noise_to_delta()
```
jspacek, line 53: This is the problematic test; I have tried adjusting parameters to make it faster, but it is still very slow 😿

pipeline_dp/pld_budget_accounting.py
```diff
@@ -0,0 +1,114 @@
...
+                f"Noise type {noise_kind} is unsupported.")
+        return pldBudget
+
+    def compute_budgets(self):
+        optimal_noise = self.optimize_noise_to_delta()
+        self.noise = optimal_noise  # i set the noise for the entire budget accountant
+
+    def optimize_noise_to_delta(self) -> float:
+        """Adjust noise to get close to given epsilon; increasing noise decreases epsilon."""
+        threshold = 0.001
```
jspacek, line 67: Threshold will be smaller, but set to larger for now because of the tests taking a long time.

pipeline_dp/pld_budget_accounting.py
```diff
@@ -0,0 +1,114 @@
...
+            raise AssertionError("Noise parameter is not yet defined.")
+        return self._noise_parameter
+
+
+@dataclass
+class RequestedPLDBudget:
+    budget: PLDBudget
+
+
+@dataclass
```
dvadym, line 24: I'd say that PLDBudgetAccountant is more like functionality class than data class, so I'd suggest to remove @dataclass.

pipeline_dp/pld_budget_accounting.py
```diff
@@ -0,0 +1,114 @@
...
+class PLDBudgetAccountant:
+    noise: float = 1  # this is where i am storing noise, does this make sense?
+    max_noise: float = 50
+    sensitivity: float = 1
```
dvadym, line 28: sensitivity is the feature of a specific mechanism, meantime PLDBudgetAccountant corresponds to the whole pipeline.
More details below on a comment on request_budget

pipeline_dp/pld_budget_accounting.py
```diff
@@ -0,0 +1,114 @@
...
+                f"Epsilon must be positive, not {self.total_epsilon}.")
+        if self.total_delta is not None and self.total_delta < 0:
+            raise ValueError(
+                f"Delta must be non-negative, not {self.total_delta}.")
+        if self.noise > self.max_noise:
+            raise ValueError(
+                f"Noise = {self.noise} is larger than the maximum = {self.max_noise}."
+            )
+
+    def request_budget(self, noise_kind: NoiseKind) -> PLDBudget:
```
dvadym, line 48: Sensitivity is the part of a mechanism specification, so I'd suggest to replace
   NoiseKind ->  
```
class MechanismSpec:  # @dataclass or maybe namedtuple
   NoiseKind
   sensitivity
   weight
``` 
weight is similar to [weight]( # similar to [weight](https://github.com/OpenMined/PipelineDP/blob/65d00ee6c906a36b8a173ddbb83d7e3bf71905b4/pipeline_dp/budget_accounting.py#L48)) in naive accountant

pipeline_dp/pld_budget_accounting.py
```diff
@@ -0,0 +1,114 @@
...
+        pldBudget = PLDBudget()
```
dvadym, line 49: Nit (=style imprpovement): pldBudget -> pld_budget

pipeline_dp/pld_budget_accounting.py
dvadym: As you mentioned in comments the bottleneck is in creating of PLDs. I fixed that ~1 month ago, it looks like pip packages wasn't released. You can use [dp_accounting from github](https://github.com/google/differential-privacy/tree/main/python) for now (I'll ask for releasing a new version of pip package)

pipeline_dp/pld_budget_accounting.py
```diff
@@ -0,0 +1,114 @@
...
+        self.noise = optimal_noise  # i set the noise for the entire budget accountant
+
+    def optimize_noise_to_delta(self) -> float:
+        """Adjust noise to get close to given epsilon; increasing noise decreases epsilon."""
+        threshold = 0.001
+        optimal_noise = self.noise
+        pld_laplace_gaussian = self.compose_distributions(optimal_noise)
+        pld_epsilon = pld_laplace_gaussian.get_epsilon_for_delta(
+            self.total_delta)
+        (min, max) = (0, self.max_noise)
```
dvadym, line 72: Nit: remove ( )
Nit: replace min/max, to something like low/high (or other similar names). 

That's because min/max are built-in functions, it's better not to hide built-in functions.

pipeline_dp/pld_budget_accounting.py
```diff
@@ -0,0 +1,114 @@
...
+        return pldBudget
+
+    def compute_budgets(self):
+        optimal_noise = self.optimize_noise_to_delta()
+        self.noise = optimal_noise  # i set the noise for the entire budget accountant
+
+    def optimize_noise_to_delta(self) -> float:
+        """Adjust noise to get close to given epsilon; increasing noise decreases epsilon."""
+        threshold = 0.001
+        optimal_noise = self.noise
```
dvadym, line 68: maybe current_noise? It's not yet optimal ))

pipeline_dp/pld_budget_accounting.py
```diff
@@ -0,0 +1,114 @@
...
+        pld_laplace_gaussian = self.compose_distributions(optimal_noise)
+        pld_epsilon = pld_laplace_gaussian.get_epsilon_for_delta(
+            self.total_delta)
+        (min, max) = (0, self.max_noise)
+
+        while (min <= max):
+            if abs(pld_epsilon - self.total_epsilon) <= threshold:
+                return optimal_noise
```
dvadym, line 76: since we'd like to be on safe side, it's better to return "max" (which is definetely ensures that pld_eps wouldn't be more that total_epsilon)

pipeline_dp/pld_budget_accounting.py
```diff
@@ -0,0 +1,114 @@
...
+    def compose_distributions(self, noise: float):
+        pld_laplace = self.compose_laplace(noise)
+        pld_gaussian = self.compose_gaussian(noise)
+        if pld_gaussian is None:
+            return pld_laplace
+        if pld_gaussian is None:
+            return pld_gaussian
+        return pld_laplace.compose(pld_gaussian)
+
+    def compose_laplace(self, noise: float):
```
dvadym, line 96: the difference is in compose_laplace and compose_gaussian is only in creating. And it's not necessary to compose laplace and gaussians separately, so I'd suggest:
1.Have only 1 var _requested_budgets 
2.To do composition in one loop 

Clarification: In such way we would remove code duplication and simplify code. Moreover in future we will have more mechanisms and so it's better to keep all of them in one variable _requested_budgets.

dvadym (commented):
Thanks, it looks great! I left some comments and suggestions for improvements

pipeline_dp/pld_budget_accounting.py
jspacek: 👍  Fixed to loop around a single collection, thank you!

pipeline_dp/pld_budget_accounting.py
jspacek: 👍

pipeline_dp/pld_budget_accounting.py
jspacek: It is not 😄 , is `adjusted_noise` OK?

pipeline_dp/pld_budget_accounting.py
jspacek: 👍  Changed, thanks!

pipeline_dp/pld_budget_accounting.py
jspacek: 👍  Camel cases removed 🐫  Thanks!

pipeline_dp/pld_budget_accounting.py
jspacek: I believe it is using weight correctly now:

`pld.PrivacyLossDistribution.from_laplace_mechanism(
                    noise / mechanism.weight, sensitivity=mechanism.sensitivity)`

pipeline_dp/pld_budget_accounting.py
jspacek: I think I have this correctly stored in the mechanism object:

`pld.PrivacyLossDistribution.from_laplace_mechanism(
                    noise / mechanism.weight, sensitivity=mechanism.sensitivity)`

pipeline_dp/pld_budget_accounting.py
jspacek: Makes sense, removed 👍

pipeline_dp/pld_budget_accounting.py
jspacek: Thanks, it is super fast now! I was able to pip install from github 🚀

pipeline_dp/pld_budget_accounting.py
```diff
@@ -0,0 +1,143 @@
...
+        Returns:
+            The noise value adjusted for the given epsilon.
+        """
+        threshold = 1e-4
+        adjusted_noise = self._max_noise / 2
+        pld_laplace_gaussian = self.compose_distributions(adjusted_noise)
+        pld_epsilon = pld_laplace_gaussian.get_epsilon_for_delta(
+            self._total_delta)
+        low, high = 0, self._max_noise
+        while low <= high:
```
dvadym, line 108: Nit: change condition
  low + threshold < high

and then no need in check inside loop, and to make return in the end of the function

It's slightly better:
  1.More linear - no break from from loop and return is in the end of the function
  2.It's better to make condition on low, high, not on eps, there is a chance that pld_epsilon is close to the total_epsilon, but low and high are far apart, so the noise (i.e. high) might be not precise.

pipeline_dp/pld_budget_accounting.py
```diff
@@ -0,0 +1,143 @@
...
+            if abs(pld_epsilon - self._total_epsilon) <= threshold:
+                return high
+            elif pld_epsilon < self._total_epsilon:
+                high = adjusted_noise
+            elif pld_epsilon > self._total_epsilon:
+                low = adjusted_noise
+            adjusted_noise = ((high - low) / 2) + low
+            pld_laplace_gaussian = self.compose_distributions(adjusted_noise)
```
dvadym, line 116: Nit: move compoting of epsilon to the top of the loop, then no need to do computation of the eps before the loop (so it would be slightly less code)

requirements.dev.txt
```diff
@@ -5,3 +5,4 @@ pyspark
 pytest
 pylint
 yapf
+git+https://github.com/google/differential-privacy.git@327972c1ae710e8cd0a4754fffdd78c3500272ee#subdirectory=python
```
dvadym, line 8: Wow, I didn't know that it's possible to make dependencies on github repos

dvadym (commented):
| Functionality that gives the user two options
| Input: Noise and Delta, Output: PLD epsilon
| Input: Epsilon and Delta, Output: Minimized noise

For now let's do only "Epsilon and Delta, Output: Minimized noise". 

**Explanation**: "Input: Noise and Delta, Output: PLD epsilon" is also a reasonable option, but for now the idea, that users specify the privacy budget per pipeline and BudgetAccountant finds the minimum suitable noise (i.e. "splits" the budget for the user).

pipeline_dp/pld_budget_accounting.py
```diff
@@ -0,0 +1,143 @@
...
+from dataclasses import dataclass
+from pipeline_dp.aggregate_params import NoiseKind
+from dp_accounting import privacy_loss_distribution as pld
+
+
+@dataclass
+class MechanismSpec:
+    """Specifies the parameters for a mechanism.
+
+    NoiseKind is a mandatory parameter needed to define the distribution.
```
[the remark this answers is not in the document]
jspacek: Fixed! 👍

pipeline_dp/pld_budget_accounting.py
```diff
@@ -0,0 +1,143 @@
...
+            elif pld_epsilon < self._total_epsilon:
+                high = adjusted_noise
+            elif pld_epsilon > self._total_epsilon:
+                low = adjusted_noise
+            adjusted_noise = ((high - low) / 2) + low
+            pld_laplace_gaussian = self.compose_distributions(adjusted_noise)
+            pld_epsilon = pld_laplace_gaussian.get_epsilon_for_delta(
+                self._total_delta)
+
+    def compose_distributions(self,
```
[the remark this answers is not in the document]
jspacek: Changed all private functions to use `_` convention. Thanks!

pipeline_dp/pld_budget_accounting.py
```diff
@@ -0,0 +1,143 @@
...
+        Repeat the composition until the noise value gives a
+        new epsilon close to the given epsilon (within a threshold).
+        By increasing the noise we can decrease the epsilon.
+
+        Returns:
+            The noise value adjusted for the given epsilon.
+        """
+        threshold = 1e-4
+        adjusted_noise = self._max_noise / 2
+        pld_laplace_gaussian = self.compose_distributions(adjusted_noise)
```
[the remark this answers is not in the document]
jspacek: Very long name 😄  Fixed 👍

pipeline_dp/pld_budget_accounting.py
```diff
@@ -0,0 +1,143 @@
...
+
+        Repeat the composition until the noise value gives a
+        new epsilon close to the given epsilon (within a threshold).
+        By increasing the noise we can decrease the epsilon.
+
+        Returns:
+            The noise value adjusted for the given epsilon.
+        """
+        threshold = 1e-4
+        adjusted_noise = self._max_noise / 2
```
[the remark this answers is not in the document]
jspacek: New function `_calculate_max_noise` should do this correctly 👀  Thank you!

pipeline_dp/pld_budget_accounting.py
jspacek: Thank you, I think this loop is much improved now with your suggestions! 👀

pipeline_dp/pld_budget_accounting.py
jspacek: Removed the extra computations after fixing the loop logic 👍  Thanks!

requirements.dev.txt
jspacek: 🎉  A little tricky to tell it where to find the `setup.py` file, but otherwise easy!

pipeline_dp/pld_budget_accounting.py
```diff
@@ -0,0 +1,143 @@
...
+            noise_kind: The kind of distribution for the mechanism.
+            sensitivity: The sensitivity for the mechanism.
+            weight: The weight for the mechanism.
+
+        Returns:
+            A mechanism spec based on the parameters.
+        """
+        mechanism_spec = MechanismSpec(noise_kind, sensitivity, weight)
+        self._mechanisms.append(mechanism_spec)
+        return mechanism_spec
```
dvadym, line 68: The idea of the usage of this class is:

mechanism1 = pld_accountant->request_budget(..)
mechanism2 = pld_accountant->request_budget(..)
...
pld_accountant.compute_budgets()  # computes mechanism1, mechanism2 noise parameters

So this function should return some "lazy" object that represents not yet computed noise, for example MechanismNoise with one property noise.

Here is an example of such object for Naive budget accountant
https://github.com/OpenMined/PipelineDP/blob/65d00ee6c906a36b8a173ddbb83d7e3bf71905b4/pipeline_dp/budget_accounting.py#L24

pipeline_dp/pld_budget_accounting.py
```diff
@@ -0,0 +1,143 @@
...
+        """Computes the budget for the pipeline.
+
+        Composes the mechanisms and adjusts the amount of
+        noise based on given epsilon. Sets the noise for the
+        entire pipeline.
+        """
+        if len(self._mechanisms) == 0:
+            return
+        adjusted_noise = self._compute_and_adjust()
+        self._noise = adjusted_noise
```
dvadym, line 80: This function also should update MechanismNoise objects

(more details about those object comment to request_budget above)

Here is an example in BudgetAccountant
https://github.com/OpenMined/PipelineDP/blob/65d00ee6c906a36b8a173ddbb83d7e3bf71905b4/pipeline_dp/budget_accounting.py#L106

dvadym (commented):
Thanks! I left some comments

pipeline_dp/budget_accounting.py
```diff
@@ -100,7 +100,182 @@ def compute_budgets(self):
         for requested_budget in self._requested_budgets:
             eps = delta = 0
             if total_weight_eps:
-                eps = requested_budget.use_eps * self._eps * requested_budget.weight / total_weight_eps
+                numerator = requested_budget.use_eps * self._eps * requested_budget.weight
+                eps = numerator / total_weight_eps
             if total_weight_delta:
-                delta = requested_budget.use_delta * self._delta * requested_budget.weight / total_weight_delta
+                numerator = requested_budget.use_delta * self._delta * requested_budget.weight
```
jspacek, line 106: 👋  Changing a couple of things in this file due to failing lint (lines too long)

pipeline_dp/budget_accounting.py
```diff
@@ -100,7 +100,182 @@ def compute_budgets(self):
...
+        entire pipeline.
+        """
+        if not self._mechanisms:
+            return
+        minimum_noise = self._compute_and_adjust()
+        self.minimum_noise = minimum_noise
+        for mechanism in self._mechanisms:
+            mechanism._noise = minimum_noise / mechanism.weight
+
+    def _compute_and_adjust(self) -> float:
```
dvadym, line 201: Maybe something like
_find_minimum_noise ?

pipeline_dp/budget_accounting.py
```diff
@@ -100,7 +100,182 @@ def compute_budgets(self):
...
+        """ Adjust the noise value to the minimum.
+
+        Repeat the PLD composition until the noise value gives a
```
dvadym, line 204: Nit: Repeat the PLD composition ->  Uses binary search to ...   (it's more precise)

dvadym (commented):
Thanks it looks great! I left some comments, mostly about comments :)

pipeline_dp/budget_accounting.py
```diff
@@ -100,7 +101,188 @@ def compute_budgets(self):
...
+        """
+        if self._noise_standard_deviation is None:
+            raise AssertionError(
+                "Noise standard deviation is not calculated yet.")
+        return self._noise_standard_deviation
+
+
+@dataclass
+class MechanismSpecInternal:
+    """Stores sensitivity and weight not exposed in MechanismSpec."""
```
dvadym, line 137: Nit: Please remove default arguments, anyway all those values should be provided in places where this class is used (they have default values in request_budget)

pipeline_dp/budget_accounting.py
```diff
@@ -100,7 +100,182 @@ def compute_budgets(self):
...
+                eps = numerator / total_weight_eps
             if total_weight_delta:
-                delta = requested_budget.use_delta * self._delta * requested_budget.weight / total_weight_delta
+                numerator = requested_budget.use_delta * self._delta * requested_budget.weight
+                delta = numerator / total_weight_delta
             requested_budget.budget.set_eps_delta(eps, delta)
+
+
+@dataclass
+class MechanismSpec:
```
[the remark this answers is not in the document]
dvadym: Thanks, that's great!

pipeline_dp/budget_accounting.py
```diff
@@ -100,7 +101,188 @@ def compute_budgets(self):
...
+
+        _validate_epsilon_delta(total_epsilon, total_delta)
+
+        self._total_epsilon = total_epsilon
+        self._total_delta = total_delta
+        self._mechanisms = []
+        self.minimum_noise_std = None
+        self._pld_discretization = pld_discretization
+
+    def request_budget(self, noise_kind: NoiseKind) -> MechanismSpec:
```
dvadym, line 175: Please add sensitivity and weight as arguments (with default values = 1) and pass them to `MechanismSpecInternal`

pipeline_dp/budget_accounting.py
```diff
@@ -100,7 +101,188 @@ def compute_budgets(self):
...
+        """
+        if not self._mechanisms:
+            return
+        minimum_noise_std = self._find_minimum_noise_std()
+        self.minimum_noise_std = minimum_noise_std
+        for mechanism in self._mechanisms:
+            mechanism.mechanism_spec._noise_standard_deviation = minimum_noise_std / mechanism.weight
+
+    def _find_minimum_noise_std(self) -> float:
+        """Find a noise value to the minimum.
```
dvadym, line 209: Nit: Finds the minimum noise which satisfies the total budget.

pipeline_dp/budget_accounting.py
```diff
@@ -100,7 +101,188 @@ def compute_budgets(self):
...
+            pld_epsilon = pld.get_epsilon_for_delta(self._total_delta)
+            if pld_epsilon <= self._total_epsilon:
+                high = mid
+            elif pld_epsilon > self._total_epsilon:
+                low = mid
+
+        return high
+
+    def _calculate_max_noise_std(self) -> float:
+        """Calculate maximum noise std for given epsilon.
```
dvadym, line 233: Nit: Calculate**s** an upper bound for the noise to satisfies the budget.

Explanation: the budget is both total_epsilon and total_delta

pipeline_dp/budget_accounting.py
```diff
@@ -100,7 +101,188 @@ def compute_budgets(self):
...
+
+        Returns:
```
dvadym, line 235: It looks like Returns section doesn't provide additional infomation to function description. Please drop it.

pipeline_dp/budget_accounting.py
```diff
@@ -100,7 +101,188 @@ def compute_budgets(self):
...
+            noise_std: The noise of the distributions to construct.
+
+        Returns:
+            A PrivacyLossDistribution object for the pipeline.
+        """
+        composed, pld = None, None
+
+        for mechanism_spec_internal in self._mechanisms:
+            if mechanism_spec_internal.mechanism_spec.noise_kind == NoiseKind.LAPLACE:
+                pld = pldlib.PrivacyLossDistribution.from_laplace_mechanism(
```
dvadym, line 260: Please add a comment before:
\# The Laplace distribution parameter = std/sqrt(2).

tests/budget_accounting_test.py
```diff
@@ -48,5 +52,102 @@ def test_compute_budgets(self):
...
+        expected_noise = 9.87554931640625
+        epsilon = 2.4857963874870936
+        delta = 1e-5
+        accountant = PLDBudgetAccountant(epsilon, delta, 1e-2)
+        for _ in range(0, 4):
+            accountant.request_budget(NoiseKind.LAPLACE)
+            accountant.request_budget(NoiseKind.GAUSSIAN)
+
+        for mechanism_internal in accountant._mechanisms:
+            mechanism_internal.weight = 2
```
dvadym, line 123: Please pass weights in request_budget

tests/budget_accounting_test.py
```diff
@@ -48,5 +52,102 @@ def test_compute_budgets(self):
...
+        expected_noise = 9.8876953125
+        epsilon = 2.4838584185571655
+        delta = 1e-5
+        accountant = PLDBudgetAccountant(epsilon, delta, 1e-2)
+        for _ in range(0, 4):
+            accountant.request_budget(NoiseKind.LAPLACE)
+            accountant.request_budget(NoiseKind.GAUSSIAN)
+
+        for mechanism_internal in accountant._mechanisms:
+            mechanism_internal.sensitivity = 2
```
dvadym, line 142: Please test with different sensitivities and weights for different mechanisms as well

dvadym (commented):
Thanks it looks great. Some minor comments.

pipeline_dp/budget_accounting.py
```diff
@@ -100,7 +101,192 @@ def compute_budgets(self):
...
+    """Specifies the parameters for a mechanism.
+
+    NoiseKind defines the kind of noise distribution.
+    noise is the minimized noise standard deviation.
+    """
+    noise_kind: NoiseKind
+    _noise_standard_deviation: float = None
+
+    @property
+    def noise(self):
```
dvadym, line 123: Nit: noise_standard_deviation

tests/budget_accounting_test.py
```diff
@@ -48,5 +53,147 @@ def test_compute_budgets(self):
...
+
+        @dataclass
+        class ComputeBudgetTestCase:
+            name: str
+            expected_epsilon: float
+            delta: float
+            expected_pipeline_noise_std: float
+            mechanisms: []
+
+        testcases = [
```
jspacek, line 94: @dvadym I thought it might be nice to try table driven tests in Python, but perhaps I got a little carried away and it is too complex, if so let me know and I will revert to the old style. Thank you!

tests/budget_accounting_test.py
```diff
@@ -48,5 +53,147 @@ def test_compute_budgets(self):
...
+            count: int
+            expected_noise_std: float
+            noise_kind: NoiseKind
+            weight: float
+            sensitivity: float
+
+        @dataclass
+        class ComputeBudgetTestCase:
+            name: str
+            expected_epsilon: float
```
dvadym, line 89: Nit: epsilon
It looks like not expected_eps, but the input epsilon of the BudgetAccountant

tests/budget_accounting_test.py
```diff
@@ -48,5 +53,147 @@ def test_compute_budgets(self):
...
+            delta: float
+            expected_pipeline_noise_std: float
+            mechanisms: []
+
+        testcases = [
+            ComputeBudgetTestCase(name="laplace_mechanisms",
```
dvadym, line 95: Thanks for having tests in table, I realized that we need to cover additional cases.

Please add 2 additional tests, with `delta = 0`
1. 2 Laplace mechanisms, sensitivity=1, weight = 1 (with delta = 0, PLD should coincide with a standard eps-dp arithmetic)
2. 1 Gaussian mechanism (delta = 0, can't be Gaussian, so compute_budgets should raise Exception

tests/budget_accounting_test.py
dvadym: Thanks, I like this approach!

tests/budget_accounting_test.py
```diff
@@ -48,5 +53,147 @@ def test_compute_budgets(self):
...
+                case.expected_epsilon, actual_epsilon, 3,
+                f"failed test {case.name} expected epsilon {case.expected_epsilon} got {actual_epsilon}"
+            )
+            accountant.compute_budgets()
+            self.assertEqual(
+                case.expected_pipeline_noise_std, accountant.minimum_noise_std,
+                f"failed test {case.name} expected pipeline noise {case.expected_pipeline_noise_std} "
+                f"got {accountant.minimum_noise_std}")
+            for actual_mechanism_tuple in actual_mechanisms:
+                expected_mechanism_noise_std = actual_mechanism_tuple[0]
```
dvadym, line 189: Nit: 
expected_mechanism_noise_std, actual_mechanism = actual_mechanism_tuple

tests/budget_accounting_test.py
```diff
@@ -48,5 +53,147 @@ def test_compute_budgets(self):
...
+                        (mechanism.expected_noise_std,
+                         accountant.request_budget(
+                             mechanism.noise_kind,
+                             weight=mechanism.weight,
+                             sensitivity=mechanism.sensitivity)))
+            self.assertEqual(
+                len(actual_mechanisms), len(accountant._mechanisms),
+                f"failed test {case.name} expected len {len(actual_mechanisms)} "
+                f"got len {len(accountant._mechanisms)}")
+            compare_pld = accountant._compose_distributions(
```
dvadym, line 176: I like that approach to test _compose_distributions and compute_budgets on the same data

dvadym (commented):
Thanks it looks great! I realized that we need a few more tests :) (details are in comments)

tests/budget_accounting_test.py
jspacek: @dvadym I am not sure I handled correctly this calculation to take into account the weight and sensitivities per mechanism. I added a mention for the calculation up above.

Thanks for catching these cases!

pipeline_dp/budget_accounting.py
```diff
@@ -100,7 +101,205 @@ def compute_budgets(self):
...
+        """Computes the budget for the pipeline.
+
+        Composes the mechanisms and adjusts the amount of
+        noise based on given epsilon. Sets the noise for the
+        entire pipeline.
+        """
+        if not self._mechanisms:
+            return
+        if self._total_delta == 0:
+            standard_noise = len(self._mechanisms) / self._total_epsilon
```
jspacek, line 213: @dvadym This is the new calculation for standard Laplace (epsilon, 0)-DP. I am not sure that it handles sensitivity and weight per mechanism correctly. Thank you for taking a look!

dvadym: I think PLD library should handle the case of delta=0 (in case only laplacian mechanisms are requested), please try w/o this code and to see whether it works.

If not, we can submit this CL w/o processing of delta=0, I'll fix the library and then we can process delta=0 case with PLD.

pipeline_dp/budget_accounting.py
```diff
@@ -100,7 +101,205 @@ def compute_budgets(self):
...
+            noise_kind: The kind of noise distribution for the mechanism.
+            sensitivity: The sensitivity for the mechanism.
+            weight: The weight for the mechanism.
+
+        Returns:
+            A "lazy" mechanism spec object that doesn't contain the noise
+            standard deviation until compute_budgets is called.
+        """
+        if noise_kind == NoiseKind.GAUSSIAN and self._total_delta == 0:
+            raise AssertionError("Gaussian delta value must be greater than 0")
```
dvadym, line 194: Nit: "The Gaussian mechanism requires the pipeline delta ..."

Explanation: it's slightly more precise and suggests the API user about "the pipeline delta"

pipeline_dp/budget_accounting.py
jspacek: 👋 @dvadym It fails in `privacy_loss_distribution.py` file in `def get_epsilon_for_delta(self, delta: float) -> float:` because it returns infinity value. The example for LaPlace retrieves the delta for epsilon, but then the delta is not 0, so perhaps this is not what we want https://github.com/google/differential-privacy/blob/main/python/dp_accounting/privacy_loss_distribution_basic_example.py#L40 ?

I changed to use `get_smallest_laplace_noise` function from https://github.com/google/differential-privacy/blob/main/python/dp_accounting/accountant.py#L63. I'm not sure this will work as the function takes one sensitivity overall, and we want a sensitivity per mechanism. 

Please let me know if I should remove the delta=0 cases and tests and we can change it when the PLD library has a fix. Thank you!

pipeline_dp/budget_accounting.py
```diff
@@ -100,7 +103,209 @@ def compute_budgets(self):
...
+        """Computes the budget for the pipeline.
+
+        Composes the mechanisms and adjusts the amount of
+        noise based on given epsilon. Sets the noise for the
+        entire pipeline.
+        """
+        if not self._mechanisms:
+            return
+        if self._total_delta == 0:
+            dp_params = DifferentialPrivacyParameters(self._total_epsilon, 0)
```
dvadym, line 217: Sry, my previous comments where from assumptions that the dp_accounting library can handle that, I checked  it can't. In principle it's possible, but for now I think the easiest way is to do computation w/o the library at all for the case delta = 0.

The easiest approach is 
```
if delta == 0:
    minimum_noise_std = sum_weights/eps* sqrt(2) 
else:
     minimum_noise_std = self._find_minimum_noise_std()
<all code stays as now>
```

and then 

 **Explanation**:
1. That's because `laplace_paramer = sensitivity/eps` . No need to multiply on sensitivity here because we multiple on `sensitivity` when we compute `mechanism_noise_std` below. Sum_weights are needed in order to normalize.
2. sqrt(2) is needed in order to convert from Laplace parameter to standard deviation

I think all tests with delta=0 will require multiplication on sqrt(2).

Does it make sense?

dvadym: Sry, one more comment :)

jspacek: @dvadym No worries! Thank you for the reviews, I am learning a lot!!

1) I summed up all of the mechanism weights but I wonder, should these be included in the sum if they are equal to 1? Because we use default weight value = 1 for not-weighting. 

Sorry another question 😄 

Say we have 3 mechanisms all with weight=1, then for each mechanism we have 
`noise = sensitivity * ( sum_weights=3 / epsilon * sqrt(2) ) / mechanism.weight = 1` with `3` left over in the numerator. I think we do not want to have this constant factor 3 here.

If we have 3 mechanisms all with weight=2 then I think it works as intended to cancel out (normalize)?

`noise = sensitivity * ( sum_weights=6 / epsilon * sqrt(2) ) / mechanism.weight = 2` also with `3` left over in the numerator.

I could add a check to only add to the sum_weights if mechanism.weight > 1?

2) 👍 Makes sense to remove the constant 2 value from the variance!

dvadym (approved):

pipeline_dp/budget_accounting.py
dvadym: If weight = 1 everything is correct. It's basically normal/"naive" budget splitting. 

Assume that we the budget eps and we have 3 Laplace mechanisms and we'd like to split budget evenly between them (i.e. weights are the same in our terminology).

Then each mechansims would have eps/3 budget, so the Laplace parameters would be 

`sensitivy/(eps/3) = 3*sensitivity/eps`

When we compute using weight, it would be

`sensitivity*(sum_weight=3)/eps = 3*sensitivity/eps`
 
i.e. those computations are equivalent. We can consider that 3 factor as budget split :).

dvadym: Thanks for implementing! It was a very interesting code review!

jspacek: I see, it is part of the budget split! Thank you for clarifying and for the reviews!!

Issue #36, opened 2021-06-02
Implement DP computing of count, sum, mean, var


dvadym:
## Goals

Implement 

```
@dataclass
class MeanVarParams:
    eps: float
    delta: float
    low: float
    high: float
    sensitivity: float              # L1 for Laplace, L2 for Gaussian
    noise_kind: NoiseKind   # Laplace or Gaussian

def compute_dp_count(count, dp_parameters) -> dp_count

def compute_dp_sum(sum, dp_parameters) -> dp_sum

def compute_dp_mean(count, sum, dp_parameters) -> (dp_count, dp_sum, dp_mean)

def compute_dp_var(count, sum, sum_squares, dp_parameters) -> (dp_count, dp_sum, dp_mean, dp_var)

```

### DP algorithm
The main way of anonymizing is adding Laplace or Gaussian noise to non-anonymized values. So computing of dp mean might be

1. Add the random noise to count and sum -> dp_count, dp_sum
2. Compute dp_mean = dp_sum/dp_count

[More details](https://docs.google.com/document/d/1cvfAU4XwglRgLc7XP9qCf5a6iZRajqVVsH7s5URY-UU/edit#bookmark=id.9qngddwqrcef) are in the reference doc.

Note: that there are some additional ways to make utility better, so the final algorithm will be slightly different.

### Secure noise implementation
The standard random sampling does not provide privacy. More details are in the [paper](https://crysp.uwaterloo.ca/courses/pet/F18/cache/Mironov.pdf).

 There is a separate [issue](https://github.com/OpenMined/PipelineDP/issues/11) to make available [PyDP wrappers](https://github.com/OpenMined/PyDP) of secure noise from [Google C++ DP library](https://github.com/google/differential-privacy/tree/main/cc/algorithms). For this issue for the development we can use numpy noise np.random.laplace and np.random.normal. We willreplace with the secure as soon as it is ready.

### Formulas for noise parameters
They might be taken from a [book](https://www.cis.upenn.edu/~aaroth/Papers/privacybook.pdf) 

1. Laplace: Def 3.3, page 32
2. Gaussian: Theorem 3.22, page 53 (it’s not an optimal formula, but it’s simple, let’s use it for now).

AlexandraImbrisca:
I would like to work on this issue if that's okay 👋🏻

dvadym:
Sure, thanks! 

In case of any questions, please let me know.

dvadym:
FYI: I've added 
[...]

PR #39, opened 2021-06-08
[WIP] DP computing of count, sum, mean, var

AlexandraImbrisca:
## Description
Closes [#36](https://github.com/OpenMined/PipelineDP/issues/36).

## Affected Dependencies
None.

## How has this been tested?
Tests will be later included.

## Checklist

### DIFF
--- a/pipeline_dp/mean_var_params.py
+++ b/pipeline_dp/mean_var_params.py
@@ -1,0 +1,117 @@
+"""DP computing of count, sum, mean, variance."""
+
+import numpy as np
+
+from dataclasses import dataclass
+from pipeline_dp.aggregate_params import NoiseKind
+
+
+@dataclass
+class MeanVarParams:
+	"""Manages the parameters used for computing the mean/variance."""
+	eps: float
+	delta: float
+	low: float
+	high: float
+	sensitivity: float  # L1 for Laplace, L2 for Gaussian
+	noise_kind: NoiseKind  # Laplace or Gaussian
+
+
+def _apply_laplace_mechanism(value, eps, sensitivity):
+	"""Applies the Laplace mechanism.
+
+	Args:
+		value: The result of querying the database.
+		eps: Parameter of (epsilon, delta)-differential privacy.
+		sensitivity: The L1-sensitivity of the query.
+	"""
+	return value + np.random.laplace(0, sensitivity / eps)
+
+
+def _apply_gaussian_mechanism(value, eps, delta, sensitivity):
+	"""Applies the Gaussian mechanism.
+
+	Args:
+		value: The result of querying the database.
+		eps, delta: Parameters of (epsilon, delta)-differential privacy.
+		sensitivity: The L2-sensitivity of the query.
+	"""
+	return value + np.random.normal(0, np.var(eps, delta, sensitivity))
+
+
+def _add_random_noise(value, dp_params):
+	"""Adds random noise to the value according to the given parameters.
+
+	Args:
+		value: Random noise is added to this value.
+		dp_params: The parameters used at computing the noise.
+
+	Raises:
+		ValueError: The noise kind is invalid.
+	"""
+	if dp_params.noise_kind == NoiseKind.LAPLACE:
+		return _apply_laplace_mechanism(value, dp_params.eps, dp_params.sensitivity)
+	if dp_params.noise_kind == NoiseKind.GAUSSIAN:
+		return _apply_gaussian_mechanism(value, dp_params.eps, dp_params.delta, dp_params.sensitivity)
+	raise ValueError("Noise kind must be either Laplace or Gaussian.")
+
+
+def compute_dp_count(count: int, dp_params: MeanVarParams):
+	"""Computes DP count.
+
+	Args:
+		count: Non-DP count.
+		dp_params: The parameters used at computing the noise.
+
+	Raises:
+		ValueError: The noise kind is invalid.
+	"""
+	return _add_random_noise(count, dp_params)
+
+
+def compute_dp_sum(sum: float, dp_params: MeanVarParams):
+	"""Computes DP sum.
+
+	Args:
+		sum: Non-DP sum.
+		dp_params: The parameters used at computing the noise.
+
+	Raises:
+		ValueError: The noise kind is invalid.
+	"""
+	return _add_random_noise(sum, dp_params)
+
+
+def compute_dp_mean(count: int, sum: float, dp_params: MeanVarParams):
+	"""Computes DP mean.
+
+	Args:
+		count: Non-DP count.
+		sum: Non-DP sum.
+		dp_params: The parameters used at computing the noise.
+
+	Raises:
+		ValueError: The noise kind is invalid.
+	"""
+	dp_count = compute_dp_count(count, dp_params)
+	dp_sum = compute_dp_sum(sum, dp_params)
+	return dp_count, dp_sum, dp_sum / dp_count
+
+
+def compute_dp_var(count: int, sum: float, sum_squares: float, dp_params: MeanVarParams):
+	"""Computes DP variance.
+
+	Args:
+		count: Non-DP count.
+		sum: Non-DP sum.
+		sum_squares: Non-DP sum of squares.
+		dp_params: The parameters used at computing the noise.
+
+	Raises:
+		ValueError: The noise kind is invalid.
+	"""
+	dp_count = compute_dp_count(count, dp_params)
+	dp_sum = compute_dp_sum(sum, dp_params)
+	dp_sum_squares = compute_dp_sum(sum_squares, dp_params)
+	dp_mean = compute_dp_mean(count, sum, dp_params)
+	return dp_count, dp_sum, dp_mean, dp_sum_squares / dp_count - np.power(dp_mean, 2)


pipeline_dp/mean_var_params.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+@dataclass
+class MeanVarParams:
+	"""Manages the parameters used for computing the mean/variance."""
+	eps: float
+	delta: float
+	low: float
+	high: float
+	sensitivity: float  # L1 for Laplace, L2 for Gaussian
```
dvadym, line 16: I thought again, and it's probably better to replace sensitivity with

**max_partitions_contributed**
**max_contributions_per_partition**

and then to compute sensitivity in this file using formulas from the design doc
https://docs.google.com/document/d/1cvfAU4XwglRgLc7XP9qCf5a6iZRajqVVsH7s5URY-UU/edit#bookmark=id.vpk6kpu64m5u

In that formula:
s0 is max_partitions_contributed
s_infinity is max_contributions_per_partition

pipeline_dp/mean_var_params.py
```diff
@@ -0,0 +1,117 @@
...
+
+def _apply_gaussian_mechanism(value, eps, delta, sensitivity):
+	"""Applies the Gaussian mechanism.
+
+	Args:
+		value: The result of querying the database.
+		eps, delta: Parameters of (epsilon, delta)-differential privacy.
+		sensitivity: The L2-sensitivity of the query.
+	"""
+	return value + np.random.normal(0, np.var(eps, delta, sensitivity))
```
dvadym, line 39: is np.var supposed to be a function that computes variance of the noise by the formula from the "DP book??

dvadym (commented):
Thanks, it's a good start!

Let's for now focus only on count and sum. There are some tricks how to compute mean and var with a better utility. 

I left some comments. 

The next steps, please
1. update MeanVarParams as in the comments
2.Introduce functions for computing L1, L2 sensitivities (more details in comments)
3.Call those functions in compute_dp_count and compute_dp_sum

And if you have any questions, don't hesitate to ask!

pipeline_dp/mean_var_params.py
AlexandraImbrisca: Actually no, I replaced it with the formula from the book.

pipeline_dp/mean_var_params.py
```diff
@@ -0,0 +1,123 @@
...
+
+def _apply_gaussian_mechanism(value, dp_params):
+	"""Applies the Gaussian mechanism.
+
+	Args:
+		value: The result of querying the database.
+		dp_params: The parameters used at computing the noise.
+	"""
+	l2_sensitivity = np.sqrt(dp_params.max_partitions_contributed) * dp_params.max_contributions_per_partition
+	sigma = np.sqrt(2 * np.log(1.25 / dp_params.delta)) * l2_sensitivity / dp_params.eps
```
dvadym, line 42: Please 
1.extract computing sigma to a separate function (it would be easier then to replace with another formula)
2.add a link to the source of this formula (i.e. link to the book and theorem number)

pipeline_dp/mean_var_params.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+def _apply_laplace_mechanism(value, dp_params):
+	"""Applies the Laplace mechanism.
+
+	Args:
+		value: The result of querying the database.
+		dp_params: The parameters used at computing the noise.
+	"""
+	l1_sensitivity = dp_params.max_partitions_contributed * dp_params.max_contributions_per_partition
```
dvadym, line 28: Sensitivity depends on the type of aggregation. Namely sensitivity is a maximum change by adding/removing one user. This formula works when each contribution is 1 (like in count). For sum, each contribution is within [low, high], so the maximum contribution is max(|low|, |high|). 

Please please add methods for computation l1_sensitivity/l2_sensitiviy to MeanVarParams  (like count_l1_sensitivity, sum_l1_sensitivity, count_l2_sensitivity, sum_l2_sensitivity)

pipeline_dp/mean_var_params.py
```diff
@@ -0,0 +1,123 @@
...
+	eps: float
+	delta: float
+	low: float
+	high: float
+	max_partitions_contributed: int
+	max_contributions_per_partition: int
+	noise_kind: pipeline_dp.NoiseKind  # Laplace or Gaussian
+
+
+def _apply_laplace_mechanism(value, dp_params):
```
dvadym, line 21: Sensitivity also depends on type of aggregation. Pease replace dp_params -> l1_sensitivity and compute sensitivity in compute_dp_count, compute_dp_sum.

pipeline_dp/mean_var_params.py
```diff
@@ -0,0 +1,123 @@
...
+		value: The result of querying the database.
+		dp_params: The parameters used at computing the noise.
+	"""
+	l1_sensitivity = dp_params.max_partitions_contributed * dp_params.max_contributions_per_partition
+
+	# TODO: use the secure noise instead of np.random
+	return value + np.random.laplace(0, l1_sensitivity / dp_params.eps)
+
+
+def _apply_gaussian_mechanism(value, dp_params):
```
dvadym, line 34: Sensitivity also depends on type of aggregation, please replace dp_params -> l2_sensitivity.

dvadym (commented):
Thanks! I left some comments

pipeline_dp/mean_var_params.py
AlexandraImbrisca: Thanks for the explanation! Does the aggregation type only affect linf_sensitivity? If so, l1_sensitivity and l2_sensitivity have the same dependency on l0_sensitivity and linf_sensitivity. I created functions for computing l0_sensitivity/linf_sensitivity based on the type of aggregation and two general functions that compute l1_sensitivity/linf_sensitivity given l0_sensitivity & linf_sensitivity.

pipeline_dp/mean_var_params.py
```diff
@@ -0,0 +1,132 @@
...
+def _apply_laplace_mechanism(value: float, eps: float, l1_sensitivity: float):
+    # TODO: use the secure noise instead of np.random
+    return value + np.random.laplace(0, l1_sensitivity / eps)
+
+
+def _apply_gaussian_mechanism(value: float, eps: float, delta: float, l2_sensitivity: float):
+    sigma = _compute_sigma(eps, delta, l2_sensitivity)
+
+    # TODO: use the secure noise instead of np.random
+    return value + np.random.normal(0, np.power(sigma, 2) * (eps, delta, l2_sensitivity))
```
dvadym, line 52: 1.It looks like we don't need (eps, delta, l2_sensitivity) in this line
2.np.power(sigma, 2) -> sigma (according to [doc](https://numpy.org/doc/stable/reference/random/generated/numpy.random.normal.html) np.random.normal parameter is std, i.e. sigma

pipeline_dp/mean_var_params.py
```diff
@@ -0,0 +1,132 @@
...
+    low: float
+    high: float
+    max_partitions_contributed: int
+    max_contributions_per_partition: int
+    noise_kind: pipeline_dp.NoiseKind  # Laplace or Gaussian
+
+    def l0_sensitivity(self):
+        return self.max_partitions_contributed
+
+    def linf_sensitivity(self, metric):
```
dvadym, line 23: I like this approach with having one function for linf sensitivity!

pipeline_dp/mean_var_params.py
```diff
@@ -0,0 +1,132 @@
...
+    """Manages the parameters used for computing the mean/variance."""
+    eps: float
+    delta: float
+    low: float
+    high: float
+    max_partitions_contributed: int
+    max_contributions_per_partition: int
+    noise_kind: pipeline_dp.NoiseKind  # Laplace or Gaussian
+
+    def l0_sensitivity(self):
```
dvadym, line 20: I like this approach with having l0 sensitivty function (to emphasize how l0 sensitivity is computed from the parameters that an API user specified

pipeline_dp/mean_var_params.py
```diff
@@ -0,0 +1,132 @@
...
+        return self.max_partitions_contributed
+
+    def linf_sensitivity(self, metric):
+        return self.max_contributions_per_partition * {
```
dvadym, line 24: This is a pretty cool trick, but I'd suggest to have something line

if metric == COUNT:
   return ...
if metric == SUM:
   return
raise ValueError(...)

Clarification: it's slightly longer, but it's more readable and it's easy to see what's l_inf of each of metrics

pipeline_dp/mean_var_params.py
```diff
@@ -0,0 +1,132 @@
...
+            pipeline_dp.Metrics.COUNT: 1,
+            pipeline_dp.Metrics.SUM: np.max(self.low, self.high)
```
dvadym, line 26: self.low -> abs(self.low), self.high -> abs(self.high)

low and high might be negative, the sensitivity is absolute value of the maximum change of metrics, that's why we need to take abs()

pipeline_dp/mean_var_params.py
```diff
@@ -0,0 +1,132 @@
...
+
+import numpy as np
+import pipeline_dp
+
+from dataclasses import dataclass
+
+
+@dataclass
+class MeanVarParams:
+    """Manages the parameters used for computing the mean/variance."""
```
dvadym, line 11: Nit: 
1.I'd suggest to drop "Manages" (w/o this work no change in meaning)
2.Please add computing the dp sum, count, mean, variance

pipeline_dp/mean_var_params.py
```diff
@@ -0,0 +1,132 @@
+"""DP computing of count, sum, mean, variance."""
```
dvadym, line 1: Nit: DP -> Differential privacy  (it's nice to have in general comments w/o abbrevations)

pipeline_dp/mean_var_params.py
```diff
@@ -0,0 +1,132 @@
...
+
```
dvadym, line 2: Please rename a file to dp_computations.py
You can do  `git mv <old name> <new name>` it should save all comments

dvadym (changes requested):
Thanks! Sum and Count looks great. 

Please add unit tests for all functions, incl
l1/l2/linf_sensitifity, compute_sigma, apply_laplace/gaussian_mechanism

**Except**: compute_dp_mean, compute_dp_var (we will work on them later)

**Note:** For functions which generate random input: generate many samples (like 10K) and compute mean and std and to check that it's close to expected
You can use np.mean, np.std

I'll suggest some more advance tests later.

AlexandraImbrisca:
Thanks for all your suggestions!

I created most of the tests, but I have a question about the functions. Almost all the functions are not part of any class, and testing requires separate imports. For simplicity, I am currently importing all of them.

I am not very familiar with unit testing in Python, is there a better way to include the tests?

tests/dp_computations_test.py
```diff
@@ -0,0 +1,77 @@
...
+        # SUM aggregation
+        # Positive low, positive high
+        self.assertEqual(params.linf_sensitivity(metric=pipeline_dp.Metrics.SUM), 15)
+
+        # Negative low, positive high
+        params.low = -2
+        self.assertEqual(params.linf_sensitivity(metric=pipeline_dp.Metrics.SUM), 15)
+
+        # Negative low, negative high
+        params.high = -1
```
dvadym, line 40: Nice, checking different cases with positive and negative low/high!

dvadym (commented):
Thanks it looks good! I left some comments

AlexandraImbrisca:
Thanks for the suggestions!

The tests take more time now (~ 30 seconds), and I created some helper functions to test Laplace/Gaussian noise. Some tests fail (those that use the Gaussian mechanism) if delta is smaller than ~0.06. Could we use different values based on the noise type?

tests/dp_computations_test.py
```diff
@@ -0,0 +1,114 @@
...
+        self.assertEqual(pipeline_dp.compute_sigma(eps=0.5, delta=1e-10, l2_sensitivity=10),
+                         np.sqrt(2 * np.log(1.25 / 1e-10)) * 20)
+
+    def _test_laplace_noise(self, results, value, eps, l1_sensitivity):
+        self.assertAlmostEqual(np.mean(results), value, delta=0.075)
+        self.assertAlmostEqual(np.std(results), np.sqrt(2) * l1_sensitivity / eps, delta=0.075)
+        self.assertAlmostEqual(skew(results), 0, delta=0.075)
+        self.assertAlmostEqual(kurtosis(results), 3, delta=0.075)
+
+    def _test_gaussian_noise(self, results, value, eps, delta, l2_sensitivity):
```
dvadym, line 56: Yeah, that's a nice idea to have helpers methods for testing laplace and gaussian!

dvadym:
Yeah, it's understandable that those tests would fail sometimes since they are stochastic.

Could you please put delta=0.1 in `assertAlmostEqual` everywhere and run all tests 100 times and check how many times they fail?

[Here](https://gist.github.com/aychedee/5244147) is how to run tests many times with unittest (or you can use other way). It's only for verification how often tests fail and calibration of delta, no need to add that to PR.

pipeline_dp/__init__.py
```diff
@@ -1,6 +1,8 @@
 from pipeline_dp.budget_accounting import BudgetAccountant
 from pipeline_dp.aggregate_params import AggregateParams
 from pipeline_dp.aggregate_params import Metrics
+from pipeline_dp.aggregate_params import NoiseKind
+from pipeline_dp.dp_computations import *
```
dvadym, line 5: Nit: remove `from pipeline_dp.dp_computations import *`

You can use
`from pipeline_dp.dp_computations import *` 
in tests

**Explanation**: __init__.py contains everything that is exposed in the PipelineDP Public API. `dp_computations` are not public API, they are part of implementation. And it's better to decrease number of objects/functions exposed in Public API  - it's less confusing for users and the Public API should be stable (in order not to break user).

AlexandraImbrisca:
Great ideas! :) I ran all the 9 tests 100 times, and there are 15 failures out of 900 runs.

tests/dp_computations_test.py
```diff
@@ -0,0 +1,132 @@
...
+    def test_compute_sigma(self):
+        self.assertEqual(
+            pipeline_dp.dp_computations.compute_sigma(eps=1, delta=1, l2_sensitivity=10),
+            np.sqrt(2 * np.log(1.25)) * 10)
+        self.assertEqual(
+            pipeline_dp.dp_computations.compute_sigma(eps=0.5, delta=1e-10, l2_sensitivity=10),
+            np.sqrt(2 * np.log(1.25 / 1e-10)) * 20)
+
+    def _test_laplace_noise(self, results, value, eps, l1_sensitivity):
+        self.assertAlmostEqual(np.mean(results), value, delta=0.075)
```
dvadym, line 57: please change 0.075 -> 0.1 everywhere. It's better to keep things simpler. Moreover it should decrease chances of failng to less <1%

dvadym (commented):
Thanks, it looks like a reasonable failing rate. Just one comment, please change 0.075 -> 0.1

AlexandraImbrisca:
My bad! I updated delta = 0.1 in all assertAlmostEqual calls.

dvadym (approved):
Thanks a lot of implementing! 

Looking forward to mean and variance :) !

Issue #37, opened 2021-06-02
Implement Accumulator Framework for aggregation


dvadym:
# Context
Computing of DP metrics will include pipeline aggregations. For this we need to have some ways to compute those aggregations. This issue is about implementing of basic structure of object that are responsible for aggregations. Let's call such object accumulators.

**Note:** The ideas of accumulators and their API are influence of Apache Beam accumulators (see example 5 on [page](https://beam.apache.org/documentation/transforms/python/aggregation/combineperkey/)).

**Note:** There is an [issue](https://github.com/OpenMined/PipelineDP/issues/9) on implementing **SumAccumulator**. This issue is on introducing the accumulator framework. Which is important since we will have multiple types of accumulators and there might be multiple accumulators per 1 aggregation.

# Goals
To implement

1. Abstract base class

```
class Accumulator(ABC):
     @abstractmethod
     def add_value(v)
     @abstractmethod
     def add_accumulator(self, accumulator:Accumulator) -> Accumulator # returns itself
     @abstractmethod
    def compute_metrics()
```

2. CompoundAccumulator

```
  class CompoundAccumulator(Accumulator):
      def __init__(self, accumulators)
      def merge(cls,  accumulators: Iterable[CompoundAccumulator]) -> CompoundAccumulator 
      def add_value(v)
      def add_accumulator(self, accumulator:CompoundAccumulator) -> CompoundAccumulator # returns itself
      def compute_metrics()
```

  which keeps multiple accumulators (of different types) and delegates **merge()** and **compute_metrics()** to those accumulators.

3. Serialization of accumulator classes.
4. `AccumulatorFactory` is the factory for creating accumulator of specific types with specific parameters (details in next subsection)

## AccumulatorFactory
**API**:
```
class AccumulatorFactory:
   def __init__(params: [AggregateParams])
   def initialize()
   def create(values:List) -> Accumulator
```

[AggregateParams](https://github.com/OpenMined/PipelineDP/blob/5f8e2b5e1265af6e29f3c2fb79d0c9846a63ba43/pipeline_dp/aggregate_params.py#L22) in sources
**Notes** 
1. `AccumulatorFactory` is responsible for creating a right set of accumulators based on `params`.
2. `AccumulatorFactory` will be created during the pipeline construction. `create()` method will be called during the pipeline execution.
3. There will be some additional parts of API (described later).
[...]

PR #46, opened 2021-06-13
Implement reduce_accumulators_per_key in SparkRDDOperations (# 41)

### LINKED ISSUE #41: Using accumulators in SparkOperations
# Context

Currently the development of [Accumulator Framework](https://github.com/OpenMined/PipelineDP/issues/37) is in progress. The idea is that **Accumulator Framework** is independent of the pipeline framework. This issue is about implementing usage of **Accumulator Framework** with Spark.

The interface of accumulators:

```
class Accumulator(ABC):
     def merge(cls,  accumulators: Iterable[Accumulator]) -> Accumulator
     def add(v)
     def compute_metrics()
```

# Goals
To implement the following method of [SparkRDDOperations](https://github.com/OpenMined/PipelineDP/blob/65d00ee6c906a36b8a173ddbb83d7e3bf71905b4/pipeline_dp/pipeline_operations.py#L102):

`def reduce_accumulators_per_key(self, col):`

where **col** contains elements:  (key, accumulator) 

and the return is a collection of elements (key, accumulator), such that all elements per each key are merged.

# Notes
It looks like **reduce_accumulators_per_key** might be implement with usage of [reduce_by_key](https://spark.apache.org/docs/latest/api/python/reference/api/pyspark.RDD.reduceByKey.html).

boyaryn:
## Description
Implementation of accumulator reduction for Spark processing engine.
Resolves #41.

## Affected Dependencies
N/A

## How has this been tested?
- Tests are to be added as they need the [Accumulator Framework](https://github.com/OpenMined/PipelineDP/issues/37) which is to be defined.

## Checklist

### DIFF
--- a/pipeline_dp/pipeline_operations.py
+++ b/pipeline_dp/pipeline_operations.py
@@ -54,6 +54,10 @@ class PipelineOperations(abc.ABC):
     def count_per_element(self, col, stage_name: str):
         pass
 
+    @abc.abstractmethod
+    def reduce_accumulators_per_key(self, col, stage_name: str):
+        pass
+
 
 class BeamOperations(PipelineOperations):
     """Apache Beam adapter."""
@@ -156,6 +160,19 @@ class SparkRDDOperations(PipelineOperations):
     def count_per_element(self, rdd, stage_name: str = None):
         return rdd.map(lambda x: (x, 1))\
             .reduceByKey(lambda x, y: (x + y))
+
+    def reduce_accumulators_per_key(self, rdd, stage_name: str = None):
+        """Reduce the input RDD so that all elements per each key are merged.
+
+        Args:
+          rdd: input rdd which contains tuples (key, accumulator)
+          stage_name: name of the stage
+
+        Returns:
+          An RDD of tuples (key, accumulator).
+
+        """
+        rdd.reduceByKey(lambda acc1, acc2: acc1.merge([acc2]))
 
 
 class LocalPipelineOperations(PipelineOperations):
--- a/pipeline_dp/pipeline_operations.py
+++ b/pipeline_dp/pipeline_operations.py
@@ -254,7 +254,7 @@ class SparkRDDOperations(PipelineOperations):
             .reduceByKey(lambda x, y: (x + y))
 
     def reduce_accumulators_per_key(self, rdd, stage_name: str = None):
-        return rdd.reduceByKey(lambda acc1, acc2: acc1.merge([acc2]))
+        return rdd.reduceByKey(lambda acc1, acc2: acc1.add_accumulator(acc2))
 
 
 class LocalPipelineOperations(PipelineOperations):
--- a/pipeline_dp/pipeline_operations.py
+++ b/pipeline_dp/pipeline_operations.py
@@ -55,6 +55,16 @@ class PipelineOperations(abc.ABC):
         pass
 
     @abc.abstractmethod
+    """Reduce the input collection so that all elements per each key are merged.
+
+    Args:
+      col: input collection which contains tuples (key, accumulator)
+      stage_name: name of the stage
+
+    Returns:
+      A collection of tuples (key, accumulator).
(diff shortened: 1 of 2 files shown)


pipeline_dp/pipeline_operations.py
```diff
@@ -157,6 +161,19 @@ def count_per_element(self, rdd, stage_name: str = None):
         return rdd.map(lambda x: (x, 1))\
             .reduceByKey(lambda x, y: (x + y))
 
+    def reduce_accumulators_per_key(self, rdd, stage_name: str = None):
+        """Reduce the input RDD so that all elements per each key are merged.
```
dvadym, line 165: Please move this comment to a base class (just replacing, RDD -> collections). The idea that the code that uses PipelineOperations don't need to know which framework they use

pipeline_dp/pipeline_operations.py
```diff
@@ -157,6 +161,19 @@ def count_per_element(self, rdd, stage_name: str = None):
...
+
+        Args:
+          rdd: input rdd which contains tuples (key, accumulator)
+          stage_name: name of the stage
+
+        Returns:
+          An RDD of tuples (key, accumulator).
+
+        """
+        rdd.reduceByKey(lambda acc1, acc2: acc1.merge([acc2]))
```
dvadym, line 175: let's assume for now that merge takes a list (then let's discuss on accumulator framework how to allow acceptance of 1 accumulator)

dvadym (changes requested):
Thanks, it looks good! 

Could you please add unit tests? You can use some simple mock object. Like Accumulator that does counting

pipeline_dp/pipeline_operations.py
boyaryn: Good idea. Updated.

pipeline_dp/pipeline_operations.py
boyaryn: okay

boyaryn:
> Thanks, it looks good!
> 
> Could you please add unit tests? You can use some simple mock object. Like Accumulator that does counting

I'll add unit tests. Is it better to wait when the Accumulator interface is available in `main` so that the tests can check real-world cases?

chinmayshah99:
> I'll add unit tests. Is it better to wait when the Accumulator interface is available in `main` so that the tests can check real-world cases?
Let's add some mock tests for now as @dvadym suggested. Once the `Accumulator` interface is available, we can move to more real-world test cases.

dvadym:
I agree, please create very a simple MockAccumulator (eg. CounterAccumulator). Anyway in unittests it's better to use mock objects that real accumulators (which will be pretty complicated objects, since they will have support of Differential privacy)

dvadym:
FYI: [SumAccumulator] Mock accumulator (https://github.com/OpenMined/PipelineDP/blob/e33a7c724179dba77b7646597f2af80e47ffe5a7/tests/pipeline_operations_test.py#L316) was implemented for testing of BeamOperations. 

Please use SumAccumulator for testing

pipeline_dp/pipeline_operations.py
```diff
@@ -254,8 +253,8 @@ def count_per_element(self, rdd, stage_name: str = None):
         return rdd.map(lambda x: (x, 1))\
             .reduceByKey(lambda x, y: (x + y))
 
-    def reduce_accumulators_per_key(self, col, stage_name: str = None):
-        raise NotImplementedError()
+    def reduce_accumulators_per_key(self, rdd, stage_name: str = None):
+        return rdd.reduceByKey(lambda acc1, acc2: acc1.merge([acc2]))
```
dvadym, line 257: acc1.merge -> acc1.add_accumulator(acc2)

dvadym (changes requested):
Thanks, it looks good. Just a minor fix.

dvadym (approved):
Thanks for contributing!

PR #51, opened 2021-07-01
[WIP] DP computing of mean and variance

AlexandraImbrisca:
## Description
Initial implementation of [compute_dp_mean](https://github.com/OpenMined/PipelineDP/issues/36).

## Affected Dependencies
None.

## How has this been tested?
Unit tests are included in tests\dp_computations_test.py.

## Checklist

### DIFF
--- a/pipeline_dp/dp_computations.py
+++ b/pipeline_dp/dp_computations.py
@@ -26,8 +26,14 @@ class MeanVarParams:
         if metric == pipeline_dp.Metrics.SUM:
             return self.max_contributions_per_partition * max(
                 abs(self.low), abs(self.high))
+        if metric == pipeline_dp.Metrics.MEAN:
+            return self.max_contributions_per_partition * abs(
+                self.high - self.low) / 2
         # TODO: add values for mean and variance
         raise ValueError("Invalid metric")
+
+    def middle(self):
+        return self.low + (self.high - self.low) / 2
 
 
 def compute_l1_sensitivity(l0_sensitivity: float, linf_sensitivity: float):
@@ -70,6 +76,32 @@ def _add_random_noise(value: float, eps: float, delta: float,
     raise ValueError("Noise kind must be either Laplace or Gaussian.")
 
 
+def equally_split_budget(eps: float, delta: float, no_mechanisms: int):
+    """Equally splits the budget (eps, delta) between a given number of mechanisms.
+
+    Args:
+        eps, delta: Parameters of (epsilon, delta)-differential privacy.
+        no_mechanisms: The number of mechanisms between which we split the budget.
+
+    Raises:
+        ValueError: The number of mechanisms must be a natural non-zero number.
+
+    Returns:
+        An array with the split budgets.
+    """
+    if no_mechanisms <= 0:
+        raise ValueError(
+            "The number of mechanisms must be a natural non-zero number.")
+
+    budgets = [(eps / no_mechanisms, delta / no_mechanisms) for _ in
+               range(no_mechanisms - 1)]
+    last_mechanism_budget = (eps - (no_mechanisms - 1) * (eps / no_mechanisms),
+                             delta - (no_mechanisms - 1) * (
+                                     delta / no_mechanisms))
+    budgets.append(last_mechanism_budget)
+    return budgets
+
+
 def compute_dp_count(count: int, dp_params: MeanVarParams):
     """Computes DP count.
 
@@ -104,3 +136,36 @@ def compute_dp_sum(sum: float, dp_params: MeanVarParams):
     return _add_random_noise(sum, dp_params.eps, dp_params.delta,
                              l0_sensitivity, linf_sensitivity,
                              dp_params.noise_kind)
+
+
+def compute_dp_mean(count: int, sum: float, dp_params: MeanVarParams):
+    """Computes DP mean.
+
+    Args:
+        count: Non-DP count.
+        sum: Non-DP sum.
+        dp_params: The parameters used at computing the noise.
+
+    Raises:
+        ValueError: The noise kind is invalid.
+    """
+    middle = dp_params.middle()
+    normalized_sum = sum - count * middle
+
+    l0_sensitivity = dp_params.l0_sensitivity()
+
+    # Splits the budget equally between the two mechanisms.
+    (sum_eps, sum_delta), (count_eps, count_delta) = equally_split_budget(
+        dp_params.eps, dp_params.delta, 2)
+
+    dp_normalized_sum = _add_random_noise(normalized_sum, sum_eps, sum_delta,
+                                          l0_sensitivity,
+                                          dp_params.linf_sensitivity(
+                                              pipeline_dp.Metrics.MEAN),
+                                          dp_params.noise_kind)
+    dp_count = _add_random_noise(count, count_eps, count_delta, l0_sensitivity,
+                                 dp_params.linf_sensitivity(
+                                     pipeline_dp.Metrics.COUNT),
+                                 dp_params.noise_kind)
+    dp_mean = dp_normalized_sum / dp_count + middle
+    return dp_count, dp_normalized_sum + count * middle, dp_mean
--- a/pipeline_dp/dp_computations.py
+++ b/pipeline_dp/dp_computations.py
@@ -22,7 +22,7 @@ class MeanVarParams:
         return self.max_partitions_contributed
 
     def linf_sensitivity(self, metric):
-        """Returns the L1 sensitivity of the parameters based on the metric.
+        """Returns the Linf sensitivity of the parameters based on the metric.
 
         Args:
             metric: The metric performed.
@@ -37,7 +37,7 @@ class MeanVarParams:
                 abs(self.low), abs(self.high))
         if metric == pipeline_dp.Metrics.MEAN:
             return self.max_contributions_per_partition * abs(
-                self.high - self.low) / 2
+                self.middle() + self.low)
         # TODO: add value for variance
         raise ValueError("Invalid metric")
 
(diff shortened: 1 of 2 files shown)


pipeline_dp/dp_computations.py
```diff
@@ -104,3 +205,36 @@ def compute_dp_sum(sum: float, dp_params: MeanVarParams):
...
+
+    Raises:
+        ValueError: The noise kind is invalid.
+    """
+    middle = dp_params.middle()
+    normalized_sum = sum - count * middle
+
+    l0_sensitivity = dp_params.l0_sensitivity()
+
+    # Splits the budget equally between the two mechanisms.
```
AlexandraImbrisca, line 226: @dvadym Is there a better way to split the budget between the mechanisms?

pipeline_dp/dp_computations.py
```diff
@@ -18,39 +18,101 @@ class MeanVarParams:
     noise_kind: pipeline_dp.NoiseKind  # Laplace or Gaussian
 
     def l0_sensitivity(self):
+        """"Returns the L0 sensitivity of the parameters."""
         return self.max_partitions_contributed
 
     def linf_sensitivity(self, metric):
+        """Returns the L1 sensitivity of the parameters based on the metric.
```
dvadym, line 25: L1 -> Linf

pipeline_dp/dp_computations.py
```diff
@@ -104,3 +205,36 @@ def compute_dp_sum(sum: float, dp_params: MeanVarParams):
...
+    """Computes DP mean.
+
+    Args:
+        count: Non-DP count.
+        sum: Non-DP sum.
+        dp_params: The parameters used at computing the noise.
+
+    Raises:
+        ValueError: The noise kind is invalid.
+    """
```
dvadym, line 220: Returns: the tuple of anonymized count, sum and mean.

pipeline_dp/dp_computations.py
dvadym: For (eps, delta)-dp it's a good way. Mean computations for C++ DP library also uses such division. Later, there will be usage of PLD (Privacy Loss Distributions), it will be possible to split (eps, delta) in "more optimal" way.

tests/dp_computations_test.py
```diff
@@ -172,6 +172,73 @@ def test_compute_dp_sum(self):
...
+            params.eps, params.delta, 2)
+        l0_sensitivity = params.l0_sensitivity()
+
+        # Laplace Mechanism
+        results = [
+            pipeline_dp.dp_computations.compute_dp_mean(count=10, sum=100,
+                                                        dp_params=params)
+            for _ in range(1000000)
+        ]
+        count_values, sum_values, mean_values = zip(*results)
```
dvadym, line 215: Could you please check that np.mean(mean_values) is close to the non-dp mean and that np.std(mean_values) is small enough. The same for Gaussian

pipeline_dp/dp_computations.py
dvadym: Thanks for adding comments!

dvadym (commented):
Thanks, it looks good!

I've asked dasmdasm to make a review of the algorithm implementation, he has implemented DP mean and var in C++ library, so he has much experience in this.

pipeline_dp/dp_computations.py
```diff
@@ -18,39 +18,101 @@ class MeanVarParams:
...
+        """
         if metric == pipeline_dp.Metrics.COUNT:
             return self.max_contributions_per_partition
         if metric == pipeline_dp.Metrics.SUM:
             return self.max_contributions_per_partition * max(
                 abs(self.low), abs(self.high))
-        # TODO: add values for mean and variance
+        if metric == pipeline_dp.Metrics.MEAN:
+            return self.max_contributions_per_partition * abs(
+                self.high - self.low) / 2
```
dasmdasm, line 40: nit: You have the middle() function, why not use it here?

pipeline_dp/dp_computations.py
```diff
@@ -70,6 +145,32 @@ def _add_random_noise(value: float, eps: float, delta: float,
...
+    Returns:
+        An array with the split budgets.
+    """
+    if no_mechanisms <= 0:
+        raise ValueError(
+            "The number of mechanisms must be a natural non-zero number.")
+
+    budgets = [(eps / no_mechanisms, delta / no_mechanisms) for _ in
+               range(no_mechanisms - 1)]
+    last_mechanism_budget = (eps - (no_mechanisms - 1) * (eps / no_mechanisms),
```
dasmdasm, line 167: It's safer to actually sum the budgets and subtract them from the total epsilon and delta, rather than calculating what the sum should be.

pipeline_dp/dp_computations.py
```diff
@@ -104,3 +205,36 @@ def compute_dp_sum(sum: float, dp_params: MeanVarParams):
...
+                                          l0_sensitivity,
+                                          dp_params.linf_sensitivity(
+                                              pipeline_dp.Metrics.MEAN),
+                                          dp_params.noise_kind)
+    dp_count = _add_random_noise(count, count_eps, count_delta, l0_sensitivity,
+                                 dp_params.linf_sensitivity(
+                                     pipeline_dp.Metrics.COUNT),
+                                 dp_params.noise_kind)
+    dp_mean = dp_normalized_sum / dp_count + middle
+    return dp_count, dp_normalized_sum + count * middle, dp_mean
```
dasmdasm, line 240: dp_normalized_sum + count * middle is not differentially private (because count is not differentially private), so we can't return it. As an example, a simple attack would be to calculate dp_normalized_sum = (dp_mean - middle) * dp_count, then we can recover count * middle, and then count directly (since middle is public knowledge).

If you want to release a dp_sum that is different than the one you can get via post-processing (dp_mean * dp_count), then you'll need to spend budget to noise it separately.

dasmdasm (changes requested):

pipeline_dp/dp_computations.py
AlexandraImbrisca: Sounds good, thanks!

tests/dp_computations_test.py
AlexandraImbrisca: Sure! I am not quite sure what value should delta have here. After testing it multiple times, I noticed that the mean might differ by even ~7.0

pipeline_dp/dp_computations.py
AlexandraImbrisca: Thanks for the explanation! It is more clear now. I will modify the function to return dp_mean * dp_count.

tests/dp_computations_test.py
dvadym: Yeah, this makes sense. 
denominator = 10 + lap(4), when denominator is small, mean is large -> which makes disproportional contribution in np.mean() computation. This is just an example, that DP doesn't work nice with small number of elements (10 here).

Please update sum: 100 -> 10000, count: 10 -> 1000. (count = 1000 are reasonable number when DP becomes applicable)
It should provide reasonable close mean.

AlexandraImbrisca: Thanks for the explanation and sorry for the late response! I have a small follow up question about the standard deviation of sum/mean. As I understand it, the mean is a ratio between two Laplace/Gaussian distributions each one with its own std. I tried to compute the std of the ratio based on the std of numerator and denominator, but I couldn't find a relation. Is there something that I am missing out?

dvadym: Correct, mean is the the ratio of 2 Laplace/Gaussian distributions. Unfortunately it's pretty hard to work with ratio distributions ([wiki](https://en.wikipedia.org/wiki/Ratio_distribution)). 
But often there is no need to work with ratio distributions, for eg. in case of mean, the idea that if the noise is small in comparison to the count, then 
(sum+noise_1)/(count+noise_2) is not far from sum/count.

When the noise is not small in comparison to the count, the noise will distort the results significantly and in such cases DP is not applicable. But that's a fundamental problem with anonymization - it's hard (if possible) to anonymize small datasets. That's why I suggested to change the test data for more suitable for DP data.

AlexandraImbrisca: I changed the test data, and the mean is close enough now. Thanks!

pipeline_dp/dp_computations.py
```diff
@@ -70,6 +145,40 @@ def _add_random_noise(value: float, eps: float, delta: float,
...
+
+    Returns:
+        An array with the split budgets.
+    """
+    if no_mechanisms <= 0:
+        raise ValueError(
+            "The number of mechanisms must be a positive integer.")
+
+    # These variables are used to keep track of the budget used instead
+    # of estimating it. In this way, we can make use of the whole budget.
```
dvadym, line 166: Nit: remove "instead of estimating it." (it's not clear what estimating mean).
Please change it: In this way, we can improve accuracy of floating-point operations.

pipeline_dp/dp_computations.py
```diff
@@ -70,6 +145,40 @@ def _add_random_noise(value: float, eps: float, delta: float,
...
+    eps_used = 0
```
dvadym, line 167: Nit: eps_used = delta_used = 0 or eps_used, delta_used = 0, 0

tests/dp_computations_test.py
```diff
@@ -172,6 +172,68 @@ def test_compute_dp_sum(self):
...
+                               count_delta) = pipeline_dp.dp_computations.equally_split_budget(
+            params.eps, params.delta, 2)
+        l0_sensitivity = params.l0_sensitivity()
+
+        # Laplace Mechanism
+        results = [
+            pipeline_dp.dp_computations.compute_dp_mean(count=1000, sum=10000,
+                                                        dp_params=params)
+            for _ in range(1000000)
+        ]
```
dvadym, line 214: Cool usage of zip!

tests/dp_computations_test.py
```diff
@@ -172,6 +172,68 @@ def test_compute_dp_sum(self):
...
+        # Multiple mechanisms.
+        expected_budgets = [(0.5 / 5, 1e-10 / 5) for _ in range(4)]
+        expected_budgets.append((0.5 - 4 * (0.5 / 5), 1e-10 - 4 * (1e-10 / 5)))
+
+        self.assertEqual(
+            pipeline_dp.dp_computations.equally_split_budget(0.5, 1e-10, 5),
+            expected_budgets
+        )
+
+    def test_compute_dp_mean(self):
```
dvadym, line 194: What's the running time of newly added tests?

tests/dp_computations_test.py
```diff
@@ -172,6 +172,68 @@ def test_compute_dp_sum(self):
...
+            eps=0.5,
+            delta=1e-10,
+            low=1,
+            high=20,
+            max_partitions_contributed=1,
+            max_contributions_per_partition=1,
+            noise_kind=pipeline_dp.NoiseKind.LAPLACE)
+
+        (sum_eps, sum_delta), (count_eps,
+                               count_delta) = pipeline_dp.dp_computations.equally_split_budget(
```
dvadym, line 205: this line looks pretty long. Can you try run again `make format`?

dvadym (commented):
Thanks it looks great! A few minor comments

tests/dp_computations_test.py
AlexandraImbrisca: Sure :)

tests/dp_computations_test.py
AlexandraImbrisca: Both tests take around 35 seconds.

dvadym: Thanks, it's reasonable running time

dvadym (approved):
Thanks for implementing mean!

PR #54, opened 2021-07-08
Implementation of `_select_private_partitions`

levzlotnik:
## Description
Implementation of pipelined `_select_private_partitions` as specified in #8 using `pydp.algorithms.partition_selection.PartitionSelectionStrategy`. 

## Affected Dependencies
**New Dependency:** OpenMined/PyDP aka `python-dp`.

## How has this been tested?
- Mocking `create_*_parititon_strategy` to deterministically generate private partitions.
- Passing toy data through the `_select_private_partitions` function
- Measuring expected result against outcome

## Checklist

### DIFF
--- a/pipeline_dp/dp_engine.py
+++ b/pipeline_dp/dp_engine.py
@@ -1,12 +1,14 @@
 """DP aggregations."""
 
-from typing import Callable
+from typing import Callable, Optional
 
 from dataclasses import dataclass
 from pipeline_dp.aggregate_params import AggregateParams
 from pipeline_dp.budget_accounting import BudgetAccountant
 from pipeline_dp.pipeline_operations import PipelineOperations
 from pipeline_dp.report_generator import ReportGenerator
+
+from pydp.algorithms.partition_selection import PartitionSelectionStrategy
 
 
 @dataclass
@@ -25,10 +27,12 @@ class DPEngine:
     """Performs DP aggregations."""
 
     def __init__(self, budget_accountant: BudgetAccountant,
-                 ops: PipelineOperations):
+                 ops: PipelineOperations, 
+                 partition_selection_strategy: Optional[PartitionSelectionStrategy] = None):
         self._budget_accountant = budget_accountant
         self._ops = ops
         self._report_generators = []
+        self._partition_selection_strategy = partition_selection_strategy
 
     def _add_report_stage(self, text):
         self._report_generators[-1].add_stage(text)
@@ -110,3 +114,20 @@ class DPEngine:
         return self._ops.flat_map(col,
                                   unnest_cross_partition_bound_sampled_per_key,
                                   "Unnest")
+    
+    def _select_private_partitions(self, col):
+        """
+        Selects and publishes private partitions.
+
+        Args:
+            col: collection, with types for each element: 
+                (partition_key, privacy_id_count, value)
+        
+        Returns:
+            collection of elements (partition_key, data)
+        """
+        def filter_fn(row):
+            return self._partition_selection_strategy.should_keep(row[1])
+
+        filtered_values = self._ops.filter(col, filter_fn)
+        return self._ops.group_by_key(filtered_values, lambda tup: (tup[0], tup[2]))
--- a/pipeline_dp/dp_engine.py
+++ b/pipeline_dp/dp_engine.py
@@ -8,9 +8,9 @@ from pipeline_dp.aggregate_params import AggregateParams
 from pipeline_dp.budget_accounting import BudgetAccountant, Budget
 from pipeline_dp.pipeline_operations import PipelineOperations
 from pipeline_dp.report_generator import ReportGenerator
+from pipeline_dp.accumulator import Accumulator
 
 from pydp.algorithms.partition_selection import create_truncated_geometric_partition_strategy
-from .accumulator import Accumulator
 
 
 @dataclass
@@ -120,9 +120,9 @@ class DPEngine:
 
         Args:
             col: collection, with types for each element: 
-                (partition_key, accumulator: Accumulator)
-            max_partitions_contributed: how many contributed partitions we can get at 
-                most from this operations.
+                (partition_key, Accumulator)
+            max_partitions_contributed: maximum amount of partitions that one privacy unit
+                might contribute.
         
         Returns:
             collection of elements (partition_key, accumulator)
@@ -130,7 +130,7 @@ class DPEngine:
         budget = self._budget_accountant.request_budget(weight=1, use_eps=True, use_delta=True)
         
         def filter_fn(captures: Tuple[Budget, int], row: Tuple[Any, Accumulator]) -> bool:
-            """Lazily create a partition selection strategy and use it to determine which 
+            """Lazily creates a partition selection strategy and uses it to determine which 
             partitions to keep."""
             budget, max_partitions = captures 
             accumulator = row[1] 
--- a/pipeline_dp/dp_engine.py
+++ b/pipeline_dp/dp_engine.py
@@ -1,14 +1,16 @@
 """DP aggregations."""
 
-from typing import Callable, Optional
+from functools import partial
+from typing import Any, Callable, Tuple
 
 from dataclasses import dataclass
 from pipeline_dp.aggregate_params import AggregateParams
-from pipeline_dp.budget_accounting import BudgetAccountant
+from pipeline_dp.budget_accounting import BudgetAccountant, Budget
 from pipeline_dp.pipeline_operations import PipelineOperations
 from pipeline_dp.report_generator import ReportGenerator
 
-from pydp.algorithms.partition_selection import PartitionSelectionStrategy
+from pydp.algorithms.partition_selection import create_truncated_geometric_partition_strategy
+from .accumulator import Accumulator
 
 
 @dataclass
@@ -27,12 +29,10 @@ class DPEngine:
     """Performs DP aggregations."""
 
     def __init__(self, budget_accountant: BudgetAccountant,
-                 ops: PipelineOperations, 
-                 partition_selection_strategy: Optional[PartitionSelectionStrategy] = None):
+                 ops: PipelineOperations):
         self._budget_accountant = budget_accountant
         self._ops = ops
         self._report_generators = []
-        self._partition_selection_strategy = partition_selection_strategy
 
     def _add_report_stage(self, text):
         self._report_generators[-1].add_stage(text)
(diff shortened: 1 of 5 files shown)


levzlotnik:
@dvadym Hey Vadym, is this what you had in mind in terms of implementation?

pipeline_dp/dp_engine.py
```diff
@@ -110,3 +114,20 @@ def unnest_cross_partition_bound_sampled_per_key(pid_pk_v):
...
                                   unnest_cross_partition_bound_sampled_per_key,
                                   "Unnest")
+    
+    def _select_private_partitions(self, col):
+        """
+        Selects and publishes private partitions.
+
+        Args:
+            col: collection, with types for each element: 
+                (partition_key, privacy_id_count, value)
```
dvadym, line 124: sry, data format was slightly outdated, data format should be
`(partition_key, accumulator)` (src of [accumulator](https://github.com/OpenMined/PipelineDP/blob/5f8e2b5e1265af6e29f3c2fb79d0c9846a63ba43/pipeline_dp/accumulator.py#L12))

please assume that accumulator has `privacy_id_count` property (it will be added soon)

pipeline_dp/dp_engine.py
```diff
@@ -25,10 +27,12 @@ class DPEngine:
     """Performs DP aggregations."""
 
     def __init__(self, budget_accountant: BudgetAccountant,
-                 ops: PipelineOperations):
+                 ops: PipelineOperations, 
+                 partition_selection_strategy: Optional[PartitionSelectionStrategy] = None):
         self._budget_accountant = budget_accountant
         self._ops = ops
         self._report_generators = []
+        self._partition_selection_strategy = partition_selection_strategy
```
dvadym, line 35: Partitions selection strategy can be different for different aggregations, so let's create it inside _select_private_partions (more details in comments there)

pipeline_dp/dp_engine.py
```diff
@@ -110,3 +114,20 @@ def unnest_cross_partition_bound_sampled_per_key(pid_pk_v):
         return self._ops.flat_map(col,
                                   unnest_cross_partition_bound_sampled_per_key,
                                   "Unnest")
+    
+    def _select_private_partitions(self, col):
+        """
+        Selects and publishes private partitions.
```
dvadym, line 120: Nit (=style improvement)
"""Selects private partitions  ( no need in publishes)

pipeline_dp/dp_engine.py
```diff
@@ -110,3 +114,20 @@ def unnest_cross_partition_bound_sampled_per_key(pid_pk_v):
...
+
+        Args:
+            col: collection, with types for each element: 
+                (partition_key, privacy_id_count, value)
+        
+        Returns:
+            collection of elements (partition_key, data)
+        """
+        def filter_fn(row):
+            return self._partition_selection_strategy.should_keep(row[1])
```
dvadym, line 130: As mentioned earlier we need to create PartitionSelectionStrategy in this function, there are some nuances.

1.For now let's use only `TruncatedGeometric` selection strategy (budget accounting is not ready yet for other methods).
2.`TruncatedGeometric` requires (eps, delta, max_partitions_contributed)
  - please add `max_partitions_contributed` to arguments of  `_select_private_partitions`
  - (eps, delta) are computed lazily after the pipeline is constructed (and before it started executing on workers), which means that PartitionSelectionStrategy should be created lazily, namely
      1. By calling self._budget_accountant.[request_budget()](https://github.com/OpenMined/PipelineDP/blob/5f8e2b5e1265af6e29f3c2fb79d0c9846a63ba43/pipeline_dp/budget_accounting.py#L72), you can get `Budget` object, but budget.eps, budget.delta will be available only during the running (e.g. when `filtern_fn` is called)
      2. In `filter_fn` it's possible to create `PartitionSelectionStrategy` and use it (I'm not sure how expensive is creating PyBind11 object, we can start from creating it on each call of `filter_fn` and if it's expensive to figure out how to create it only once per a worker)

pipeline_dp/dp_engine.py
```diff
@@ -110,3 +114,20 @@ def unnest_cross_partition_bound_sampled_per_key(pid_pk_v):
...
+
+        filtered_values = self._ops.filter(col, filter_fn)
+        return self._ops.group_by_key(filtered_values, lambda tup: (tup[0], tup[2]))
```
dvadym, line 133: In this line it's more correct to use `_ops.map` instead of group_by_key, but since input data format update only `filter` is needed.

dvadym (commented):
Thanks, it's a good start! Let's iterate on this. 

There are some nuances, I've added some comments. No need to add yet tests, let's first iterate quick on making the design correct.

levzlotnik:
I'm not entirely clear on the use case.
When I `self._budged_accountant.request_budget()` per row, where each row is (by my guess) a partition pre-aggregation tuple - what weight should I use for the budget?  
My intuition is that the weight for `eps,delta` should be proportional to the 
In essence:

```python
def filter_fn(row):
   budget = self._budget_accountant.request_budget(weight=?, use_eps=True, use_delta=True)
   partition_selector = create_partition_strategy("truncated_geometric", budget.eps, budget.delta, max_partitions)
   return partition_selector.should_keep(row[1].privacy_id_count)
```

I'm a bit unsure about what kind of computation there is, maybe it would be more helpful to define a SQL pseudo-code side by side with the expected python query? Is it something along the lines of  

<table>
<tr>
<td> SQL </td> <td> PipelineDP </td>
</tr>
<tr>
<td> 

```sql
SELECT partition_key, accumulator
FROM (
    SELECT partition_key, 
        accumulator(value) as accumulator,
        tsgd(budget(weight)) as tsgd
    FROM collection
    GROUP BY partition_key
)
WHERE tsgd.should_keep(accumulator.privacy_id)
```

</td>
<td>

```python
groups = dp_engine._ops.group_by_key("partition_key")
groups = dp_engine._ops.map(groups, 
   lambda row: row[0], accumulator(row[1])
)
groups = dp_engine._select_private_partitions(groups, 
    max_partitions_contributed)
```

</td>
</tr>
</table>

dvadym:
Sorry, I was not precise enough. `request_budget() `should be once per `_select_private_partitions`.

The idea is that (eps, delta) budget is per operation (in this case per partition selection) not per row, not per partition, so changing lines in your example:

```
budget = self._budget_accountant.request_budget(weight=1, use_eps=True, use_delta=True) # let's set weight = 1 by default
def filter_fn(row):
    partition_selector = create_partition_strategy("truncated_geometric", 
          budget.eps, budget.delta, max_partitions)
    return partition_selector.should_keep(row[1].privacy_id_count)
```

Snippets of usage make sense. I like the idea of snippets. I'll try to add such snippets in future bug descriptions.

**More details on accumulators**: accumulators will be created in a chain of aggregations (for now 2 group_by_keys, but it may be changed in future), but conceptually it's the same as group_by_key/map in your example.

Does it answer your questions?

pipeline_dp/dp_engine.py
```diff
@@ -110,3 +114,31 @@ def unnest_cross_partition_bound_sampled_per_key(pid_pk_v):
...
                                   unnest_cross_partition_bound_sampled_per_key,
                                   "Unnest")
+    
+    def _select_private_partitions(self, col, max_partitions_contributed: int):
+        """Selects and publishes private partitions.
+
+        Args:
+            col: collection, with types for each element: 
+                (partition_key, accumulator: Accumulator)
+            max_partitions_contributed: how many contributed partitions we can get at
```
dvadym, line 124: maximum amount partition that one privacy unit might contribute.

pipeline_dp/dp_engine.py
```diff
@@ -110,3 +114,31 @@ def unnest_cross_partition_bound_sampled_per_key(pid_pk_v):
...
+            max_partitions_contributed: how many contributed partitions we can get at 
+                most from this operations.
+        
+        Returns:
+            collection of elements (partition_key, accumulator)
+        """
+        budget = self._budget_accountant.request_budget(weight=1, use_eps=True, use_delta=True)
+        
+        def filter_fn(captures: Tuple[Budget, int], row: Tuple[Any, Accumulator]) -> bool:
+            """Lazily create a partition selection strategy and use it to determine which
```
dvadym, line 133: Nit: create -> creates, use -> uses

[Styleguide](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods) says "The docstring should be descriptive-style"

requirements.dev.txt
```diff
@@ -5,4 +5,5 @@ pyspark
 pytest
 pylint
 yapf
-git+https://github.com/google/differential-privacy.git@327972c1ae710e8cd0a4754fffdd78c3500272ee#subdirectory=python
\ No newline at end of file
+git+https://github.com/google/differential-privacy.git@327972c1ae710e8cd0a4754fffdd78c3500272ee#subdirectory=python
+python-dp
```
dvadym, line 9: Python-DP requires compilation, while PipelineDP not, and it might be inconvenient to build py-dp. Let's make dependency on py-dp pip package (probably now pip package has not yet been updated, but that's not a problem, in 1-2 weeks the new version will be released)

pipeline_dp/dp_engine.py
```diff
@@ -110,3 +114,31 @@ def unnest_cross_partition_bound_sampled_per_key(pid_pk_v):
         return self._ops.flat_map(col,
                                   unnest_cross_partition_bound_sampled_per_key,
                                   "Unnest")
+    
+    def _select_private_partitions(self, col, max_partitions_contributed: int):
+        """Selects and publishes private partitions.
+
+        Args:
+            col: collection, with types for each element: 
+                (partition_key, accumulator: Accumulator)
```
dvadym, line 123: Nit: (partition_key, Accumulator)

pipeline_dp/dp_engine.py
```diff
@@ -1,13 +1,17 @@
...
 
 from dataclasses import dataclass
 from pipeline_dp.aggregate_params import AggregateParams
-from pipeline_dp.budget_accounting import BudgetAccountant
+from pipeline_dp.budget_accounting import BudgetAccountant, Budget
 from pipeline_dp.pipeline_operations import PipelineOperations
 from pipeline_dp.report_generator import ReportGenerator
 
+from pydp.algorithms.partition_selection import create_truncated_geometric_partition_strategy
+from .accumulator import Accumulator
```
dvadym, line 13: Please use absolute paths for import

https://google.github.io/styleguide/pyguide.html#224-decision

Also this style guide decision says import only modules, not objects, but since in this file classes are imported, let's keep style of this file (and then sometime to refactor import in the whole file)

dvadym (commented):
Thank you, it looks good. I left some comments. I think now we are ready to add tests.

Maybe the easiest way is to mock `create_truncated_geometric_partition_strategy` with [unittest.patch](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch). It would allow to avoid stochastic tests. If you see another options please suggest

requirements.dev.txt
levzlotnik: It seems `py-dp` doesn't exist, while `python-dp` doesn't really require compilation as it's published in PyPi as a whole package.

levzlotnik:
Started implementing tests. The draft is already there, but it's failing because it doesn't include the accumulators.
Working on converting the draft data into one with accumulators.

tests/dp_engine_test.py
```diff
@@ -137,6 +153,72 @@ def test_aggregate_report(self):
...
+                                            max_partitions_contributed):
+        with patch("pipeline_dp.dp_engine.create_truncated_geometric_partition_strategy",
+                   new=_mock_partition_strategy_factory(min_users)) as mock_factory:
+            data_filtered = engine._select_private_partitions(groups, max_partitions_contributed)
+            engine._budget_accountant.compute_budgets()
+            self.assertListEqual(list(data_filtered), expected_partitions)
+
+    def test_select_private_partitions(self):
+        input_col = [
+            ("pid1", ( 'pk1', 1 )),
```
dvadym, line 167: You can create MockAccumulator which has private_id_count and create input_col of (partition_key, MockAccumulator). MockAccumulator might be NamedTuple for example.

And then no need to do group_by_key below

requirements.dev.txt
levzlotnik: I'm sorry, I'm not sure what do you mean.  
When we run `pip install python-dp` (or when `python-dp` is listed in `requirements.txt`) it downloads the wheel with the `_pydp.so` ready, so we don't have to compile anything. What is the issue?

tests/dp_engine_test.py
levzlotnik: Done something equivalent.

dvadym: ok, it makes sense.

dvadym (approved):
Thanks!

PR #56, opened 2021-07-12
Added MultiProcLocalPipelineOperations

levzlotnik:
## Description
Created a `multiprocessing` equivalent of `LocalPipelineOperations`, but with `n_jobs`. Most of the functionality should be exactly the same as `LocalPipelineOperations` but run on multi-threaded environment.  
Differences are:
- `MultiProcLocalPipelineOperations.group_by_key`, `samples_fixed_per_key` and `count_per_element` will be **out of order** because the processes contributing to the groups are running out of order, even if `ordered=True` was passed.
- ~We cannot guarantee full laziness of any of the functions. `multiprocessing` strives when we use "partial laziness", i.e. we launch jobs asynchronously and return sometime later to collect the results. During this time - the process pool works to later provide us with immediate results.~ EDIT: No longer true. Implemented classes that trigger iterations only when needed.

## Affected Dependencies
None.

## How has this been tested?
Most tests were taken from the `LocalPipelineOperations` tests, some were modified/removed:
- Anything relying on the order of values for `group_by_key`, `samples_fixed_per_key` and `count_per_element` was modified to not rely on order
- ~Anything relying on the laziness was removed. `multiprocessing.Pool.imap` isn't truly lazy in the full sense of the word.~ EDIT: No longer true. Added tests for laziness.

## Checklist

### DIFF
--- a/pipeline_dp/pipeline_operations.py
+++ b/pipeline_dp/pipeline_operations.py
@@ -1,6 +1,11 @@
 """Adapters for working with pipeline frameworks."""
 
+from functools import partial
+import os
+import multiprocessing as mp
+from . import accumulator
 import random
+from tkinter import W
 import numpy as np
 
 import abc
@@ -323,3 +328,147 @@ class LocalPipelineOperations(PipelineOperations):
 
     def reduce_accumulators_per_key(self, col, stage_name: str = None):
         raise NotImplementedError()
+
+class MultiProcLocalPipeLineOperations(PipelineOperations):
+    def __init__(self, n_jobs: typing.Optional[int]=None,
+                chunksize: int=1,
+                ordered=True, 
+                **pool_kwargs):
+        if n_jobs is None:
+            n_jobs = len(os.sched_getaffinity())
+        self.n_jobs = n_jobs
+        self.chunksize = chunksize
+        self.ordered = ordered
+        self.pool = mp.Pool(n_jobs, **pool_kwargs)
+        if ordered:
+            self._pool_map_fn = self.pool.imap
+        else:
+            self._pool_map_fn = self.pool.imap_unordered
+
+    def map(self, col, fn, stage_name: str):
+        return self._pool_map_fn(fn, col, chunksize=self.chunksize)
+
+    def flat_map(self, col, fn, stage_name: str):
+        return (e for x in self.map(col, fn, stage_name) for e in x)
+
+    def map_tuple(self, col, fn, stage_name: str):
+        def mapped_fn(captures: typing.Tuple[typing.Callable], row):
+            func, = captures
+            return func(*row)
+        mapped_fn = partial(mapped_fn, (fn, ))
+        return self.map(col, mapped_fn, stage_name)
+
+    def map_values(self, col, fn, stage_name: str):
+        def mapped_fn(captures: typing.Tuple[typing.Callable], row):
+            func, = captures
+            return row[0], func(row[1])
+        mapped_fn = partial(mapped_fn, (fn, ))
+        return self.map(col, mapped_fn, stage_name)
+
+    def group_by_key(self, col, stage_name: str):
+        # NOTE - this cannot be implemented in an ordered manner without (almost) serial execution!
+        #   both keys and groups will be out of order
+        with mp.Manager() as manager:
+            results_dict = manager.dict()
+            def insert_row(captures, row):
+                manager_, results_dict_ = captures
+                key, val = row
+                if results_dict_.get(key, None) is None:
+                    results_dict_[key] = manager_.list()
+                results_dict_[key].append(val)
+            insert_row = partial(insert_row, (manager, results_dict))
+            self.pool.map_async(insert_row, col, self.chunksize).wait()
+            return ((k, v) for k, v in results_dict.items())
+
+    def _filter_ordered(self, col, fn, stage_name: str):
+        def mapped_fn(captures: typing.Tuple[typing.Callable], row):
+            func, = captures
+            return row, func(row)
+        mapped_fn = partial(mapped_fn, (fn, ))
+        return (
+            row for row, keep in self.map(col, mapped_fn, stage_name) if keep
+        )
+
+    def _filter_unordered(self, col, fn, stage_name: str):
+        # TODO - implement using multiprocessing.Queue
+        #   details: We want to make the filtering happen on the workers themselves
+        #       rather than on the main thread.
+        return self._filter_ordered(col, fn, stage_name)
+
+    def filter(self, col, fn, stage_name: str):
+        if self.ordered:
+            return self._filter_ordered(col, fn, stage_name)
+        return self._filter_unordered(col, fn, stage_name)
+
+    def _filter_by_key_ordered(self,
+                                col,
+                                public_partitions,
+                                data_extractors,
+                                stage_name: typing.Optional[str] = None):
+        def mapped_fn(captures, row):
+            public_partitions_, data_extractors_ = captures
+            partition = data_extractors_.partition_extractor(row)
+            return row, (partition in public_partitions_)
+        mapped_fn = partial(mapped_fn, (public_partitions, data_extractors))
+
+        return (
+            row for row, keep in self.map(col, mapped_fn, stage_name) if keep
+        )
+
+    def _filter_by_key_unordered(self,
+                                col,
+                                public_partitions,
+                                data_extractors,
+                                stage_name: typing.Optional[str] = None):
+        # TODO - implement using multiprocessing.Queue
+        #   details: We want to make the filtering happen on the workers themselves
+        #       rather than on the main thread.
+        return self._filter_by_key_ordered(col, public_partitions, data_extractors, stage_name)
+
+    def filter_by_key(self,
+                        col,
+                        public_partitions,
+                        data_extractors,
+                        stage_name: typing.Optional[str] = None):
+        if self.ordered:
+            return self._filter_by_key_ordered(col, public_partitions,
+             data_extractors, stage_name)
+        return self._filter_by_key_unordered(col, public_partitions, 
+            data_extractors, stage_name)
+
+    def keys(self, col, stage_name: str):
+        # no point in passing through multiproc.
+        return (k for k, v in col)
+
+    def values(self, col, stage_name: str):
+        # no point in passing through multiproc.
+        return (v for k, v in col)
+
+    def sample_fixed_per_key(self, col, n: int, stage_name: str):
+        def mapped_fn(captures, row):
+            n_, = captures
+            partition_key, values = row
+            samples = values
+            if len(samples) > n_:
+                samples = random.sample(samples, n_)
+            return partition_key, samples
+        mapped_fn = partial(mapped_fn, (n,))
+        groups = self.group_by_key(col, stage_name)
+        return self.map(groups, mapped_fn, stage_name)
+
+    def count_per_element(self, col, stage_name: str):
+        groups = self.group_by_key(col, stage_name)
+        return self.map_values(groups, len, stage_name)
+
+    def reduce_accumulators_per_key(self, col, stage_name: str):
+        """Reduces the input collection so that all elements per each key are merged.
+
+            Args:
+              col: input collection which contains tuples (key, accumulator)
+              stage_name: name of the stage
+
+            Returns:
+              A collection of tuples (key, accumulator).
+
+            """
+        return self.map_values(col, accumulator.merge)
--- a/pipeline_dp/pipeline_operations.py
+++ b/pipeline_dp/pipeline_operations.py
@@ -371,6 +371,23 @@ class _LazyMultiProcIterator:
         n_jobs: typing.Optional[int],
         **pool_kwargs
     ):
+        """Utilizes the `multiprocessing.Pool.map` for distributed execution of 
+        a function `job` on an iterable `job_inputs`.
+
+        Parameters
+        ==========
+        job: 
+            the function to be called on each input
+
+        job_inputs: 
+            iterable containing all the inputs
+
+        chunksize: 
+            see [multiprocessing.Pool.map signature](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.Pool.map).  
+
+        n_jobs: 
+            see [multiprocessing.Pool constructor](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.Pool) arguments
+        """
         self.job = job
         self.chunksize = chunksize
         self.job_inputs = job_inputs
@@ -380,6 +397,7 @@ class _LazyMultiProcIterator:
         self._pool = None
 
     def _init_pool(self):
+        """Creates the multiprocessing.Pool object that will manage the distributed computation."""
         self._pool = mp.Pool(
             self.n_jobs,
             initializer=_pool_worker_init,
@@ -389,6 +407,8 @@ class _LazyMultiProcIterator:
         return self._pool
 
     def _trigger_iterations(self):
+        """Trigger the Pool operation that iterates over inputs and produces outputs.
+        """
         if self._outputs is None:
             self._outputs = self._init_pool().map(
                 _pool_worker, 
@@ -408,6 +428,12 @@ class _LazyMultiProcGroupByIterator(_LazyMultiProcIterator):
                 chunksize: int,
                 n_jobs: typing.Optional[int],
                 **pool_kwargs):
+        """Utilizes mp.Pool for distributed group by computation.
+        The results are held in a `mp.Manager.dict[KeyType, np.Manager.list[ValueType]]`.
+
+        The `mp.Manager.{dict, list}` objects are managed by the `manager` to allow multiprocess-safe
+        access to the containers.
+        """
         self.manager = mp.Manager()
         self.results_dict = self.manager.dict()
         def insert_row(captures, row):
@@ -434,6 +460,12 @@ class _LazyMultiProcCountIterator(_LazyMultiProcIterator):
                 chunksize: int,
                 n_jobs: typing.Optional[int],
                 **pool_kwargs):
+        """Utilizes mp.Pool for distributed group by computation.
+        The results are held in a `mp.Manager.dict[KeyType, int]`.
+
+        The `mp.Manager.dict` object is managed by the `manager` to allow multiprocess-safe
+        access to the container.
+        """
         self.manager = mp.Manager()
         self.results_dict = self.manager.dict()
         def insert_row(captures, key):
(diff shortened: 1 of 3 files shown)


pipeline_dp/pipeline_operations.py
```diff
@@ -323,3 +329,172 @@ def count_per_element(self, col, stage_name: typing.Optional[str] = None):
...
+def _pool_worker_init(func):
+    global _pool_current_func
+    _pool_current_func = func
+
+def _pool_worker(row):
+    return _pool_current_func(row)
+
+class MultiProcLocalPipelineOperations(PipelineOperations):
+    def __init__(self, n_jobs: typing.Optional[int]=None,
+                chunksize: int=1,
```
dvadym, line 345: chunksize=1 means 1 element per task?

pipeline_dp/pipeline_operations.py
```diff
@@ -323,3 +329,172 @@ def count_per_element(self, col, stage_name: typing.Optional[str] = None):
...
+        #     return row[0], func(row[1])
+        # mapped_fn = partial(mapped_fn, (fn, ))
+        return self.map(col, lambda x: (x[0], fn(x[1])), stage_name)
+
+    def group_by_key(self, col, stage_name: typing.Optional[str]=None):
+        # NOTE - this cannot be implemented in an ordered manner without (almost) serial execution!
+        #   both keys and groups will be out of order
+        keys = set(self.keys(col))
+        with mp.Manager() as manager:
+            results_dict = manager.dict({
```
dvadym, line 387: Is manager.dict protected somehow against race conditions?

dvadym (commented):
Thanks it looks cool. I left some comments

pipeline_dp/pipeline_operations.py
levzlotnik: [Yes](https://docs.python.org/3/library/multiprocessing.html#sharing-state-between-processes), it's a proxy for the `dict` object. Internally it's a `mp.Queue` that feeds requests for `dict.__setitem__` to a central server process holding the actual `dict` object.

chinmayshah99:
Just an idea, rather than using `multiprocessing` module, we could use `concurrent.futures` as it has a way cleaner interface.

levzlotnik:
Hi @dvadym & @chinmayshah99
Now this code works and implements laziness as required from the pipeline operations.  
Could you please review it?

pipeline_dp/pipeline_operations.py
```diff
@@ -1,6 +1,11 @@
 """Adapters for working with pipeline frameworks."""
 
-import collections
+from enum import Enum
+from functools import partial
+import os
+import multiprocessing as mp
+from tkinter import W
```
chinmayshah99, line 7: is tkinter needed?

chinmayshah99 (changes requested):

pipeline_dp/pipeline_operations.py
levzlotnik: Was added by accident. Removed this now

chinmayshah99 (approved):

pipeline_dp/pipeline_operations.py
```diff
@@ -349,3 +346,213 @@ def count_per_element(self, col, stage_name: typing.Optional[str] = None):
...
+    def group_by_key(self, col, stage_name: typing.Optional[str] = None):
+        # NOTE - this cannot be implemented in an ordered manner without (almost) serial execution!
+        #   both keys and groups will be out of order
+        return _LazyMultiProcGroupByIterator(col, self.chunksize, 
+                                             self.n_jobs, 
+                                             **self.pool_kwargs)
+
+    def filter(self, col, fn, stage_name: typing.Optional[str] = None):
+        ordered_predicates = self.map(col, fn, stage_name)
+        return (row for row, keep in zip(col, ordered_predicates) if keep)
```
dasmdasm, line 505: Can we guarantee that ordered_predicates and col will have the same order?

pipeline_dp/pipeline_operations.py
```diff
@@ -349,3 +346,213 @@ def count_per_element(self, col, stage_name: typing.Optional[str] = None):
...
+                      public_partitions,
+                      data_extractors,
+                      stage_name: typing.Optional[str] = None):
+        def mapped_fn(captures, row):
+            public_partitions_, data_extractors_ = captures
+            key = data_extractors_.partition_extractor(row)
+            return key, (key in public_partitions_)
+
+        mapped_fn = partial(mapped_fn, (public_partitions, data_extractors))
+        ordered_key_keep = self.map(col, mapped_fn, stage_name)
```
dasmdasm, line 517: Same question about ordering.

pipeline_dp/pipeline_operations.py
```diff
@@ -349,3 +346,213 @@ def count_per_element(self, col, stage_name: typing.Optional[str] = None):
...
+        return _LazyMultiProcGroupByIterator(col, self.chunksize, 
+                                             self.n_jobs, 
+                                             **self.pool_kwargs)
+
+    def filter(self, col, fn, stage_name: typing.Optional[str] = None):
+        ordered_predicates = self.map(col, fn, stage_name)
+        return (row for row, keep in zip(col, ordered_predicates) if keep)
+
+    def filter_by_key(self, col,
+                      public_partitions,
```
dasmdasm, line 508: nit: It's a little weird to be referring to public partitions in a "filter by key" operation, since this is at a lower level than the concept of "partitions."

pipeline_dp/pipeline_operations.py
levzlotnik: Yes, see [`mp.Pool.map`](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.Pool.map) as opposed to [`mp.Pool.imap_unordered`](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.Pool.imap_unordered).

pipeline_dp/pipeline_operations.py
levzlotnik: Same answer.

pipeline_dp/pipeline_operations.py
levzlotnik: This is part of the API that was specified by the `PipelineOperations` abstract base class.

dasmdasm (approved):
LGTM

levzlotnik:
Hey @dvadym @chinmayshah99 @dasmdasm, can we merge it?

tests/pipeline_operations_test.py
```diff
@@ -385,6 +386,254 @@ def test_local_flat_map(self):
...
+
+    @classmethod
+    def setUpClass(cls):
+        cls.ops = MultiProcLocalPipelineOperations(n_jobs=1)
+        cls.data_extractors = DataExtractors(
+            partition_extractor=cls.partition_extract,
+            privacy_id_extractor=cls.privacy_id_extract,
+            value_extractor=cls.value_extract)
+
+    def sortDataset(self, data):
```
dvadym, line 415: Nit: sort_dataset

tests/pipeline_operations_test.py
```diff
@@ -385,6 +386,254 @@ def test_local_flat_map(self):
...
+            return tuple(self.sortDataset(x) for x in data)
+        data = list(data)
+        def sortfn(x):
+            # make everything a sortable thing!
+            if isinstance(x, list):
+                return self.sortDataset(tuple(x))
+            return x
+        return sorted(data, key=sortfn)
+
+    def assertDatasetsEqual(self, first, second, toplevel=True):
```
dvadym, line 428: it looks like we need to compare first and second as multisets. It looks like in Python [Counter](https://docs.python.org/3/library/collections.html#collections.Counter) is best approach for multisets. Wouldn't it work?

pipeline_dp/pipeline_operations.py
```diff
@@ -349,3 +346,213 @@ def count_per_element(self, col, stage_name: typing.Optional[str] = None):
...
+        return Sentinel.EOI
+    return _pool_current_func(row)
+
+
+class _LazyMultiProcIterator:
+    def __init__(
+        self,
+        job: typing.Callable,
+        job_inputs: typing.Iterable,
+        chunksize: int = 1,
```
dvadym, line 388: Nit: it looks like `chunksize` and `n_jobs` are always set in `PipelineOperations`. So no need to make them Optional and set default values here and `_LazyMultiProcGroupByIterator`

pipeline_dp/pipeline_operations.py
```diff
@@ -349,3 +346,213 @@ def count_per_element(self, col, stage_name: typing.Optional[str] = None):
 
     def reduce_accumulators_per_key(self, col, stage_name: str = None):
         return self.map_values(self.group_by_key(col), accumulator.merge)
+
+
+class Sentinel(Enum):
+    EOI = "EndOfIterations"
```
dvadym, line 352: Why do we need Sentinel? If it's neccassary please added explanation in comments

pipeline_dp/pipeline_operations.py
```diff
@@ -349,3 +346,213 @@ def count_per_element(self, col, stage_name: typing.Optional[str] = None):
...
+                chunksize: int,
+                n_jobs: typing.Optional[int],
+                **pool_kwargs):
+        self.manager = mp.Manager()
+        self.results_dict = self.manager.dict()
+        def insert_row(captures, key):
+            (results_dict_,) = captures
+            results_dict_[key] += 1
+
+        insert_row = partial(insert_row, (self.results_dict,))
```
dvadym, line 459: AFAIU using partial is a part of workaround. right?

pipeline_dp/pipeline_operations.py
```diff
@@ -349,3 +346,213 @@ def count_per_element(self, col, stage_name: typing.Optional[str] = None):
...
+    _pool_current_func = func
+
+
+def _pool_worker(row):
+    if row is Sentinel.EOI:
+        return Sentinel.EOI
+    return _pool_current_func(row)
+
+
+class _LazyMultiProcIterator:
```
dvadym, line 383: Could you please add documentation for this class with a short explanation how it works. Something like, that paralellization is achived with using parallel map from multiprocessing.

pipeline_dp/pipeline_operations.py
```diff
@@ -349,3 +346,213 @@ def count_per_element(self, col, stage_name: typing.Optional[str] = None):
...
+
+    def __iter__(self):
+        if isinstance(self.job_inputs, _LazyMultiProcIterator):
+            self.job_inputs._trigger_iterations()
+        self._trigger_iterations()
+        yield from _multiproc_iter_output(self._outputs)
+
+
+class _LazyMultiProcGroupByIterator(_LazyMultiProcIterator):
+    def __init__(self, job_inputs: typing.Iterable,
```
dvadym, line 423: Could you please add documentation for this class with a short explanation how it works? With mentionening that results_dict is used for collecting results between processes and that it protects us from any parallel issues.

pipeline_dp/pipeline_operations.py
```diff
@@ -349,3 +346,213 @@ def count_per_element(self, col, stage_name: typing.Optional[str] = None):
...
+
+    def _trigger_iterations(self):
+        if self._outputs is None:
+            keys = set(k for k, v in self.job_inputs)
+            self.results_dict.update({k: self.manager.list() for k in keys})
+            self._init_pool().map(_pool_worker, self.job_inputs, self.chunksize)
+            self._outputs = ((k, list(v)) for k, v in self.results_dict.items())
+
+class _LazyMultiProcCountIterator(_LazyMultiProcIterator):
+    def __init__(self, job_inputs: typing.Iterable,
```
dvadym, line 449: Could you please add documentation for this class with a short explanation how it works? With mentioning that results_dict is used for collecting results between processes and that it protects us from any parallel issues.

dvadym (commented):
Sry for the late reply, I was on vacation. Thanks, it looks great, I left some comments and questions.

pipeline_dp/pipeline_operations.py
levzlotnik: Yes, it allows the function's state to be well defined when distributing it across processes.

pipeline_dp/pipeline_operations.py
levzlotnik: It was an artifact of previous (unsuccessful) attempts of using `pool.imap_unordered`. Now it's not needed, I removed it.

pipeline_dp/pipeline_operations.py
levzlotnik: Done.

tests/pipeline_operations_test.py
levzlotnik: This is a somewhat generic implementation of comparing two datasets that can hold arbitrary objects. Value types like `int, str, float` are compared as is, tuples are compared by value as well (unless they hold only 2 elements in which case we assume it might be a key-value pair where the value is another dataset), and lists are treated as multisets.  
[`collections.Counter`](https://docs.python.org/3/library/collections.html#collections.Counter) won't work in this case because some of the types I'm handling are **mutable** (lists for example) and hence cannot serve as dictionary keys.

tests/pipeline_operations_test.py
levzlotnik: Done.

tests/pipeline_operations_test.py
levzlotnik: In essence, I implicitly define `Dataset`:
```python
ValueType = int | str | float
DatasetType = ValueType | Tuple[ValueType, 'DatasetType'] | Tuple['DatasetType'] | List['DatasetType']
```
Where:  
1. `ValueType` is compared by value directly.
2. `Tuple[ValueType, 'DatasetType']` compares the first element by value, and the second element using `DatasetType` rules.
3. `Tuple['DatasetType']` compares **ordered elementwise** elements, each element is assumed `DatasetType`.
4. `List['DatasetType']` compares the elements in a "multiset" fashion, where each element is assumed `DatasetType`.

pipeline_dp/pipeline_operations.py
levzlotnik: Done, hope to satisfaction.

pipeline_dp/pipeline_operations.py
levzlotnik: Done, hope to satisfaction.

pipeline_dp/pipeline_operations.py
levzlotnik: Done, hope to satisfaction.

pipeline_dp/pipeline_operations.py
```diff
@@ -349,3 +346,213 @@ def count_per_element(self, col, stage_name: typing.Optional[str] = None):
...
+        mapped_fn = partial(mapped_fn, (n,))
+        groups = self.group_by_key(col, stage_name)
+        return self.map(groups, mapped_fn, stage_name)
+
+    def count_per_element(self, col, stage_name: typing.Optional[str] = None):
+        return _LazyMultiProcCountIterator(col, self.chunksize, 
+                                           self.n_jobs, **self.pool_kwargs)
+
+    def reduce_accumulators_per_key(self, col, stage_name: typing.Optional[str] = None):
+        """Reduces the input collection so that all elements per each key are merged.
```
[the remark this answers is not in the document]
levzlotnik: Done

pipeline_dp/pipeline_operations.py
```diff
@@ -349,3 +346,213 @@ def count_per_element(self, col, stage_name: typing.Optional[str] = None):
...
+
+    def map_tuple(self, col, fn, stage_name: typing.Optional[str] = None):
+        return self.map(col, lambda row: fn(*row), stage_name)
+
+    def map_values(self, col, fn, stage_name: typing.Optional[str] = None):
+        return self.map(col, lambda x: (x[0], fn(x[1])), stage_name)
+
+    def group_by_key(self, col, stage_name: typing.Optional[str] = None):
+        # NOTE - this cannot be implemented in an ordered manner without (almost) serial execution!
+        #   both keys and groups will be out of order
```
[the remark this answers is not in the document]
levzlotnik: Done

pipeline_dp/pipeline_operations.py
```diff
@@ -349,3 +346,218 @@ def count_per_element(self, col, stage_name: typing.Optional[str] = None):
...
+            self.n_jobs,
+            initializer=_pool_worker_init,
+            initargs=(self.job,),
+            **self.pool_kwargs
+        )
+        return self._pool
+
+    def _trigger_iterations(self):
+        """Trigger the Pool operation that iterates over inputs and produces outputs.
+        """
```
dvadym, line 411: Nit: """... """

requirements.dev.txt
```diff
@@ -2,7 +2,11 @@ numpy
 apache-beam
 absl-py
 pyspark
+
+# pytest stuff
```
dvadym, line 6: do we need this line?

tests/pipeline_operations_test.py
dvadym: Thanks it makes sense

pipeline_dp/pipeline_operations.py
```diff
@@ -349,3 +346,218 @@ def count_per_element(self, col, stage_name: typing.Optional[str] = None):
...
+        job: typing.Callable,
+        job_inputs: typing.Iterable,
+        chunksize: int,
+        n_jobs: typing.Optional[int],
+        **pool_kwargs
+    ):
+        """Utilizes the `multiprocessing.Pool.map` for distributed execution of 
+        a function `job` on an iterable `job_inputs`.
+
+        Parameters
```
dvadym, line 377: Nit: according to [Google style guide](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods) (and for consistency with other functions), the format of function comments:
Args:
   ...

pipeline_dp/pipeline_operations.py
dvadym: Thanks, it looks great!

pipeline_dp/pipeline_operations.py
dvadym: Thanks!

pipeline_dp/pipeline_operations.py
dvadym: Thanks!

dvadym (commented):
Just a few minor comments

requirements.dev.txt
chinmayshah99: It added for readability for separating different requiremnts. Not a bad idea.

dvadym: Thanks, SG

dvadym (approved):
Thanks a lot of implementing Multiprocess pipeline operations!

PR #57, opened 2021-07-12
DP computing of variance

AlexandraImbrisca:
## Description
Implementation of [compute_dp_var](https://github.com/OpenMined/PipelineDP/issues/36).

## Affected Dependencies
None.

## How has this been tested?
Unit tests are included in tests\dp_computations_test.py.

## Checklist

### DIFF
--- a/pipeline_dp/dp_computations.py
+++ b/pipeline_dp/dp_computations.py
@@ -38,12 +38,19 @@ class MeanVarParams:
         if metric == pipeline_dp.Metrics.MEAN:
             return self.max_contributions_per_partition * abs(
                 self.middle() - self.low)
+        if metric == pipeline_dp.Metrics.VAR:
+            return self.max_contributions_per_partition * abs(
+                self.middle_squares() - self.low ** 2)
         # TODO: add value for variance
         raise ValueError("Invalid metric")
 
     def middle(self):
         """"Returns the middle point of the interval [low, high]."""
         return self.low + (self.high - self.low) / 2
+
+    def middle_squares(self):
+        """"Returns the middle point of the interval [low^2, high^2]."""
+        return self.low ** 2 + (self.high ** 2 - self.low ** 2) / 2
 
 
 def compute_l1_sensitivity(l0_sensitivity: float, linf_sensitivity: float):
@@ -247,3 +254,56 @@ def compute_dp_mean(count: int, sum: float, dp_params: MeanVarParams):
                                  dp_params.noise_kind)
     dp_mean = dp_normalized_sum / dp_count + middle
     return dp_count, dp_mean * dp_count, dp_mean
+
+
+def compute_dp_var(count: int, sum: float, sum_squares: float,
+                   dp_params: MeanVarParams):
+    """Computes DP variance.
+
+    Args:
+        count: Non-DP count.
+        sum: Non-DP sum.
+        sum_squares: Non-DP sum of squares.
+        dp_params: The parameters used at computing the noise.
+
+    Raises:
+        ValueError: The noise kind is invalid.
+
+    Returns:
+        The tuple of anonymized count, sum, sum_squares and variance.
+    """
+    sum_middle = dp_params.middle()
+    sum_squares_middle = dp_params.middle_squares()
+
+    normalized_sum = sum - count * sum_middle
+    normalized_sum_squares = sum_squares - count * sum_squares_middle
+
+    l0_sensitivity = dp_params.l0_sensitivity()
+
+    # Splits the budget equally between the two mechanisms.
+    (count_eps, count_delta), (sum_eps, sum_delta), (
+        sum_squares_eps, sum_squares_delta) = equally_split_budget(
+        dp_params.eps, dp_params.delta, 3)
+
+    dp_count = _add_random_noise(count, count_eps, count_delta, l0_sensitivity,
+                                 dp_params.linf_sensitivity(
+                                     pipeline_dp.Metrics.COUNT),
+                                 dp_params.noise_kind)
+    dp_normalized_sum = _add_random_noise(normalized_sum, sum_eps, sum_delta,
+                                          l0_sensitivity,
+                                          dp_params.linf_sensitivity(
+                                              pipeline_dp.Metrics.MEAN),
+                                          dp_params.noise_kind)
+    dp_normalized_sum_squares = _add_random_noise(normalized_sum_squares,
+                                                  sum_squares_eps,
+                                                  sum_squares_delta,
+                                                  l0_sensitivity,
+                                                  dp_params.linf_sensitivity(
+                                                      pipeline_dp.Metrics.VAR),
+                                                  dp_params.noise_kind)
+
+    dp_mean = dp_normalized_sum / dp_count + sum_middle
+    dp_mean_squares = dp_normalized_sum_squares / dp_count + sum_squares_middle
+    dp_var = dp_mean_squares - dp_mean ** 2
+
+    return dp_count, dp_mean * dp_count, dp_mean_squares * dp_count, dp_var
--- a/pipeline_dp/dp_computations.py
+++ b/pipeline_dp/dp_computations.py
@@ -290,9 +290,10 @@ def compute_dp_var(count: int, sum: float, sum_squares: float,
                                  dp_params.max_contributions_per_partition,
                                  dp_params.noise_kind)
 
-    # Computes and adds noise to the mean.
     linf_sensitivity = dp_params.max_contributions_per_partition * abs(
         compute_middle(dp_params.low, dp_params.high) - dp_params.low)
+
+    # Computes and adds noise to the mean.
     dp_mean = _compute_mean(count, dp_count, sum, dp_params.low, dp_params.high,
                             sum_eps, sum_delta, l0_sensitivity,
                             linf_sensitivity, dp_params.noise_kind)
--- a/pipeline_dp/dp_computations.py
+++ b/pipeline_dp/dp_computations.py
@@ -39,9 +39,10 @@ class MeanVarParams:
             return self.max_contributions_per_partition * abs(
                 self.middle() - self.low)
         if metric == pipeline_dp.Metrics.VAR:
-            return self.max_contributions_per_partition * abs(
-                self.middle_squares() - self.low ** 2)
-        # TODO: add value for variance
+            max_contribution = max(self.low**2, self.high
+                                   **2) if self.low * self.high < 0 else abs(
+                                       self.high**2 - self.low**2)
+            return self.max_contributions_per_partition * max_contribution
         raise ValueError("Invalid metric")
 
     def middle(self):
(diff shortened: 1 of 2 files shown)


pipeline_dp/dp_computations.py
```diff
@@ -38,13 +38,20 @@ def linf_sensitivity(self, metric):
...
         # TODO: add value for variance
         raise ValueError("Invalid metric")
 
     def middle(self):
         """"Returns the middle point of the interval [low, high]."""
         return self.low + (self.high - self.low) / 2
 
+    def middle_squares(self):
+        """"Returns the middle point of the interval [low^2, high^2]."""
+        return self.low ** 2 + (self.high ** 2 - self.low ** 2) / 2
```
dasmdasm, line 53: This is incorrect. As an example, imagine low = -5, high = 5. This will return 25 + (25 - 25) / 2 = 25. But actually, the minimum squared value is 0 (0 ** 2), and the maximum squared value is 25 (5 ** 2 or -5 **2), so the correct midpoint is 12.5.

pipeline_dp/dp_computations.py
```diff
@@ -247,3 +254,56 @@ def compute_dp_mean(count: int, sum: float, dp_params: MeanVarParams):
...
+                                                  sum_squares_eps,
+                                                  sum_squares_delta,
+                                                  l0_sensitivity,
+                                                  dp_params.linf_sensitivity(
+                                                      pipeline_dp.Metrics.VAR),
+                                                  dp_params.noise_kind)
+
+    dp_mean = dp_normalized_sum / dp_count + sum_middle
+    dp_mean_squares = dp_normalized_sum_squares / dp_count + sum_squares_middle
+    dp_var = dp_mean_squares - dp_mean ** 2
```
dasmdasm, line 307: There's a lot of code duplication between this and compute_dp_mean. Can we share code? In particular, it seems like we should be able to have a function that takes:
- A raw sum
- A noisy count
- An upper and a lower bound

and use it to calculate the means in both variance and mean and the mean of squares in variance.

pipeline_dp/dp_computations.py
```diff
@@ -38,13 +38,20 @@ def linf_sensitivity(self, metric):
         if metric == pipeline_dp.Metrics.MEAN:
             return self.max_contributions_per_partition * abs(
                 self.middle() - self.low)
+        if metric == pipeline_dp.Metrics.VAR:
+            return self.max_contributions_per_partition * abs(
+                self.middle_squares() - self.low ** 2)
```
dasmdasm, line 43: This is incorrect. self.low ** 2 is not the smallest squared value if self.low < 0.

pipeline_dp/dp_computations.py
```diff
@@ -247,3 +254,56 @@ def compute_dp_mean(count: int, sum: float, dp_params: MeanVarParams):
                                  dp_params.noise_kind)
     dp_mean = dp_normalized_sum / dp_count + middle
     return dp_count, dp_mean * dp_count, dp_mean
+
+
+def compute_dp_var(count: int, sum: float, sum_squares: float,
```
dasmdasm, line 259: There's another trick we can use for even more efficient variance calculations. The variance of [x_0, x_1, ..., x_n] is the same as the variance of [x_0 - k, x_1 - k, ... , x_n - k]. In other words, the variance doesn't change when you translate all the elements in the dataset by a fixed amount.

Right now you're calculating the mean by normalizing all the elements to the midpoint, and calculating the mean of squares by normalizing all elements to the midpoint of squares. E.g.

for each input:
  sum += input - midpoint
  sum_of_squares += input ** 2 - midpoint_of_squares(lower, upper)

noisy_sum = add_noise(sum)
noisy_sum_of_squares = add_noise(sum)

noisy_mean = noisy_sum / noisy_count + midpoint
noisy_mean_of_squares = noisy_sum_of_squares / noisy_count + midpoint_of_squares(lower, upper)

Instead, we can translate the elements so they're centered around 0 before doing the rest of the calculation, e.g.

for each input:
  centered_input = input - midpoint
  sum += centered_input  # The midpoint is already 0.
  sum_of_squares += centered_input ** 2 - midpoint_of_squares(lower - midpoint, upper - midpoint)

  noisy_sum = add_noise(sum)
  noisy_sum_of_squares = add_noise(sum_of_squares)
  
  noisy_mean = noisy_sum / noisy_count
  noisy_mean_of_squares = noisy_sum_of_squares / noise_count + midpoint_of_squares(lower - midpoint, upper - midpoint)

Then we use those values to calculate the variance (note we don't un-normalize the mean). We can also post-process the noisy mean to shift it back to the original range. We probably don't need to implement that in this version, but it's something to consider adding in the future.

dasmdasm (changes requested):

pipeline_dp/dp_computations.py
AlexandraImbrisca: Thanks for the review! :) About these more efficient calculations, should I create a new issue/pull request with the implementation?

dvadym: Yeah, let's do it in the subsequent CL

pipeline_dp/dp_computations.py
```diff
@@ -39,9 +39,10 @@ def linf_sensitivity(self, metric):
             return self.max_contributions_per_partition * abs(
                 self.middle() - self.low)
         if metric == pipeline_dp.Metrics.VAR:
-            return self.max_contributions_per_partition * abs(
-                self.middle_squares() - self.low ** 2)
-        # TODO: add value for variance
+            max_contribution = max(self.low**2, self.high
+                                   **2) if self.low * self.high < 0 else abs(
+                                       self.high**2 - self.low**2)
+            return self.max_contributions_per_partition * max_contribution
```
dasmdasm, line 45: High-level concern: This method doesn't really make sense. There's no such thing as the linf sensitivity of a mean or a variance because each is created by post-processing several differently noised quantities. We're using these to mean "the sensitivity of a normalized sum" or "the sensitivity of a normalized sum of squares." I suspect the difference between the stated purpose of the method and what it's actually doing may cause confusion.

Also, calculating sensitivity like this makes things more error-prone because we have to keep this code in sync with the code that actually performs the aggregations. If we change the normalization process, we also have to modify the code here. Separating this calculation out makes it more likely that we'll forget to update both things.

I know this isn't a problem unique to this PR, but I think it becomes particularly bad for variance. Maybe now is an opportunity to improve things?

dasmdasm (changes requested):

pipeline_dp/dp_computations.py
AlexandraImbrisca: Sure, thanks for suggesting this! I think it was clearer with count/sum when noise was added to only one value, but it definitely became confusing with mean/var. For now, I adapted the functions to directly set the Linf sensitivity for each mechanism. What do you think?

pipeline_dp/dp_computations.py
```diff
@@ -215,39 +193,40 @@ def compute_dp_sum(sum: float, dp_params: MeanVarParams):
         ValueError: The noise kind is invalid.
     """
     l0_sensitivity = dp_params.l0_sensitivity()
-    linf_sensitivity = dp_params.linf_sensitivity(pipeline_dp.Metrics.SUM)
+    linf_sensitivity = dp_params.max_contributions_per_partition * max(
+        dp_params.low, dp_params.high)
```
dasmdasm, line 197: I think you need to abs these.

pipeline_dp/dp_computations.py
```diff
@@ -271,11 +250,15 @@ def compute_dp_mean(count: int, sum: float, dp_params: MeanVarParams):
...
 
     dp_count = _add_random_noise(count, count_eps, count_delta, l0_sensitivity,
-                                 dp_params.linf_sensitivity(
-                                     pipeline_dp.Metrics.COUNT),
+                                 dp_params.max_contributions_per_partition,
                                  dp_params.noise_kind)
-    dp_mean = _compute_mean(count, dp_count, sum, sum_eps, sum_delta,
-                            l0_sensitivity, pipeline_dp.Metrics.MEAN, dp_params)
+
+    linf_sensitivity = dp_params.max_contributions_per_partition * abs(
```
dasmdasm, line 256: nit: maybe "sum_linf_sensitivity"?

pipeline_dp/dp_computations.py
```diff
@@ -215,39 +193,40 @@ def compute_dp_sum(sum: float, dp_params: MeanVarParams):
...
     Args:
         count: Non-DP count.
         dp_count: DP count.
         sum: Non-DP sum.
+        low, high: The lowest/highest contribution.
         eps, delta: The budget allocated.
         l0_sensitivity: The L0 sensitivity.
-        metric: The metric used at computing the Linf sensitivity.
-        dp_params: The parameters used at computing the noise.
+        linf_sensitivity: The Linf sensitivity.
```
dasmdasm, line 216: Why not pass in max_contributions_per_partition instead and compute linf_sensitivity in this method? That way the caller doesn't have to account for the normalization.

tests/dp_computations_test.py
```diff
@@ -198,6 +170,7 @@ def test_compute_dp_mean(self):
                         _) = pipeline_dp.dp_computations.equally_split_budget(
                             params.eps, params.delta, 2)
         l0_sensitivity = params.l0_sensitivity()
+        linf_sensitivity = params.max_contributions_per_partition
```
dasmdasm, line 173: nit: maybe "count_linf_sensitivity"?

tests/dp_computations_test.py
```diff
@@ -242,6 +213,7 @@ def test_compute_dp_var(self):
             _, _) = pipeline_dp.dp_computations.equally_split_budget(
                 params.eps, params.delta, 3)
         l0_sensitivity = params.l0_sensitivity()
+        linf_sensitivity = params.max_contributions_per_partition
```
dasmdasm, line 216: nit: maybe "count_linf_sensitivity"?

pipeline_dp/dp_computations.py
```diff
@@ -298,20 +298,15 @@ def compute_dp_var(count: int, sum: float, sum_squares: float,
...
-
-    if dp_params.low < 0 and dp_params.high > 0:
-        linf_sensitivity = dp_params.max_contributions_per_partition * max(
-            dp_params.low**2, dp_params.high**2)
-    else:
-        linf_sensitivity = dp_params.max_contributions_per_partition * abs(
-            dp_params.high**2 - dp_params.low**2)
+    squares_low, squares_high = dp_params.squares_interval()
+    linf_sensitivity = dp_params.max_contributions_per_partition * abs(
+        squares_high - squares_low)
```
dasmdasm, line 303: This doesn't account for normalization.

dasmdasm (changes requested):
Overall, looking good! I think after this we should be done.

pipeline_dp/dp_computations.py
dasmdasm: This seems like a good solution. It's much clearer now. Thanks!

pipeline_dp/dp_computations.py
AlexandraImbrisca: My bad, thanks for spotting this!

pipeline_dp/dp_computations.py
AlexandraImbrisca: Sure, sounds good since the sensitivity is also computed similarly.

dasmdasm (approved):

dasmdasm:
Thanks for all your work, this looks good!

dvadym (approved):
Thanks a lot for implementing DP variance!

Issue #43, opened 2021-06-10
Vector Summation accumulator


dvadym:
## Context

Currently the [Accumulator Framework](https://github.com/OpenMined/PipelineDP/issues/37) development is in progress. 
The accumulator is the object that accumulates values and produces some the aggregated result. The computing of all Differential Private (DP) metrics will be done in corresponding accumulators.

## Goals

To implement class `VectorSummationAccumulator` which computes a Differential Private sum of input vectors of the same size. 
`VectorSummationAccumulator` should be a sub-class of [Accumulator](https://github.com/OpenMined/PipelineDP/blob/5f8e2b5e1265af6e29f3c2fb79d0c9846a63ba43/pipeline_dp/accumulator.py#L12) class.

### Example of aggregations

The input is [(1, 0, 0), (0, 1, 0), (0, 1, 0), (0, 0, 1)].

The output should be (1, 2, 1) + the random noise vector

Note, that it's possible to compute histograms using vector summation primitive.

## API

```
@dataclass
class VectorSummationParams:
   noise:NoiseKind
   max_norm:float


class VectorSummationAccumulator(Accumulator): 
    def __init__(params:VectorSummationParams, values: Iterable[Tuple[double]])
    def add_value(self, value:Tuple[double])  # returns self
    def add_accumulator(self, accumulator: 'VectorSummationAccumulator')  # returns self
    def compute_metrics(self) -> Tuple[double]
```

## Stages of the development

1. Implement vector summation computation.
2. Add anonymization to output (this is better to do in a separate PR)

lovroprepolec:
I am interested in implementing this feature.

dvadym:
Sure, go ahead! Thanks!

dvadym:
FYI: API of the class was updated (to represent the current development in accumulator framework)

lovroprepolec:
Ack. Thx.

levzlotnik:
Hey, I'd like to take this

dvadym:
Sure, thank you!

levzlotnik:
Hey, this was implemented in #67, we can close this issue now.

dvadym:
levzlotnik@, could you please do plumbing for Vector summation in order to be able to run Vector Summation E2E?

[Here](https://github.com/OpenMined/PipelineDP/pull/87/files) (look at the first change in accumulator.py) is an example of doing this for sum

PR #64, opened 2021-08-05
Added Vector Summation Accumulator

levzlotnik:
## Description
Added `VectorSummationAccumulator` as per #43.  

## Note
Only implemented the summation part. The anonymization part will be implemented in a separate PR as requested by @dvadym .

## Affected Dependencies
None.

## How has this been tested?
Added some test cases in `tests/accumulator_tests.py` next to the generic ones.

## Checklist

### DIFF
--- a/pipeline_dp/accumulator.py
+++ b/pipeline_dp/accumulator.py
@@ -46,7 +46,7 @@ class Accumulator(abc.ABC):
     """
         pass
 
-    def _check_accumulator(self, accumulator: 'Accumulator'):
+    def _check_mergeable(self, accumulator: 'Accumulator'):
         if not isinstance(accumulator, type(self)):
             raise TypeError(
                 f"The accumulator to be added is not of the same type: "
@@ -105,7 +105,7 @@ class CompoundAccumulator(Accumulator):
 
     The expectation is that the internal accumulators are of the same type and
     are in the same order."""
-
+        self._check_mergeable(accumulator)
         if len(accumulator.accumulators) != len(self.accumulators):
             raise ValueError(
                 "Accumulators in the input are not of the same size." +
@@ -176,6 +176,7 @@ class CountAccumulator(Accumulator):
 
     def add_accumulator(self,
                         accumulator: 'CountAccumulator') -> 'CountAccumulator':
+        self._check_mergeable(accumulator)
         self._count += accumulator._count
         return self
 
@@ -195,7 +196,7 @@ class VectorSummationAccumulator(Accumulator):
 
     def __init__(self, params: VectorSummationParams,
                  values: Iterable[Tuple[float]]) -> None:
-        self.sum_params = params
+        self._params = params
         values = iter(values)
         try:
             self._vec_sum = np.array(next(values))
@@ -218,11 +219,12 @@ class VectorSummationAccumulator(Accumulator):
     def add_accumulator(
         self, accumulator: 'VectorSummationAccumulator'
     ) -> 'VectorSummationAccumulator':
-        self._check_accumulator(accumulator)
+        self._check_mergeable(accumulator)
         self.add_value(accumulator._vec_sum)
         return self
 
     def compute_metrics(self):
+        # TODO - add DP anonymization
         return tuple(self._vec_sum)
 
 
@@ -240,6 +242,7 @@ class SumAccumulator(Accumulator):
 
     def add_accumulator(self,
                         accumulator: 'SumAccumulator') -> 'SumAccumulator':
+        self._check_mergeable(accumulator)
         self._sum += accumulator._sum
 
     def compute_metrics(self) -> float:
--- a/pipeline_dp/accumulator.py
+++ b/pipeline_dp/accumulator.py
@@ -4,9 +4,10 @@ import pickle
 from dataclasses import dataclass
 from functools import reduce
 
-from typing import Iterable, Tuple
+from typing import Iterable, Tuple, Union
 import pipeline_dp
-from pipeline_dp.aggregate_params import NoiseKind
+# from pipeline_dp.aggregate_params import NoiseKind
+from pipeline_dp import aggregate_params
 import numpy as np
 
 
@@ -187,26 +188,32 @@ class CountAccumulator(Accumulator):
 
 @dataclass
 class VectorSummationParams:
-    noise_kind: NoiseKind
+    noise_kind: aggregate_params.NoiseKind
     max_norm: float
 
+_FloatVector = Union[Tuple[float], np.ndarray]
 
 class VectorSummationAccumulator(Accumulator):
     _vec_sum: np.ndarray
 
     def __init__(self, params: VectorSummationParams,
-                 values: Iterable[Tuple[float]]) -> None:
+                 values: Iterable[_FloatVector]) -> None:
         self._params = params
         values = iter(values)
         try:
-            self._vec_sum = np.array(next(values))
+            value = next(values)
+            if not isinstance(value, np.ndarray):
+                value = np.array(value)
+            self._vec_sum = value
             for val in values:
                 self.add_value(val)
         except StopIteration:
             self._vec_sum = None
 
-    def add_value(self, value: Tuple[float]):
-        value = np.array(value)  # type: np.ndarray
+    def add_value(self, value: _FloatVector):
+        if not isinstance(value, np.ndarray):
+            value = np.array(value)
+
         if self._vec_sum is None:
             self._vec_sum = value
         else:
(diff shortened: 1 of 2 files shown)


pipeline_dp/accumulator.py
```diff
@@ -174,6 +185,47 @@ def compute_metrics(self) -> float:
...
+    noise_kind: NoiseKind
+    max_norm: float
+
+
+class VectorSummationAccumulator(Accumulator):
+    _vec_sum: np.ndarray
+
+    def __init__(self, params: VectorSummationParams,
+                 values: Iterable[Tuple[float]]) -> None:
+        self.sum_params = params
```
dvadym, line 198: Nit: _sum_params  (maybe simply _params)

pipeline_dp/accumulator.py
```diff
@@ -174,6 +185,47 @@ def compute_metrics(self) -> float:
...
+        values = iter(values)
+        try:
+            self._vec_sum = np.array(next(values))
+            for val in values:
```
dvadym, line 202: values is already iterable, why not simply
`for val in values: `
w/o processing of StopIteration ?

We can assume that values is not empty (this object is created after GroupBy by key for each key which is present)

pipeline_dp/accumulator.py
```diff
@@ -174,6 +185,47 @@ def compute_metrics(self) -> float:
...
+                self.add_value(val)
+        except StopIteration:
+            self._vec_sum = None
+
+    def add_value(self, value: Tuple[float]):
+        value = np.array(value)  # type: np.ndarray
```
dvadym, line 208: I'm thinking, whether np.array is better (less space, more effective?) than tuple for serializing and sending between workers, don't you know?

pipeline_dp/accumulator.py
```diff
@@ -42,6 +46,13 @@ def add_value(self, value):
     """
         pass
 
+    def _check_accumulator(self, accumulator: 'Accumulator'):
```
dvadym, line 49: Maybe _check_is_mergeable? or _can_merge?

tests/accumulator_test.py
```diff
@@ -198,9 +218,7 @@ def add_value(self, v):
 
     def add_accumulator(self,
                         accumulator: 'MeanAccumulator') -> 'MeanAccumulator':
-        if not isinstance(accumulator, MeanAccumulator):
-            raise TypeError(
-                "The accumulator to be added is not of the same type.")
+        self._check_accumulator(accumulator)
```
dvadym, line 221: Thanks for improving existing code!

dvadym (commented):
Thanks it looks good. I left some comments and questions

pipeline_dp/accumulator.py
levzlotnik: `np.ndarray` is hands down better. Less memory, enjoys low level and hardware vectorization. I'm not certain if all serializations on it are as compact, but at the very least:
```python
import numpy as np
import pickle

# probably around 8000 byes of memory including metadata like dtype, shape and strides
x = np.arange(1000, dtype=np.int64)

x_p = pickle.dumps(x)
print(len(x_p)) # 8152

x_t = tuple(x) # same array but in tuple
x_t_p = pickle.dumps(x_t)
print(len(x_p)) # 19100
```
My assumption is that numpy is extremely optimized for computation and data movement, whether it's with or without multiprocessing.

tests/accumulator_test.py
levzlotnik: No problem :)

pipeline_dp/accumulator.py
levzlotnik: In the tests we have empty accumulators like `XAccumulator(None, []).add_value(x)`.  
IMHO this is an edge case that must be treated in an expected manner. For example, a user that wrote some custom code for joining two tables and then grouped by some multi-key might make a mistake in implementation, and they should be able to get an indication of what went wrong.  
We can either leave the aggregated value as `None` or throw an error. @dvadym up to you :)

pipeline_dp/accumulator.py
levzlotnik: Done.

pipeline_dp/accumulator.py
levzlotnik: `_check_mergeable` is good enough?

pipeline_dp/accumulator.py
dvadym: Wow. Thanks for checking! Yeah, np arrays look much better

pipeline_dp/accumulator.py
```diff
@@ -174,6 +186,48 @@ def compute_metrics(self) -> float:
...
+class VectorSummationParams:
+    noise_kind: NoiseKind
+    max_norm: float
+
+
+class VectorSummationAccumulator(Accumulator):
+    _vec_sum: np.ndarray
+
+    def __init__(self, params: VectorSummationParams,
+                 values: Iterable[Tuple[float]]) -> None:
```
dvadym, line 198: np.arrays are used very widely in Python and they are so much effective,  that it would be very ineffective for API clients which have data in np.arrays to convert to tuples. So I think it's worth to allow value to be not only tuple but np.array as well.

pipeline_dp/accumulator.py
```diff
@@ -3,7 +3,11 @@
 import pickle
 from dataclasses import dataclass
 from functools import reduce
+
+from typing import Iterable, Tuple
 import pipeline_dp
+from pipeline_dp.aggregate_params import NoiseKind
```
dvadym, line 9: Nit: import only modules (according to [styleguide](https://google.github.io/styleguide/pyguide.html#22-imports) )

pipeline_dp/accumulator.py
levzlotnik: Even though it's only a type hint and the users are free to use `np.ndarray` - I agree with that point. We should allow users (or even soft-requiring them) to use `np.ndarray`.

pipeline_dp/accumulator.py
levzlotnik: Done

pipeline_dp/accumulator.py
levzlotnik: Done

tests/accumulator_test.py
```diff
@@ -129,14 +129,34 @@ def test_merge_accumulators(self):
 
         self.assertEqual(merged_accumulator.compute_metrics(), 10)
 
+        vec_sum_accumulator1 = accumulator.VectorSummationAccumulator(
+            params=None,
+            values=[(15, 2)])
+        vec_sum_accumulator2 = accumulator.VectorSummationAccumulator(
+            params=None,
+            values=[(27, 40)])
```
dvadym, line 137: Please use np.array in some of the input, to test that np.arrays are processed correctly

pipeline_dp/accumulator.py
```diff
@@ -174,6 +187,56 @@ def compute_metrics(self) -> float:
...
+    noise_kind: aggregate_params.NoiseKind
+    max_norm: float
+
+_FloatVector = Union[Tuple[float], np.ndarray]
+
+class VectorSummationAccumulator(Accumulator):
+    _vec_sum: np.ndarray
+
+    def __init__(self, params: VectorSummationParams,
+                 values: Iterable[_FloatVector]) -> None:
```
dvadym, line 200: As mentioned in a comment below, let's assume that values is List (and then it would be possible to simplify the code)

dvadym (commented):
Thanks, it looks great, just a few small comments

pipeline_dp/accumulator.py
levzlotnik: I just noticed that during `add_value` I check for `self._vec_sum is None` - I can easily just loop over `values` without assuming anything except that `values` is an iterable. No need to assume it's a `List`, also I think it's better to leave as `Iterable` as maybe in the future we'd like to do total aggregation over the entire dataset without group-by key.

pipeline_dp/accumulator.py
```diff
@@ -3,7 +3,12 @@
 import pickle
 from dataclasses import dataclass
 from functools import reduce
+
+from typing import Iterable, Tuple, Union
 import pipeline_dp
+# from pipeline_dp.aggregate_params import NoiseKind
```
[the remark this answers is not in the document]
levzlotnik: Done.

tests/accumulator_test.py
levzlotnik: Done.

pipeline_dp/accumulator.py
dvadym: Ok, it makes sense! I agree we don't need to limit ourselves to list w/o necessity.

dvadym (approved):
Thanks!

PR #65, opened 2021-08-06
Implement Accumulator Framework for aggregation - integration with dp_engine and creation of accumulator params

preethiraghavan1:
## Description
Implementing
1. create_accumulator_params for CountAccumulator
2. Integrating accumulator framework into the pipeline computations

This is still in progress. Sending it for review for feedback. Will write tests later.

## Affected Dependencies
List any dependencies that are required for this change.

## How has this been tested?
- Tests will follow. This is for initial review.

## Checklist
- [x ] I have followed the [Contribution Guidelines](https://github.com/OpenMined/.github/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://github.com/OpenMined/.github/blob/master/CODE_OF_CONDUCT.md)
- [ x] I have commented my code following the [OpenMined Styleguide](https://github.com/OpenMined/.github/blob/master/STYLEGUIDE.md)
- [ x] I have labeled this PR with the relevant [Type labels](https://github.com/OpenMined/.github/labels?q=Type%3A)

### DIFF
--- a/pipeline_dp/accumulator.py
+++ b/pipeline_dp/accumulator.py
@@ -2,11 +2,25 @@ import abc
 import typing
 import pickle
 from functools import reduce
+from dataclasses import dataclass
+import pipeline_dp
 
+@dataclass
+class AccumulatorParams:
+    accumulator_type: type = None
+    constructor_params: typing.Any = None
 
 def merge(accumulators: typing.Iterable['Accumulator']) -> 'Accumulator':
   """Merges the accumulators."""
   return reduce(lambda acc1, acc2: acc1.add_accumulator(acc2), accumulators)
+
+
+def create_accumulator_params(aggregation_params:
+  pipeline_dp.AggregateParams,  budget_accountant: pipeline_dp.BudgetAccountant
+                              ) -> typing.List[
+  AccumulatorParams]:
+
+  raise NotImplemented()  # implementation will be done later
 
 
 class Accumulator(abc.ABC):
@@ -100,3 +114,24 @@ class CompoundAccumulator(Accumulator):
     """Computes and returns a list of metrics computed by internal
     accumulators."""
     return [accumulator.compute_metrics() for accumulator in self.accumulators]
+
+
+class AccumulatorFactory:
+  def __init__(self, params: pipeline_dp.AggregateParams,
+               budget_accountant: pipeline_dp.BudgetAccountant):
+    self._params = params
+    self._budget_accountant = budget_accountant
+
+  def initialize(self):
+    self._accumulator_params = create_accumulator_params(self._params, self._budget_accountant)
+
+  def create(self, values: typing.List) -> Accumulator:
+    accumulators = [accumulator_param.accumulator_type(
+      accumulator_param.constructor_params, values) for
+                    accumulator_param, value in zip(
+      self._accumulator_params, values)]
+
+    if len(accumulators) == 1:
+      return accumulators[0]
+
+    return CompoundAccumulator(accumulators)
--- a/pipeline_dp/accumulator.py
+++ b/pipeline_dp/accumulator.py
@@ -5,44 +5,46 @@ from functools import reduce
 from dataclasses import dataclass
 import pipeline_dp
 
+
 @dataclass
 class AccumulatorParams:
     accumulator_type: type = None
     constructor_params: typing.Any = None
 
+
 def merge(accumulators: typing.Iterable['Accumulator']) -> 'Accumulator':
-  """Merges the accumulators."""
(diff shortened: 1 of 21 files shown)


preethiraghavan1:
Hi @dvadym ,

I added just the logic here in this PR. Let me know if this is in the right direction. I will add the test cases after the feedback.

Thanks
Preethi

pipeline_dp/dp_engine.py
```diff
@@ -50,7 +51,22 @@ def aggregate(self, col, params: AggregateParams,
...
+        accumulator_factory.initialize()
+        aggregator_fn = accumulator_factory.create
+
+        # extract the columns
+        col = self._ops.map_tuple(
+            col, lambda row: (data_extractors.privacy_id_extractor(row),
+                              data_extractors.partition_extractor(row),
+                              data_extractors.value_extractor(row)),
+            "Extract (privacy_id, partition_key, value))")
+        result = self._bound_contributions(
```
dvadym, line 67: Nit: result -> col (and below) (it's not yet result)

pipeline_dp/dp_engine.py
```diff
@@ -50,7 +51,22 @@ def aggregate(self, col, params: AggregateParams,
...
+            col, params.max_partitions_contributed,
+            params.max_contributions_per_partition, aggregator_fn)
```
dvadym, line 69: Please add comment with types afterwards, something like
\# col : ((privacy_id, partition_key), accumulator)

pipeline_dp/dp_engine.py
```diff
@@ -50,7 +51,22 @@ def aggregate(self, col, params: AggregateParams,
...
+                                                 budget_accountant=None)
+        accumulator_factory.initialize()
+        aggregator_fn = accumulator_factory.create
+
+        # extract the columns
+        col = self._ops.map_tuple(
+            col, lambda row: (data_extractors.privacy_id_extractor(row),
+                              data_extractors.partition_extractor(row),
+                              data_extractors.value_extractor(row)),
+            "Extract (privacy_id, partition_key, value))")
```
dvadym, line 66: Please add comment with types afterwards, something like
\# col : (privacy_id, partition_key, value)

pipeline_dp/dp_engine.py
```diff
@@ -38,7 +39,7 @@ def _add_report_stage(self, text):
         self._report_generators[-1].add_stage(text)
 
     def aggregate(self, col, params: AggregateParams,
-                  data_extractors: DataExtractors):  # pylint: disable=unused-argument
+                  data_extractors: DataExtractors):
```
dvadym, line 42: Please add a test for this function. Since it's not finished and the result will change in future it doesn't make sense yet to check results. But we can check that the computation graph is correct, i.e. that all needed functions are called with correct parameters. e.g. we can override _bound_contributions and check that arguments is correct (and the same with other methods used in this function)

dvadym (commented):
Thanks a lot! The implementation looks great. I think it's ready for writing tests. I left some comments.

pipeline_dp/accumulator.py
```diff
@@ -22,8 +22,15 @@ def create_accumulator_params(
     aggregation_params: pipeline_dp.AggregateParams,
     budget_accountant: pipeline_dp.BudgetAccountant
 ) -> typing.List[AccumulatorParams]:
-
-    raise NotImplemented()  # implementation will be done later
+    accumulator_params = []
+    if pipeline_dp.Metrics.COUNT in aggregation_params.metrics:
```
dvadym, line 26: Please add test for this function

tests/dp_engine_test.py
```diff
@@ -152,7 +155,13 @@ def test_aggregate_none(self):
         self.assertIsNone(
             pipeline_dp.DPEngine(None, None).aggregate(None, None, None))
 
-    def test_aggregate_report(self):
+    @patch('pipeline_dp.accumulator.create_accumulator_params')
+    def test_aggregate_report(self, mock_create_accumulator_params_function):
```
dvadym, line 159: Thanks for extending this test! In future when Reports will be added we will be able just to add checks on the content of _report_generators.

For context, AggregationReports will provide for the library user a way to understand how computations were performed (eg. which noise is added etc)

dvadym (commented):
Thanks, it looks great. Just one comment below

tests/dp_engine_test.py
preethiraghavan1: Got it! Thanks!

pipeline_dp/accumulator.py
preethiraghavan1: Done!

dvadym (approved):
Thank you!

PR #66, opened 2021-08-06
Add noise to SumAccumulator.

### LINKED ISSUE #61: DP Sum Accumulator
## Goals

To implement class `SumAccumulator` which computes a Differential Private sum of elements. 
`SumAccumulator` should be a sub-class of [Accumulator](https://github.com/OpenMined/PipelineDP/blob/84bbf45740dcdc197c2c20a75b8eaaf6b1f2ecab/pipeline_dp/accumulator.py#L29) class. There is a class [`CountAccumulator`](https://github.com/OpenMined/PipelineDP/blob/84bbf45740dcdc197c2c20a75b8eaaf6b1f2ecab/pipeline_dp/accumulator.py#L159) (DP part hasn't been yet implemented), which might be used as an example.

## API

```
@dataclass
class SumParams:
   noise:NoiseKind
   noise_std:float

class SumAccumulator(Accumulator): 
    def __init__(params:SumParams, values)
    def add_value(self, value)  # returns self
    def add_accumulator(self, accumulator: 'SumAccumulator')  # returns self
    def compute_metrics(self) -> float   
```

## Stages of the development

1. Implement sum computation.
2. Add anonymization to output, by using function from [dp_computations](https://github.com/OpenMined/PipelineDP/blob/main/pipeline_dp/dp_computations.py) (this is better to do in a separate PR) (more details next section)

## Adding anonymization
The idea is to use function [dp_sum](https://github.com/OpenMined/PipelineDP/blob/84bbf45740dcdc197c2c20a75b8eaaf6b1f2ecab/pipeline_dp/dp_computations.py#L198).

For dp_sum the following parameters are needed:
```
    eps: float
    delta: float
    max_partitions_contributed: int
    max_contributions_per_partition: int
    noise_kind: pipeline_dp.NoiseKind 
```

zachferr:
## Description
I added noise to the SumAccumulator. I'm not entirely sure I used the MeanVarParams correctly though.
This will close #61.

## Affected Dependencies
N/A

## How has this been tested?
- Unit tests.

## Checklist

### DIFF
--- a/pipeline_dp/accumulator.py
+++ b/pipeline_dp/accumulator.py
@@ -3,9 +3,12 @@ import typing
 import pickle
 from dataclasses import dataclass
 from functools import reduce
+
+from typing import Iterable, Tuple, Union
 import pipeline_dp
 from pipeline_dp import dp_computations
-
+from pipeline_dp import aggregate_params
+import numpy as np
 
 @dataclass
 class AccumulatorParams:
@@ -42,6 +45,13 @@ class Accumulator(abc.ABC):
     Returns: self.
     """
         pass
+
+    def _check_mergeable(self, accumulator: 'Accumulator'):
+        if not isinstance(accumulator, type(self)):
+            raise TypeError(
+                f"The accumulator to be added is not of the same type: "
+                f"{accumulator.__class__.__name__} != "
+                f"{self.__class__.__name__}")
 
     @abc.abstractmethod
     def add_accumulator(self, accumulator: 'Accumulator') -> 'Accumulator':
@@ -95,7 +105,7 @@ class CompoundAccumulator(Accumulator):
 
     The expectation is that the internal accumulators are of the same type and
     are in the same order."""
-
+        self._check_mergeable(accumulator)
         if len(accumulator.accumulators) != len(self.accumulators):
             raise ValueError(
                 "Accumulators in the input are not of the same size." +
@@ -166,6 +176,7 @@ class CountAccumulator(Accumulator):
 
     def add_accumulator(self,
                         accumulator: 'CountAccumulator') -> 'CountAccumulator':
+        self._check_mergeable(accumulator)
         self._count += accumulator._count
         return self
 
@@ -175,6 +186,49 @@ class CountAccumulator(Accumulator):
 
 
 @dataclass
+class VectorSummationParams:
+    noise_kind: aggregate_params.NoiseKind
+    max_norm: float
+
+_FloatVector = Union[Tuple[float], np.ndarray]
+
+class VectorSummationAccumulator(Accumulator):
+    _vec_sum: np.ndarray
+
+    def __init__(self, params: VectorSummationParams,
+                 values: Iterable[_FloatVector]) -> None:
+        self._params = params
+        self._vec_sum = None
+        for val in values:
+            self.add_value(val)
+
+    def add_value(self, value: _FloatVector):
+        if not isinstance(value, np.ndarray):
+            value = np.array(value)
+
+        if self._vec_sum is None:
+            self._vec_sum = value
+        else:
+            if self._vec_sum.shape != value.shape:
+                raise TypeError(
+                    f"Shape mismatch: {self._vec_sum.shape} != {value.shape}")
+            self._vec_sum += value
+        return self
+
+    def add_accumulator(
+        self, accumulator: 'VectorSummationAccumulator'
+    ) -> 'VectorSummationAccumulator':
+        self._check_mergeable(accumulator)
+        self.add_value(accumulator._vec_sum)
+        return self
+
+    def compute_metrics(self):
+        # TODO - add DP anonymization
+        if self._vec_sum is None:
+            raise IndexError("No data provided for metrics computation.")
+        return self._vec_sum
+
+
 class SumParams:
     noise: dp_computations.MeanVarParams
 
@@ -190,6 +244,7 @@ class SumAccumulator(Accumulator):
 
     def add_accumulator(self,
                         accumulator: 'SumAccumulator') -> 'SumAccumulator':
+        self._check_mergeable(accumulator)
         self._sum += accumulator._sum
 
     def compute_metrics(self) -> float:
--- a/pipeline_dp/accumulator.py
+++ b/pipeline_dp/accumulator.py
@@ -3,7 +3,6 @@ import typing
 import pickle
 from dataclasses import dataclass
 from functools import reduce
-from dataclasses import dataclass
 import pipeline_dp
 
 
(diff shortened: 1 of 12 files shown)


pipeline_dp/accumulator.py
```diff
@@ -4,6 +4,7 @@
 from dataclasses import dataclass
 from functools import reduce
 import pipeline_dp
+from pipeline_dp.dp_computations import MeanVarParams
```
dvadym, line 7: By [styleguide](https://google.github.io/styleguide/pyguide.html#22-imports), only modules should be imported (with some standard classes exceptions).

tests/accumulator_test.py
```diff
@@ -263,17 +264,41 @@ class SumAccumulatorTest(unittest.TestCase):
...
+                              pipeline_dp.NoiseKind.GAUSSIAN)), list(range(6)))
         self.assertEqual(sum_accumulator.compute_metrics(), 15)
 
         sum_accumulator.add_value(5)
         self.assertEqual(sum_accumulator.compute_metrics(), 20)
 
         sum_accumulator_2 = accumulator.SumAccumulator(
-            accumulator.SumParams(), list(range(3)))
+            accumulator.SumParams(
+                MeanVarParams(1, 1, 0, 0, 1, 1,
```
dvadym, line 277: This test (w/o noise) makes sense. It's possible to achieve that the result is almost w/o noise, by setting very high (eps, delta) - like eps = 1e20, delta = 1 (delta might be considered as probability, it's in range [0, 1])

tests/accumulator_test.py
```diff
@@ -263,17 +264,41 @@ class SumAccumulatorTest(unittest.TestCase):
...
+                              pipeline_dp.NoiseKind.GAUSSIAN)), list(range(3)))
         sum_accumulator.add_accumulator(sum_accumulator_2)
         self.assertEqual(sum_accumulator.compute_metrics(), 23)
 
+    def test_with_noise(self):
+        sum_accumulator = accumulator.SumAccumulator(
+            accumulator.SumParams(
+                MeanVarParams(2, 1, 1, 1, 1, 3,
```
dvadym, line 285: you can set higher eps (like 100), then you can use smaller, delta in `assertAlmostEqual`

tests/accumulator_test.py
```diff
@@ -263,17 +264,41 @@ class SumAccumulatorTest(unittest.TestCase):
...
+                              pipeline_dp.NoiseKind.GAUSSIAN)), list(range(6)))
+        self.assertAlmostEqual(sum_accumulator.compute_metrics(), 15, None,
+                               None, 3)
+
+        sum_accumulator.add_value(5)
+        self.assertAlmostEqual(sum_accumulator.compute_metrics(), 20, None,
```
dvadym, line 291: Nit: please use keyword arguments, then no need to set arguments with default values, the same in other calls `assertAlmostEqual`

dvadym (commented):
Thanks Zachery, it looks good! I left some comments

tests/accumulator_test.py
```diff
@@ -263,17 +264,41 @@ class SumAccumulatorTest(unittest.TestCase):
 
     def test_without_noise(self):
         sum_accumulator = accumulator.SumAccumulator(
-            accumulator.SumParams(), list(range(6)))
+            accumulator.SumParams(
+                MeanVarParams(1, 1, 0, 0, 1, 1,
```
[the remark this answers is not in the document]
zachferr: Done.

tests/accumulator_test.py
```diff
@@ -263,17 +264,41 @@ class SumAccumulatorTest(unittest.TestCase):
...
+                              pipeline_dp.NoiseKind.GAUSSIAN)), list(range(6)))
```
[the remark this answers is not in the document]
zachferr: Done

pipeline_dp/accumulator.py
zachferr: Done

tests/accumulator_test.py
zachferr: Done, thanks for the tip

tests/accumulator_test.py
zachferr: Done, but since the tests were so much closer in value I also switched to using places instead of delta.

tests/accumulator_test.py
```diff
@@ -262,18 +263,60 @@ def test_without_noise(self):
 class SumAccumulatorTest(unittest.TestCase):
 
     def test_without_noise(self):
+        no_noise = MeanVarParams(eps=1,
+                                 delta=1,
+                                 low=0,
```
dvadym, line 268: Nice approach to make noise 0, to set low and high equal!

tests/accumulator_test.py
```diff
@@ -262,18 +263,60 @@ def test_without_noise(self):
...
-            accumulator.SumParams(), list(range(3)))
+            accumulator.SumParams(no_noise), list(range(3)))
         sum_accumulator.add_accumulator(sum_accumulator_2)
         self.assertEqual(sum_accumulator.compute_metrics(), 23)
 
+    def test_with_noise(self):
+        sum_accumulator = accumulator.SumAccumulator(
+            accumulator.SumParams(
+                MeanVarParams(eps=100,
+                              delta=1,
```
dvadym, line 289: Let's use the following parameters
eps = 10
delta = 1e-5
low = 0
high = 1
and other leave as is

Explanation: the standard deviation of noise which is added is computed by the formula
abs(low-high)*f(eps, delta,...)  , higher eps, delta, lower noise. 
so setting low=high makes noise to be **0**, 

Delta is probability, which is in production should be small (it might be treated as a probability of non-anymizing)
Setting delta = 1, also makes noise to be **0** (maybe not now, because some optimal formulas are not implemented) so it's better to use some more realistic values

The noise should be less than 1

tests/accumulator_test.py
```diff
@@ -262,18 +263,60 @@ def test_without_noise(self):
...
+                              noise_kind=pipeline_dp.NoiseKind.GAUSSIAN)),
+            list(range(6)))
+        self.assertAlmostEqual(first=sum_accumulator.compute_metrics(),
+                               second=15,
+                               places=1)
+
+        sum_accumulator.add_value(5)
+        self.assertAlmostEqual(first=sum_accumulator.compute_metrics(),
+                               second=20,
+                               places=1)
```
dvadym, line 303: Please use delta=1.0 or delta = 0.5, instead of places = 1. Here and in other places

dvadym (commented):
Thanks a few minor comments

tests/accumulator_test.py
zachferr: After making the other changes I had to make the delta 4 to have the test pass consistently. Is that okay or should we do more parameter tweaking?

tests/accumulator_test.py
zachferr: Done. Thanks for the explanation!

dvadym (approved):
Thanks for implementing of SumAccumulator!

PR #76, opened 2021-09-13
Fix for spark map_tuple

dvadym:
map_tuples was implemented as map in SparkRDDPipelineOperation. This PR fixes that and adds tests.

Along the way, the creation of SparkRDDOperations() was moved to setUpClass(cls), to match BeamOperations unittests.

### DIFF
--- a/tests/pipeline_operations_test.py
+++ b/tests/pipeline_operations_test.py
@@ -156,12 +156,11 @@ class SparkRDDOperationsTest(parameterized.TestCase):
         self.assertDictEqual(result, {'a': 2, 'b': 1})
 
     def test_reduce_accumulators_per_key(self):
-        spark_operations = SparkRDDOperations()
         data = [(1, 11), (2, 22), (3, 33), (1, 14), (2, 25), (1, 16)]
         dist_data = SparkRDDOperationsTest.sc.parallelize(data)
-        rdd = spark_operations.map_values(dist_data, SumAccumulator,
+        rdd = self.ops.map_values(dist_data, SumAccumulator,
                                           "Wrap into accumulators")
-        result = spark_operations\
+        result = self.ops\
             .reduce_accumulators_per_key(rdd, "Reduce accumulator per key")\
             .map(lambda row: (row[0], row[1].get_metrics()))\
             .collect()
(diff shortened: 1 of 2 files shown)


boyaryn:
I guess we expect `SparkRDDOperations` to be stateless. What do you think about declaring its methods class methods?

tests/pipeline_operations_test.py
```diff
@@ -140,21 +138,19 @@ def test_filter_by_key_nonempty_public_partitions(self, distributed):
...
         self.assertTrue(set(result[1]).issubset({11, 14, 16}))
         self.assertSetEqual(set(result[2]), {22, 25})
         self.assertSetEqual(set(result[3]), {33})
 
     def test_count_per_element(self):
-        spark_operations = SparkRDDOperations()
         data = ['a', 'b', 'a']
         dist_data = SparkRDDOperationsTest.sc.parallelize(data)
-        rdd = spark_operations.count_per_element(dist_data)
+        rdd = self.ops.count_per_element(dist_data)
```
boyaryn, line 153: Could you also `replace spark_operations` with `self.ops` in the following method `test_reduce_accumulators_per_key`?
I couldn't figure out how to make a comment to a line which wasn't added by you. :-)

boyaryn (commented):
Thanks, the PR is good, only one small thing was overlooked.

tests/pipeline_operations_test.py
dvadym: Thanks, fixed. I don't know a way to comment on not-changed lines either :)

boyaryn (approved):

dvadym:
> I guess we expect `SparkRDDOperations` to be stateless. What do you think about declaring its methods class methods?

Now `PipelineOperations` is stateless, but it's not necessary to be so, e.g. we might want to manage `stage_name` in `BeamOperations` (managing stage_name is pretty complicate in Beam, sometimes to avoid duplicate names (which are not allowed) scopes are introduced. So, I think it's better to keep methods `PipelineOperations` to instance methods, anyway it doesn't add any overhead (nor programming/nor performance)

dvadym:
Thanks for the review!

PR #77, opened 2021-09-13
Improvement of the movie view example

dvadym:
This PR contains the following changes:

1.Using LocalOperations in the example
2.Minor fixes of API usage (e.g. calling budget_account.compute_budgets() etc).
3.Limiting metrics only to COUNT (all others are not implemented yet)
4.Minor fix in dp_engine.aggregate (map_tuple -> map)  (it's not tested yet, since dp_engine.aggregate is not fully implemented), w/o this fix the example crashes instead of returning an intermediate result.
5.Making some flags to be required
6.Added examples folders for formating by `make format`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 
 .PHONY: format
 format:
-	yapf -i -r pipeline_dp tests
+	yapf -i -r pipeline_dp tests examples
 
 .PHONY: lint
 lint:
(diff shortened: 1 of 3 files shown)


jspacek:
Code changes LGTM 👍 @dvadym 

Nit: when I run this locally, it prints out `W0915 11:59:21.926725 8664768000 budget_accounting.py:94] No budgets were requested.` at the end. Could the log level be changed to debug or info, in case this would be confusing as a first example?

jspacek (approved):

dvadym:
Thanks for the review!

> Code changes LGTM 👍 @dvadym
> 
> Nit: when I run this locally, it prints out `W0915 11:59:21.926725 8664768000 budget_accounting.py:94] No budgets were requested.` at the end. Could the log level be changed to debug or info, in case this would be confusing as a first example?

This warning is shown because budget is not requested yet. It will be fixed in one of the following PR.

PR #86, opened 2021-09-23
Making aggregate with private partition selection work

dvadym:
This PR finishes implementation of the computation graph (which is implemented by `DPEngine.aggregate` ) for computing DP metrics with private partition selection.

### DIFF
--- a/examples/movie_view_ratings.py
+++ b/examples/movie_view_ratings.py
@@ -52,15 +52,14 @@ def calc_dp_rating_metrics(movie_views, ops, public_partitions):
     dp_engine = pipeline_dp.DPEngine(budget_accountant, ops)
 
     # Specify which DP aggregated metrics to compute.
-    params = pipeline_dp.AggregateParams(
-        metrics=[
-            pipeline_dp.Metrics.COUNT,
-        ],
-        max_partitions_contributed=2,
-        max_contributions_per_partition=1,
-        low=1,
-        high=5,
-        public_partitions=public_partitions)
+    params = pipeline_dp.AggregateParams(metrics=[
+        pipeline_dp.Metrics.COUNT,
+    ],
+                                         max_partitions_contributed=2,
+                                         max_contributions_per_partition=1,
+                                         low=1,
+                                         high=5,
+                                         public_partitions=public_partitions)
 
     # Specify how to extract is privacy_id, partition_key and value from an element of movie view collection.
     data_extractors = pipeline_dp.DataExtractors(
--- a/examples/movie_view_ratings.py
+++ b/examples/movie_view_ratings.py
@@ -61,7 +61,8 @@ def calc_dp_rating_metrics(movie_views, ops, public_partitions):
                                          high=5,
                                          public_partitions=public_partitions)
 
-    # Specify how to extract is privacy_id, partition_key and value from an element of movie view collection.
+    # Specify how to extract is privacy_id, partition_key and value from an
+    # element of movie view collection.
     data_extractors = pipeline_dp.DataExtractors(
         partition_extractor=lambda mv: mv.movie_id,
         privacy_id_extractor=lambda mv: mv.user_id,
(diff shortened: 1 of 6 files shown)


tests/dp_engine_test.py
```diff
@@ -302,6 +302,73 @@ def test_select_private_partitions(self):
...
+                                                        1)
+        # Set a large budget for having the small noise and keeping all
+        # partition keys.
+        budget_accountant = NaiveBudgetAccountant(total_epsilon=1000,
+                                                  total_delta=1e-10)
+
+        col = list(range(10)) + list(range(100, 120))
+        data_extractor = pipeline_dp.DataExtractors(
+            privacy_id_extractor=lambda x: x,
+            partition_extractor=lambda x: "pk" + str(x // 100),
```
levzlotnik, line 317: Nit ([Google Style Guide](https://google.github.io/styleguide/pyguide.html#s3.10-strings)): use format or f-string:
```python
partition_extractor=lambda x: f"pk{x // 100}"
```

tests/dp_engine_test.py
```diff
@@ -302,6 +302,73 @@ def test_select_private_partitions(self):
...
+        # partition keys.
+        budget_accountant = NaiveBudgetAccountant(total_epsilon=1,
+                                                  total_delta=1e-10)
+
+        # Input collection has 100 elements, such that each privacy id
+        # contributes 1 time and each partition has 1 element.
+        col = list(range(100))
+        data_extractor = pipeline_dp.DataExtractors(
+            privacy_id_extractor=lambda x: x,
+            partition_extractor=lambda x: "pk" + str(x),
```
levzlotnik, line 352: Nit ([Google Style Guide](https://google.github.io/styleguide/pyguide.html#s3.10-strings)): Use format or f-strings
```python
partition_extractor=lambda x: f"pk{x}"
```

pipeline_dp/__init__.py
```diff
@@ -1,4 +1,4 @@
-from pipeline_dp.budget_accounting import BudgetAccountant
```
levzlotnik, line 1: Nit: Are we sure we want to reference the base class only through the `pipeline_dp.budget_accounting`?  
If someone creates a new `BudgetAccountant`, they gotta import `budget_accounting`, whereas with `{Naive,PLD}BudgetAccountant` they can just do:
```python
import pipeline_dp as pldp
from pipeline_dp import budget_accountant


def do_something_special(accountant: budget_accounting.BudgetAccountant):
    # Also note how we repeat ourselves with "budget_accounting.BudgetAccountant"...
    pass

accountant = pldp.PLDBudgetAccountant(params)
do_something_special(accountant)
```

levzlotnik (changes requested):
Minor code changes. Regarding testing - Windows & MacOSX are facing CI env issues, unrelated to the code base

tests/dp_engine_test.py
dvadym: Done, thanks!

tests/dp_engine_test.py
dvadym: Done, thanks!

examples/movie_view_ratings.py
```diff
@@ -45,23 +45,24 @@ def calc_dp_rating_metrics(movie_views, ops, public_partitions):
...
-
-    # Specify how to extract is privacy_id, partition_key and value from an element of movie view collection.
+    params = pipeline_dp.AggregateParams(metrics=[
+        pipeline_dp.Metrics.COUNT,
+    ],
+                                         max_partitions_contributed=2,
+                                         max_contributions_per_partition=1,
+                                         low=1,
+                                         high=5,
+                                         public_partitions=public_partitions)
```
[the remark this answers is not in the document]
dvadym: Yes, it was an output of the formatter, agree it's weird. I've reformatted it.

pipeline_dp/__init__.py
dvadym: Thanks, a good point! A base abstract class of a public API class is also a part of the public API.

 I've returned back import of `BudgetAccountant`.

dvadym (commented):
Thanks for the review! I've addressed comments. PTAL

PR #87, opened 2021-09-28
Add SumParams to movie example and tests

jspacek:
## Description
Add functionality for SumAccumulator through SumParams.

## Affected Dependencies
None.

## How has this been tested?
- `GenericAccumulatorTest.test_create_accumulator_params_with_sum_params`
- `SumAccumulatorTest.test_without_noise`
- `SumAccumulatorTest.test_with_noise`
- `movie_view_ratings.calc_dp_rating_metrics`

## Checklist

### DIFF
--- a/examples/movie_view_ratings.py
+++ b/examples/movie_view_ratings.py
@@ -52,24 +52,28 @@ def calc_dp_rating_metrics(movie_views, ops, public_partitions):
     dp_engine = pipeline_dp.DPEngine(budget_accountant, ops)
 
     # Specify which DP aggregated metrics to compute.
-    params = pipeline_dp.AggregateParams(metrics=[pipeline_dp.Metrics.COUNT],
-                                         max_partitions_contributed=2,
-                                         max_contributions_per_partition=1,
-                                         low=1,
-                                         high=5,
-                                         public_partitions=public_partitions)
+    params = pipeline_dp.AggregateParams(
+        noise_kind=pipeline_dp.aggregate_params.NoiseKind.LAPLACE,
+        metrics=[pipeline_dp.Metrics.COUNT, pipeline_dp.Metrics.SUM],
+        max_partitions_contributed=2,
+        max_contributions_per_partition=1,
+        low=1,
+        high=5,
+        public_partitions=public_partitions)
 
-    # Specify how to extract is privacy_id, partition_key and value from an
+    # Specify how to extract privacy_id, partition_key and value from an
     # element of movie view collection.
     data_extractors = pipeline_dp.DataExtractors(
         partition_extractor=lambda mv: mv.movie_id,
         privacy_id_extractor=lambda mv: mv.user_id,
         value_extractor=lambda mv: mv.rating)
-
+    # TODO Not sure if we want the client to have to make these calls to budget accountant
+    # but without these there will be epsilon, delta, and empty mechanisms errors
+    budget_accountant.request_budget(pipeline_dp.MechanismType.LAPLACE)
+    budget_accountant.compute_budgets()
     # Run aggregation.
     dp_result = dp_engine.aggregate(movie_views, params, data_extractors)
 
-    budget_accountant.compute_budgets()
     return dp_result
 
 
--- a/examples/movie_view_ratings.py
+++ b/examples/movie_view_ratings.py
@@ -53,7 +53,7 @@ def calc_dp_rating_metrics(movie_views, ops, public_partitions):
 
     # Specify which DP aggregated metrics to compute.
     params = pipeline_dp.AggregateParams(
-        noise_kind=pipeline_dp.aggregate_params.NoiseKind.LAPLACE,
+        noise_kind=pipeline_dp.NoiseKind.LAPLACE,
         metrics=[pipeline_dp.Metrics.COUNT, pipeline_dp.Metrics.SUM],
         max_partitions_contributed=2,
         max_contributions_per_partition=1,
@@ -67,13 +67,11 @@ def calc_dp_rating_metrics(movie_views, ops, public_partitions):
         partition_extractor=lambda mv: mv.movie_id,
         privacy_id_extractor=lambda mv: mv.user_id,
         value_extractor=lambda mv: mv.rating)
-    # TODO Not sure if we want the client to have to make these calls to budget accountant
-    # but without these there will be epsilon, delta, and empty mechanisms errors
-    budget_accountant.request_budget(pipeline_dp.MechanismType.LAPLACE)
-    budget_accountant.compute_budgets()
+
     # Run aggregation.
     dp_result = dp_engine.aggregate(movie_views, params, data_extractors)
 
+    budget_accountant.compute_budgets()
     return dp_result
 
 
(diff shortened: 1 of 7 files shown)


examples/movie_view_ratings.py
```diff
@@ -52,24 +52,29 @@ def calc_dp_rating_metrics(movie_views, ops, public_partitions):
...
-    params = pipeline_dp.AggregateParams(metrics=[pipeline_dp.Metrics.COUNT],
-                                         max_partitions_contributed=2,
-                                         max_contributions_per_partition=1,
-                                         low=1,
-                                         high=5,
-                                         public_partitions=public_partitions)
-
-    # Specify how to extract is privacy_id, partition_key and value from an
+    params = pipeline_dp.AggregateParams(
+        noise_kind=pipeline_dp.aggregate_params.NoiseKind.LAPLACE,
```
dvadym, line 56: Nit: pipeline_dp.NoiseKind.LAPLACE  (it's already in __init__.py)

examples/movie_view_ratings.py
```diff
@@ -52,24 +52,29 @@ def calc_dp_rating_metrics(movie_views, ops, public_partitions):
...
     data_extractors = pipeline_dp.DataExtractors(
         partition_extractor=lambda mv: mv.movie_id,
         privacy_id_extractor=lambda mv: mv.user_id,
         value_extractor=lambda mv: mv.rating)
-
+    # TODO Not sure if we want the client to have to make these calls to budget accountant
+    # but without these there will be epsilon, delta, and empty mechanisms errors
+    budget_accountant.request_budget(pipeline_dp.MechanismType.LAPLACE)
+    budget_accountant.request_budget(pipeline_dp.MechanismType.LAPLACE)
+    budget_accountant.compute_budgets()
```
dvadym, line 74: `compute_budgets` should be after the last call of `.aggregate()` (the idea is that budgets are computed after the DP computation pipeline is constructed, which happen in calls of `aggregate()`.

Then no need to have calls to reques_budget in this file

pipeline_dp/accumulator.py
```diff
@@ -257,9 +264,34 @@ def compute_metrics(self):
...
+    """Parameters for a sum accumulator.
+
+    Wraps epsilon and delta from the budget which are lazily loaded.
+    AggregateParams are copied into a MeanVarParams instance.
+    """
+
+    def __init__(self, budget: pipeline_dp.budget_accounting.MechanismSpec,
+                 aggregate_params: aggregate_params.AggregateParams):
+        self._budget = budget
+        self._aggregate_params = aggregate_params
```
dvadym, line 277: Thanks, I like this approach more that with inheritance!

pipeline_dp/dp_engine.py
```diff
@@ -165,6 +165,7 @@ def _select_private_partitions(self, col, max_partitions_contributed: int):
         """
         budget = self._budget_accountant.request_budget(
             mechanism_type=MechanismType.GENERIC)
+        self._budget_accountant.compute_budgets()
```
dvadym, line 168: `compute_budgets()` are needed only after finishing constructing of the DP pipeline, namely after the last call of `.aggregate()`, please remove this line

dvadym (commented):
Thanks, it looks good. I left some comments

pipeline_dp/accumulator.py
```diff
@@ -257,9 +264,34 @@ def compute_metrics(self):
...
+
+    @property
+    def delta(self):
+        return self._budget.delta
+
+    @property
+    def mean_var_params(self):
+        return dp_computations.MeanVarParams(
+            self.eps, self.delta, self._aggregate_params.low,
+            self._aggregate_params.high,
```
jspacek, line 291: @dvadym 👋 

Not sure if this is exactly right; it doesn't inherit MeanVarParams as a superclass because instantiating it in the `__init__` fails due to the dependency on budget accountant `eps must not be None.` So I created `mean_var_params` as a property so it can be called after compute budgets. 

Also, maybe we want to copy the `aggregate_params` attributes entirely so we don't rely on the other object's state? Is it OK "pythonic" to carry state in this way? 🐍

examples/movie_view_ratings.py
jspacek: 👍 Got it, fixed!

pipeline_dp/dp_engine.py
jspacek: 👍 Fixed, thank you!

examples/movie_view_ratings.py
jspacek: 👍 Thank you!

pipeline_dp/accumulator.py
jspacek: Awesome, it seems to work well. Thanks!

pipeline_dp/accumulator.py
```diff
@@ -257,9 +264,34 @@ def compute_metrics(self):
         return dp_computations.add_noise_vector(self._vec_sum, self._params)
 
 
-@dataclass
 class SumParams:
-    noise: dp_computations.MeanVarParams
+    """Parameters for a sum accumulator.
```
dvadym, line 268: Nit: SumAccumulator

pipeline_dp/accumulator.py
```diff
@@ -33,6 +33,13 @@ def create_accumulator_params(
         accumulator_params.append(
             AccumulatorParams(accumulator_type=CountAccumulator,
                               constructor_params=CountParams()))
+    if pipeline_dp.Metrics.SUM in aggregation_params.metrics:
```
dvadym, line 36: could you please add testing of Sum parameters [here](https://github.com/OpenMined/PipelineDP/blob/4049178721a5ef9e013dfce398356b9d3388dfad/tests/accumulator_test.py#L252)

pipeline_dp/accumulator.py
dvadym: do you mean to have a field in `SumParams` of type `MeanVarParams` ? If yes, it would make sense, but in `__init__` we don't have yet `eps/delta` to construct  `MeanVarParams`.

My feeling is that the current approach is reasonable. And any anyway this code will be refactored when we will introduce using of PldBudgetAccountant.

pipeline_dp/aggregate_params.py
```diff
@@ -36,6 +42,7 @@ class AggregateParams:
     """Specifies parameters for function DPEngine.aggregate()
 
   Args:
+    noise_kind: Kind of noise to use for the calculation.
```
dvadym, line 45: Nit: the **DP** calculation**s**.

pipeline_dp/accumulator.py
jspacek: Yes, to copy directly everything except eps/delta which would be more of a stateless/functional style that is not necessary here :) Thanks for clarifying!

pipeline_dp/accumulator.py
jspacek: 👍 Fixed!

pipeline_dp/accumulator.py
jspacek: Done, thanks! 👍

pipeline_dp/aggregate_params.py
jspacek: 👍 Fixed!

dvadym (approved):
Thanks a lot of making sum DP computations work end-to-end!

PR #90, opened 2021-09-30
Fix running on Spark 

dvadym:


pipeline_dp/dp_engine.py
```diff
@@ -176,11 +178,35 @@ def filter_fn(captures: Tuple[MechanismSpec, int],
             partitions to keep."""
             mechanism, max_partitions = captures
             accumulator = row[1]
-            partition_selection_strategy = create_truncated_geometric_partition_strategy(
+            partition_selection_strategy = partition_selection.create_truncated_geometric_partition_strategy(
```
dvadym, line 181: This fixes PyCapsule serialization error

pipeline_dp/dp_engine.py
```diff
@@ -90,6 +90,8 @@ def aggregate(self, col, params: AggregateParams,
             pass
         # col : (partition_key, accumulator)
 
+        col = self._fix_budget_accounting_for_spark(col, accumulator_factory)
```
boyaryn, line 93: It's great that you've solved the serialization problem for Spark.
My only concern is this line seems to violate some object-oriented design principle, probably D from SOLID. Should we redesign this? Or should it be left as is now and hopefully refactored later (I see this implementation is still in development).

boyaryn (commented):
In general, looks great!

pipeline_dp/dp_engine.py
dvadym: yeah, it makes sense. I agree that code here shouldn't know that it runs on Spark.

On other hand, PipelineOperations class knows nothing about accumulators and budgets, it know only about framework details. So updates of the accumulators should be initiated on higher level, and `DPEngine` looks the right place.

I've renamed this function to `_fix_budget_accounting_if_needed` and renamed corresponding PipelineOperation function to `is_serialization_immediate_on_reduce_by_key`. Such way we avoid dependency on knowing the framework.

WDYT?

boyaryn: IMHO the new version of the code looks much better with this application of the strategy pattern.

Thanks, approved.

boyaryn (approved):

pipeline_dp/dp_engine.py
dvadym: Thanks for review!

PR #105, opened 2021-09-30
Fix filtering of public partitions

dvadym:
This PR changes elements of output collection of `DPEngine._drop_not_public_partitions` in order to make it consistent with the rest of `DPEngine.aggregate`, namely `_drop_not_public_partitions` now returns collection of data instead of (key, data).

Along the way

`PipelineOperation.filter_by_key` was slightly refactored
1. Remove any mentions of public partition in `filter_by_key`, since this function is more generic.
2. Remove data_extractors from `filter_by_key` arguments (since they are used in `_drop_not_public_partitions`)

### DIFF
--- a/pipeline_dp/dp_engine.py
+++ b/pipeline_dp/dp_engine.py
@@ -97,7 +97,8 @@ class DPEngine:
         return col
 
     def _drop_not_public_partitions(self, col, public_partitions,
-                                    data_extractors):
+                                    data_extractors: DataExtractors):
+        """"""
         col = self._ops.map(
             col, lambda row: (data_extractors.partition_extractor(row), row),
             "Extract partition id")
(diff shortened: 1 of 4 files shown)


pipeline_dp/dp_engine.py
```diff
@@ -98,28 +97,33 @@ def aggregate(self, col, params: AggregateParams,
...
+            col, lambda row: (data_extractors.partition_extractor(row), row),
+            "Extract partition id")
+        col = self._ops.filter_by_key(col, public_partitions,
+                                      "Filtering out non-public partitions")
+        return self._ops.map_tuple(col, lambda k, v: v, "Drop key")
 
     def _bound_contributions(self, col, max_partitions_contributed: int,
                              max_contributions_per_partition: int,
                              aggregator_fn):
+        """Bounds the contribution by privacy_id in and cross partitions.
```
dvadym, line 112: This is only reformatting

pipeline_dp/dp_engine.py
```diff
@@ -98,28 +97,33 @@ def aggregate(self, col, params: AggregateParams,
         return col
 
     def _drop_not_public_partitions(self, col, public_partitions,
-                                    data_extractors):
-        return self._ops.filter_by_key(col, public_partitions, data_extractors,
-                                       "Filtering out non-public partitions")
+                                    data_extractors: DataExtractors):
+        """"""
```
jspacek, line 101: Nit: Remove empty comment line

pipeline_dp/pipeline_operations.py
```diff
@@ -124,52 +123,45 @@ def group_by_key(self, col, stage_name: str):
     def filter(self, col, fn, stage_name: str):
         return col | stage_name >> beam.Filter(fn)
 
-    def filter_by_key(self, col, keys_to_keep, data_extractors,
-                      stage_name: str):
+    def filter_by_key(self, col, keys_to_keep, stage_name: str):
 
         class PartitionsFilterJoin(beam.DoFn):
```
jspacek, line 128: Curious, my IDE complains that this method isn't implemented from the abstract super class, not sure if it's needed?

```
           def to_runner_api_parameter(self, unused_context):
                pass
```

jspacek:
Should we test what the behaviour is if a collection has duplicate keys, or is that not even possible? 

LGTM 👍

jspacek (approved):

pipeline_dp/pipeline_operations.py
dvadym: It's probably some error of parsing by IDE. This method is not abstract. 

And according to [example](https://beam.apache.org/documentation/transforms/python/elementwise/pardo/), it's needed only to implement `process()`

pipeline_dp/dp_engine.py
dvadym: Thanks, I wanted to add a comment and forgot to do it. I've added a comment

dvadym (commented):
Thanks for the review!

It's a good point to test when public_partitions have duplicating keys. I've added it to unit tests

PR #109, opened 2021-10-06
Implement clamping (clipping) for Sum

miracvbasaran:
## Description
This implements per-contribution clamping for SumAccumulator.

## Affected Dependencies
None

## How has this been tested?
- Unit tests
- E2E movie ratings script
- 
## Checklist
- [+] I have followed the [Contribution Guidelines](https://github.com/OpenMined/.github/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://github.com/OpenMined/.github/blob/master/CODE_OF_CONDUCT.md)
- [+] I have commented my code following the [OpenMined Styleguide](https://github.com/OpenMined/.github/blob/master/STYLEGUIDE.md)
- [+] I have labeled this PR with the relevant [Type labels](https://github.com/OpenMined/.github/labels?q=Type%3A)
- [+] My changes are covered by tests

### DIFF
--- a/pipeline_dp/accumulator.py
+++ b/pipeline_dp/accumulator.py
@@ -227,7 +227,7 @@ class AccumulatorClassParams:
     def __init__(self, spec: budget_accounting.MechanismSpec,
                  aggregate_params: aggregate_params.AggregateParams):
         self._mechanism_spec = spec
-        self._aggregate_params = copy.copy(aggregate_params)
+        self.aggregate_params = copy.copy(aggregate_params)
 
     @property
     def eps(self):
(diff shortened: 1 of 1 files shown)


pipeline_dp/accumulator.py
```diff
@@ -366,3 +368,10 @@ def add_accumulator(self,
     def compute_metrics(self) -> float:
         return dp_computations.compute_dp_sum(self._sum,
                                               self._params.mean_var_params)
+
+    def clamp_value(self, value):
```
dvadym, line 372: Nit: use [np.clip](https://numpy.org/doc/stable/reference/generated/numpy.clip.html). It should work for the list as well

pipeline_dp/accumulator.py
```diff
@@ -352,11 +352,15 @@ class SumParams(AccumulatorClassParams):
 class SumAccumulator(Accumulator):
 
     def __init__(self, params: SumParams, values):
-        self._sum = sum(values)
         self._params = params
+        self._sum = 0
+        for value in values:
```
dvadym, line 357: Nit: we can do it in 1 line (it's Python 😄 ) 
self._sum += np.clip(values, ...).sum()

and it will be more effective, since numpy vectorization is usually faster than loops

pipeline_dp/accumulator.py
```diff
@@ -352,11 +352,13 @@ class SumParams(AccumulatorClassParams):
 class SumAccumulator(Accumulator):
 
     def __init__(self, params: SumParams, values):
-        self._sum = sum(values)
         self._params = params
+        self._sum = np.clip(values, self._params.mean_var_params.low,
```
dvadym, line 356: A test fails because of access to `mean_var_params` at this moment, namely creating `mean_var_params` requires access `eps/delta` which are not available at that moment on Spark.

Could you please make [_aggregate_params](https://github.com/OpenMined/PipelineDP/blob/9eecae87e9ee34343ace737fa48fdc7c69dbd870/pipeline_dp/accumulator.py#L230) public (i.e. to remove _) and use `aggregate_params.low/high`?

dvadym (commented):
Thanks for implementing clipping!

PR #113, opened 2021-10-18
Debugging by removing multiple budget accountants in accumulator

jspacek:
## Description
This appears to fix the large noise variation in the demo: [Colab that pulls from OpenMined:main](https://colab.sandbox.google.com/drive/1paOyAew60fOe-v7ZSr0To-ZmUz5MzhWe#scrollTo=k_Nu7EO3jPI1) 

As shown in : [Colab that pulls from this branch jspacek:debug-demo-sum](https://colab.sandbox.google.com/drive/13dxBoNkvD9WeMMkPasatnfwDGuqHMP5t#scrollTo=k_Nu7EO3jPI1) there are no negative results after running the demo multiple times.

I'm not sure this should be the fix; perhaps another bit of code further down the line needs to divide the noise by the number of mechanisms in the budget accountant?

## How has this been tested?
TODO likely a new test in dp_engine_test e2e

## Checklist
- [ ] I have followed the [Contribution Guidelines](https://github.com/OpenMined/.github/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://github.com/OpenMined/.github/blob/master/CODE_OF_CONDUCT.md)
- [ ] I have commented my code following the [OpenMined Styleguide](https://github.com/OpenMined/.github/blob/master/STYLEGUIDE.md)

### DIFF
--- a/pipeline_dp/accumulator.py
+++ b/pipeline_dp/accumulator.py
@@ -30,22 +30,20 @@ def create_accumulator_params(
 ) -> typing.List[AccumulatorParams]:
     accumulator_params = []
     mechanism_type = aggregation_params.noise_kind.convert_to_mechanism_type()
+    budget_accountant = budget_accountant.request_budget(
+        mechanism_type)
     if pipeline_dp.Metrics.COUNT in aggregation_params.metrics:
-        budget_count = budget_accountant.request_budget(mechanism_type)
-        count_params = CountParams(budget_count, aggregation_params)
+        count_params = CountParams(budget_accountant, aggregation_params)
         accumulator_params.append(
             AccumulatorParams(accumulator_type=CountAccumulator,
                               constructor_params=count_params))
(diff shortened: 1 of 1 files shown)


pipeline_dp/accumulator.py
```diff
@@ -30,22 +30,20 @@ def create_accumulator_params(
 ) -> typing.List[AccumulatorParams]:
     accumulator_params = []
     mechanism_type = aggregation_params.noise_kind.convert_to_mechanism_type()
+    budget_accountant = budget_accountant.request_budget(
```
dvadym, line 33: The current implementation is correct, we need a call to `request_budget` for each operations, it's how the API of `BudgetAccountant` works.

This PR makes that only 1 request is done as a result each metric get the full budget (i.e. budget exceeding), and so there is less noise.

PR #119, opened 2021-11-12
Beam PrivateTTransform Map And FlatMap

preethiraghavan1:
## Description
PrivateTTransform Map And FlatMap

## How has this been tested?
-Unit tests

## Checklist
- [ x] I have followed the [Contribution Guidelines](https://github.com/OpenMined/.github/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://github.com/OpenMined/.github/blob/master/CODE_OF_CONDUCT.md)
- [ x] I have commented my code following the [OpenMined Styleguide](https://github.com/OpenMined/.github/blob/master/STYLEGUIDE.md)
- [x ] I have labeled this PR with the relevant [Type labels](https://github.com/OpenMined/.github/labels?q=Type%3A)
- [x ] My changes are covered by tests


pipeline_dp/private_beam.py
```diff
@@ -104,3 +106,36 @@ def expand(self, pcol: pvalue.PCollection) -> pvalue.PCollection:
...
+class FlatMap(PrivatePTransform):
+    """Transform class for performing FlatMap on PrivatePCollection."""
+
+    class _FlattenValues(beam.DoFn):
+        """Inner class for flattening values of key value pair.
+        Flattens (1, (2,3,4)) into ((1,2), (1,3), (1,4))"""
+
+        def process(self, row):
+            key = row[0]
+            values = row[1] if isinstance(row[1], Iterable) else [row[1]]
```
dvadym, line 131: It looks like [FlatMap](https://beam.apache.org/documentation/transforms/python/elementwise/flatmap/) accepts only iterable, let's also only support this case, no need to check here

pipeline_dp/private_beam.py
```diff
@@ -104,3 +106,36 @@ def expand(self, pcol: pvalue.PCollection) -> pvalue.PCollection:
...
+            for value in values:
+                yield key, value
+
+    def __init__(self, fn: Callable, label: Optional[str] = None):
+        super().__init__(return_anonymized=False, label=label)
+        self._fn = fn
+
+    def expand(self, pcol: pvalue.PCollection):
+        return pcol | "map values" >> beam.Map(lambda x: (x[0], self._fn(x[
+            1]))) | "flatten values" >> beam.ParDo(FlatMap._FlattenValues())
```
dvadym, line 141: Nit: call fn in constructor of `_FlattenValues` and no need in map step in this line. WDYT?

dvadym (commented):
Thanks Preethi! It looks great. I left some comments below

pipeline_dp/private_beam.py
preethiraghavan1: Yes that is a good idea. Done

pipeline_dp/private_beam.py
preethiraghavan1: Done

pipeline_dp/private_beam.py
```diff
@@ -104,3 +106,36 @@ def expand(self, pcol: pvalue.PCollection) -> pvalue.PCollection:
...
+
+class Map(PrivatePTransform):
+    """Transform class for performing Map on PrivatePCollection."""
+
+    def __init__(self, fn: Callable, label: Optional[str] = None):
+        super().__init__(return_anonymized=False, label=label)
+        self._fn = fn
+
+    def expand(self, pcol: pvalue.PCollection):
+        return pcol | "map values" >> beam.Map(lambda x: (x[0], self._fn(x[1])))
```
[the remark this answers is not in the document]
preethiraghavan1: Let me know if this looks good!

preethiraghavan1:
I messed up with fetching and had to revert a few commits.

pipeline_dp/private_beam.py
```diff
@@ -104,3 +106,43 @@ def expand(self, pcol: pvalue.PCollection) -> pvalue.PCollection:
...
+class Map(PrivatePTransform):
+    """Transform class for performing Map on PrivatePCollection."""
+
+    def __init__(self, fn: Callable, label: Optional[str] = None):
+        super().__init__(return_anonymized=False, label=label)
+        self._fn = fn
+
+    def expand(self, pcol: pvalue.PCollection):
+        return pcol | "map values" >> beam.Map(
+            lambda x: (self._privacy_id_extractor(x), self._fn(x)))
```
dvadym, line 120: It returns PCollection not PrivatePCollection. I'd suggest to do, similar way as it was done in [private_spark](https://github.com/OpenMined/PipelineDP/blob/main/pipeline_dp/private_spark.py).

Namely:

1.In `MakePrivate` to map `pcol` to PCollection `(privacy_id, value)` to pass such PCollection to the constructor of   `PrivatePCollection` .
2. In this class (i.e in Map) apply `fn` to values and to wrap the output `PCollection` by creating `PrivatePCollection` by calling the constructor of `PrivatePCollection`.
3. No need to pass `privacy_id_extractor` in `PrivatePCollection` anymore (elements of `PrivatePCollection` will be (privacy_id, value) so we will know privacy_id for each element).

WDYT?

preethiraghavan1: That's a good idea! Not having to pass the extractor all the way.

dvadym (commented):
Thanks, implementation of Map/FlatMap is LGTM!

This PR is created based on an old version of main, so it contains reverse changes from ~5-7 another pull requests. 

Could you please do the following:
1.Fetch changes in your fork (as a result form `main` branch should be synced with origin `main`)
2.Run `git pull --all` in root of the local working sources (as a result local `main` should be synced with fork `main` and hence with origin main).
3.Run
```
   git checkout map
   git merge main
```
In the last step some merge conflicts are possible. If it's hard to fix, please ping me in chat, I'll help.

dvadym (approved):
Thanks for implementing Map and FlatMap!

PR #121, opened 2021-11-19
Add report generator for dp_engine

jspacek:
## Description

This adds the report generator stages for aggregate params and contribution bounding.
It also adds a helper method for the tests so that they do not break when testing private functions.

It does not include the report stages for noise kind settings as this will be done following a refactoring. 

## Affected Dependencies
N/A

## How has this been tested?
- dp_engine_test

## Checklist

### DIFF
--- a/pipeline_dp/dp_engine.py
+++ b/pipeline_dp/dp_engine.py
@@ -106,6 +106,8 @@ class DPEngine:
             "Extract partition id")
         col = self._ops.filter_by_key(col, public_partitions,
                                       "Filtering out non-public partitions")
+        self._add_report_stage(
+            f"Public partition selection: dropped non public partitions")
         return self._ops.map_tuple(col, lambda k, v: v, "Drop key")
 
     def _bound_contributions(self, col, max_partitions_contributed: int,
@@ -134,7 +136,7 @@ class DPEngine:
             col, max_contributions_per_partition,
             "Sample per (privacy_id, partition_key)")
         self._add_report_stage(
-            f"Per-partition contribution: randomly selected not "
+            f"Per-partition contribution bounding: randomly selected not "
             f"more than {max_contributions_per_partition} contributions")
         # ((privacy_id, partition_key), [value])
         col = self._ops.map_values(
@@ -151,7 +153,7 @@ class DPEngine:
                                              "Sample per privacy_id")
 
         self._add_report_stage(
-            f"Contribution bounding: randomly selected not more than "
+            f"Cross-partition contribution bounding: randomly selected not more than "
             f"{max_partitions_contributed} partitions per user")
 
         # (privacy_id, [(partition_key, aggregator)])
@@ -191,7 +193,12 @@ class DPEngine:
 
         # make filter_fn serializable
         filter_fn = partial(filter_fn, (budget, max_partitions_contributed))
-        return self._ops.filter(col, filter_fn, "Filter private parititions")
+        self._add_report_stage(
+            lambda:
+            f"Private Partition selection: using {budget.mechanism_type.value} "
+            f"method with (eps= {budget.eps}, delta = {budget.delta})")
+
+        return self._ops.filter(col, filter_fn, "Filter private partitions")
 
     def _fix_budget_accounting_if_needed(self, col, accumulator_factory):
         """Adds MechanismSpec to accumulators.
--- a/pipeline_dp/dp_engine.py
+++ b/pipeline_dp/dp_engine.py
@@ -133,6 +133,9 @@ class DPEngine:
         col = self._ops.sample_fixed_per_key(
             col, max_contributions_per_partition,
             "Sample per (privacy_id, partition_key)")
+        self._add_report_stage(
+            f"Per-partition contribution: randomly selected not "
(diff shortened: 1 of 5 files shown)


tests/dp_engine_test.py
```diff
@@ -385,6 +384,28 @@ def test_aggregate_private_partition_selection_drop_many(self):
...
+        aggregator_params = pipeline_dp.AggregateParams(
+            noise_kind=pipeline_dp.NoiseKind.LAPLACE,
+            metrics=[agg.Metrics.COUNT],
+            max_partitions_contributed=1,
+            max_contributions_per_partition=1)
+        data_extractor = pipeline_dp.DataExtractors(
+            privacy_id_extractor=lambda x: x,
+            partition_extractor=lambda x: f"pk{x//2}",
+            value_extractor=lambda x: x)
+        dp_engine.aggregate([], aggregator_params, data_extractor)
```
dvadym, line 405: Do we need to run aggregate? Isn't a call to _add_report_stage enough?

pipeline_dp/report_generator.py
```diff
@@ -25,7 +25,7 @@ def report(self) -> str:
         if self._params is None:
             return ""
         title = f"Computing metrics: {[m.value[0] for m in self._params.metrics]}"
```
dvadym, line 27: Please m.value[0] -> m.value (and as a result will be full name of the metrics, not just first letter)

jspacek: Got it, looks much nicer now, thank you 🎉

tests/dp_engine_test.py
jspacek: Before the call to _add_report_stage, we need to add a first element to the array that includes the ReportGenerator object, otherwise it will throw an array index out of bounds exception for [-1].

But there is no need to call the entire aggregate function here, so I replaced it with a simpler call to add the first element into the array 😸 

Thank you!

pipeline_dp/dp_engine.py
```diff
@@ -147,8 +150,11 @@ def _bound_contributions(self, col, max_partitions_contributed: int,
         col = self._ops.sample_fixed_per_key(col, max_partitions_contributed,
                                              "Sample per privacy_id")
 
-        # (privacy_id, [(partition_key, aggregator)])
+        self._add_report_stage(
+            f"Contribution bounding: randomly selected not more than "
```
dvadym, line 154: Nit: Cross-partition contribution bounding

dvadym (commented):
Please add also information about partition selection.
Namely currently there are 2 places:
1. Function `_drop_not_public_partitions`  -> this function drops partitions which are not in `public_partitions` (this public partition case)
2. `_select_private_partitions` (this is private partitions case).

pipeline_dp/dp_engine.py
jspacek: Fixed, thanks for catching!

jspacek:
> Please add also information about partition selection. Namely currently there are 2 places:
> 
> 1. Function `_drop_not_public_partitions`  -> this function drops partitions which are not in `public_partitions` (this public partition case)
> 2. `_select_private_partitions` (this is private partitions case).

@dvadym Added some information for public and private partition selection and testing (but did not include the number of partitions).

pipeline_dp/dp_engine.py
```diff
@@ -185,6 +192,7 @@ def filter_fn(captures: Tuple[MechanismSpec, int],
 
         # make filter_fn serializable
         filter_fn = partial(filter_fn, (budget, max_partitions_contributed))
+        self._add_report_stage(f"Select private partitions")
```
dvadym, line 195: Maybe to have format for all messages like that:
<name of the stage>: <some additional information> (for contribution bounding it's the case).

Here it might be like
**Private Partition selection: using Truncated Geometric method with (eps=..., delta = )**

WDYT?

pipeline_dp/dp_engine.py
```diff
@@ -106,6 +106,7 @@ def _drop_not_public_partitions(self, col, public_partitions,
             "Extract partition id")
         col = self._ops.filter_by_key(col, public_partitions,
                                       "Filtering out non-public partitions")
+        self._add_report_stage(f"Dropped non public partitions")
```
dvadym, line 109: Maybe to have format for all messages like that:
: (for contribution bounding it's the case).

Here it might be like
**Public partition selection: dropped non public partitions**

WDYT?

pipeline_dp/dp_engine.py
dvadym: `eps, delta` can be taken from `budget`. But eps, delta are evaluated lazily, so which is not a problem, you can pass lambda returning to `_add_report_stage`, i.e.
_add_report_stage(lambda : f".... {budget.eps} ...")

jspacek: Looks great, I've implemented this lazily 😎

pipeline_dp/dp_engine.py
jspacek: Yes, this looks great. Completed!

dvadym (commented):
Thanks for implementing generation of reports!

PR #126, opened 2021-12-06
Introduce AccumulatorFactory hierarchy

dvadym:
This PR is a refactoring of way how accumulators are created during aggregation. [`Accumulator`](https://github.com/OpenMined/PipelineDP/blob/0cb701891eca47c55ee2ba93bc1cf5470a30a242/pipeline_dp/accumulator.py#L57) is an object that does aggregation.

**Before this PR:**
`AccumulatorFactory` contains a list of `AccumulatorParams`. `AccumulatorParams` are used for creating accumulators during aggregation.

**After this PR:**
1. `AccumulatorFactory` is renamed to `CompoundAccumulatorFactory` 
2. Abstract `AccumulatorFactory` is introduced and concrete `AccumulatorFactory` creates corresponding accumulators (e.g.`CountAccumulatorFactory` creates `CountAccumulator`)

**Why this refactoring is needed?**
This PR is 1st step of refactoring of computing metrics with `CombineFn` functions (see [on the page](https://beam.apache.org/documentation/transforms/python/aggregation/combineperkey/)  example 5). Combing with `CombineFn` is more effective and required for more tight integration with Beam.

Also this refactoring itself makes creating of accumulators more direct and as a result more readable.

### DIFF
--- a/pipeline_dp/accumulator.py
+++ b/pipeline_dp/accumulator.py
@@ -18,7 +18,7 @@ def merge(accumulators: typing.Iterable['Accumulator']) -> 'Accumulator':
     return reduce(lambda acc1, acc2: acc1.add_accumulator(acc2), accumulators)
 
 
-def create_accumulator_factories(
+def _create_accumulator_factories(
     aggregation_params: pipeline_dp.AggregateParams,
     budget_accountant: budget_accounting.BudgetAccountant
 ) -> typing.List['AccumulatorFactory']:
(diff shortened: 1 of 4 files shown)


pipeline_dp/accumulator.py
```diff
@@ -13,45 +13,34 @@
...
-    constructor_params: typing.Any
-
-
 def merge(accumulators: typing.Iterable['Accumulator']) -> 'Accumulator':
     """Merges the accumulators."""
     return reduce(lambda acc1, acc2: acc1.add_accumulator(acc2), accumulators)
 
 
-def create_accumulator_params(
+def create_accumulator_factories(
```
sushkoy, line 21: Should the convention to prefix internal methods with "_" apply here?

sushkoy (approved):

Issue #124, opened 2021-12-03
Ensure unique stage labels in BeamOperations


dvadym:
# Context
Python Beam syntax:

`collection | "label" >> transform`
eg.
`result = col | "map >> beam.Map(lambda ...)`

In Beam each transform should have unique labels. 

`BeamOperations` ([src](https://github.com/OpenMined/PipelineDP/blob/0cb701891eca47c55ee2ba93bc1cf5470a30a242/pipeline_dp/pipeline_operations.py#L98)) is PipelineDP connector to Beam. Most functions of `BeamOperations` are wrappers around corresponding Beam transforms, e.g.

```
class BeamOperations(PipelineOperations):
    def map(self, col, fn, stage_name: str):
        return col | stage_name >> beam.Map(fn)
```

Currently the following pipeline 

```
result1 = dp_engine.aggregate(...)
result2 = dp_engine.aggregate(...)
```
fails, because both calls of `aggregate` would have the same `stage_names`.

# Goal
To introduce class `UniqueStageNamesGenerator`, which keeps all already used stage names and have API like that

```
class UniqueLabelsGenerator:
    def unique(self, label):
        1. Adds 0, 1, ... to label to ensure uniqueness
        2. Save the unique label + str(number)
        3. Returns this unique label
```

Then `BeamOperations` can use this class for generating unique labels.

PR #128, opened 2021-12-22
Multiple stage names generator

jspacek:
## Description
Adds a class to generate unique stage names in Beam so we can use the same names in multiple pipelines. See #124.
## How has this been tested?
- Unit tests in `pipeline_operations_tests` for all beam operations.

## Checklist

### DIFF
--- a/pipeline_dp/pipeline_operations.py
+++ b/pipeline_dp/pipeline_operations.py
@@ -1,20 +1,15 @@
 """Adapters for working with pipeline frameworks."""
 
-from enum import Enum
 from functools import partial
-import os
 import multiprocessing as mp
-from pipeline_dp import accumulator
 import random
 import numpy as np
 
 import abc
 import apache_beam as beam
 import apache_beam.transforms.combiners as combiners
-import collections
 import pipeline_dp.accumulator as accumulator
 import typing
-from typing import Any, Optional, Callable, Tuple
 import collections
 import itertools
 
@@ -95,20 +90,50 @@ class PipelineOperations(abc.ABC):
         return False
 
 
+class UniqueLabelsGenerator:
+    """Generate unique labels for each pipeline aggregation."""
+
+    def __init__(self, suffix):
+        self._labels = set()
+        self._suffix = suffix
+
+    def _add_if_unique(self, label):
+        if label in self._labels:
+            return False
+        self._labels.add(label)
+        return True
+
+    def unique(self, label):
+        if not label:
+            label = "UNDEFINED_STAGE_NAME"
+        suffix_label = label + "_" + self._suffix
+        if self._add_if_unique(suffix_label):
+            return suffix_label
+        for i in itertools.count(1):
+            label_candidate = f"{label}_{i}_{self._suffix}"
+            if self._add_if_unique(label_candidate):
+                return label_candidate
+
+
 class BeamOperations(PipelineOperations):
     """Apache Beam adapter."""
 
+    def __init__(self, suffix: str = ""):
+        super().__init__()
+        self._ulg = UniqueLabelsGenerator(suffix)
+
     def map(self, col, fn, stage_name: str):
-        return col | stage_name >> beam.Map(fn)
+        return col | self._ulg.unique(stage_name) >> beam.Map(fn)
 
     def flat_map(self, col, fn, stage_name: str):
-        return col | stage_name >> beam.FlatMap(fn)
+        return col | self._ulg.unique(stage_name) >> beam.FlatMap(fn)
 
     def map_tuple(self, col, fn, stage_name: str):
-        return col | stage_name >> beam.Map(lambda x: fn(*x))
+        return col | self._ulg.unique(stage_name) >> beam.Map(lambda x: fn(*x))
 
     def map_values(self, col, fn, stage_name: str):
-        return col | stage_name >> beam.MapTuple(lambda k, v: (k, fn(v)))
+        return col | self._ulg.unique(stage_name) >> beam.MapTuple(lambda k, v:
+                                                                   (k, fn(v)))
 
     def group_by_key(self, col, stage_name: str):
         """Group the values for each key in the PCollection into a single sequence.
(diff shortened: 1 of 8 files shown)


pipeline_dp/pipeline_operations.py
```diff
@@ -95,20 +90,50 @@ def is_serialization_immediate_on_reduce_by_key(self):
...
+        return True
+
+    def unique(self, label):
+        if not label:
+            label = "UNDEFINED_STAGE_NAME"
+        suffix_label = label + "_" + self._suffix
+        if self._add_if_unique(suffix_label):
+            return suffix_label
+        for i in itertools.count(1):
+            label_candidate = label + "_" + str(i) + "_" + self._suffix
```
jspacek, line 113: @dvadym WDYT about appending the counter to after the label, instead of the suffix? It may be easier to trace a pipeline because the name would remain constant, eg.

Instead of `MAP_NAME_PIPELINE_SUFFIX_01` and `MAP_NAME_PIPELINE_SUFFIX _02`
We would have `MAP_NAME_1_PIPELINE_SUFFIX_0` and `MAP_NAME_2_PIPELINE_SUFFIX_0`

Does it make a difference when we want to build the execution graph? PTAL, thank you!

pipeline_dp/pipeline_operations.py
```diff
@@ -154,29 +179,32 @@ def does_keep(pk_val):
...
-        return col | stage_name >> combiners.Sample.FixedSizePerKey(n)
+        return col | self._ulb.unique(
+            stage_name) >> combiners.Sample.FixedSizePerKey(n)
 
     def count_per_element(self, col, stage_name: str):
-        return col | stage_name >> combiners.Count.PerElement()
+        return col | self._ulb.unique(
+            stage_name) >> combiners.Count.PerElement()
 
     def reduce_accumulators_per_key(self, col, stage_name: str = None):
```
jspacek, line 209: @dvadym Do we want any others to have an optional `stage_name`?

tests/pipeline_operations_test.py
```diff
@@ -72,14 +72,107 @@ def test_reduce_accumulators_per_key(self):
...
+        with test_pipeline.TestPipeline() as p:
+            p | f"SAME_BEAM_CREATE_NAME" >> beam.Create([(6, 1), (6, 2)])
+            with self.assertRaisesRegex(
+                    RuntimeError,
+                    expected_regex="A transform with label "
+                    "\"SAME_BEAM_CREATE_NAME\" already exists "
+                    "in the pipeline."):
+                p | f"SAME_BEAM_CREATE_NAME" >> beam.Create([(6, 1), (6, 2)])
+
+    def test_ops_stage_name_must_be_unique(self):
```
dvadym, line 96: I have a suggestion to restructure tests in the following way.

1.To have unit tests for `UniqueLabelGenerators`
2. To test all Beam function by mocking everything what's needed, i.e. w/o setup pipelines just to test that  `UniqueLabelGenerators` is called in all places like  `col | ulg.unique_label >> transform`

Tests for `UniqueLabelGenerators ` are just standard unittests. 
I'll explain how to do mocking for `map`:
1. Introduce `MockUniqueLabelGenerators:

```
 class MockUniqueLabelGenerators:
         def unique():
              return "unique_label"
```
2. Introduce MockPcollection
```
from unittest.mock import Mock, MagicMock, patch
def create_mock_pcollection():
    mock = Mock()
    mock.__or__ = MagicMock(return_value = mock)
    return mock
```
3.Testing of Map
 ```
 @patch("apache_beam.transforms.ptransform.PTransform.__rrshift__")
  def test_map(self, mock_rrshift):
        mock_pcollection = create_mock_pcollection()
        ops = BeamOperations(MockUniqueLabelGenerators())
        ops.map(mock_pcollection, lambda x:x, "stage_name")
        mock_rrshift.assert_called_once_with("unique_value")
```

Explanation for patch:
\@patch  creates mock for `>>` in expression   col | "stage_name" >> beam.Map(fn), and in line
` mock_rrshift.assert_called_once_with("unique_value")` we check that the argument of `>>` is acctually value returned from MockUniqueLabelGenerator

WDYT?

Please do it for now only for `map` and then in next step we will generalize it for all methods

**Review by dvadym — COMMENTED** [review]
Thanks it looks good! I left some comments

File: pipeline_dp/pipeline_operations.py [hunk 4]
```diff
@@ -95,20 +90,50 @@ def is_serialization_immediate_on_reduce_by_key(self):
...
+            if self._add_if_unique(label_candidate):
+                return label_candidate
+
+
 class BeamOperations(PipelineOperations):
     """Apache Beam adapter."""
 
+    def __init__(self, suffix):
+        super().__init__()
+        self._ulb = UniqueLabelsGenerator(suffix)
```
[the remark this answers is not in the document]
jspacek: Oops!! Fixed 👍

tests/pipeline_operations_test.py
```diff
@@ -72,14 +72,107 @@ def test_reduce_accumulators_per_key(self):
...
             result = col | "Get accumulated values" >> beam.Map(
                 lambda row: (row[0], row[1].get_metrics()))
 
             beam_util.assert_that(result,
                                   beam_util.equal_to([(6, 2), (7, 2), (8, 1)]))
 
 
+class BeamOperationsStageNameTest(unittest.TestCase):
+
+    def test_beam_create_stage_name_must_be_unique(self):
```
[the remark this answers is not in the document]
jspacek: Makes sense, thanks!

tests/pipeline_operations_test.py
jspacek: Awesome! I added tests for all the functions in Beam Pipeline operations.
WDYT about adding 'MockUniqueLabelGenerators' and `create_mock_pcollection` to the `BeamOperationsStageNameTest` class?

jspacek: @dvadym Not sure why, as I recently merged main into this branch, but one of the test fails starting this morning: https://github.com/OpenMined/PipelineDP/runs/4777317330
Is it OK?

dvadym: Thanks a lot! 

It looks like this test fails because of some Spark internal stuff. I'll investigate it later, it's independent of the current PR.

dvadym (changes requested):
Thanks a lot!

dvadym (approved):

tests/pipeline_operations_test.py
jspacek: Thanks for looking into this! 🚀

PR #135, opened 2022-01-13
Implement adding empty public partitions

### LINKED ISSUE #110: Adding missing public partitions
# Context

Atm public partition implementation is not complete. Filtering out partitions which are not in `public_partitions` is performed ([code](https://github.com/OpenMined/PipelineDP/blob/545166054b8ac1750dafa752165323b08ac7ee8e/pipeline_dp/dp_engine.py#L63)), but adding those public partitions which are not in data has not been implemented. This issue for implementing that.

# Design

Method [DPEngine.aggregate](https://github.com/OpenMined/PipelineDP/blob/545166054b8ac1750dafa752165323b08ac7ee8e/pipeline_dp/dp_engine.py#L43) does high-level management of the computation graph (contribution bounding, partition selection etc). The computation graph now is the following.

1. Drop non public partitions ([code](https://github.com/OpenMined/PipelineDP/blob/545166054b8ac1750dafa752165323b08ac7ee8e/pipeline_dp/dp_engine.py#L63))
2. Bound contributions (and aggregation per (**partition_key**, **privacy_id**))
3. Aggregation per **partition_key**.
4. Computing DP metrics.

 It looks like a pretty good place (please advice on better places if there are) to add missing public partition is between steps 2 and 3. We can merge (flatten) existed collection of (partition_key, acumulator) with (**public_partitions**, **accumulator_for_empty_set**). 

[aggregate_fn](https://github.com/OpenMined/PipelineDP/blob/545166054b8ac1750dafa752165323b08ac7ee8e/pipeline_dp/dp_engine.py#L60) should return **accumulator_for_empty_set** when called with **[]**.

miracvbasaran:
## Description
Fixes #110.

## Affected Dependencies
n/a

## How has this been tested?
- Unit tests
- movie ratings example

## Checklist

### DIFF
--- a/pipeline_dp/dp_engine.py
+++ b/pipeline_dp/dp_engine.py
@@ -91,6 +91,13 @@ class DPEngine:
         col = self._ops.map_tuple(col, lambda pid_pk, v: (pid_pk[1], v),
                                   "Drop privacy id")
         # col : (partition_key, accumulator)
+
+        if params.public_partitions:
+            col = self._add_empty_public_partitions(col,
+                                                    params.public_partitions,
+                                                    aggregator_fn)
+        # col : (partition_key, accumulator)
+
         col = self._ops.reduce_accumulators_per_key(
             col, "Reduce accumulators per partition key")
         # col : (partition_key, accumulator)
@@ -180,6 +187,19 @@ class DPEngine:
             f"Public partition selection: dropped non public partitions")
         return self._ops.map_tuple(col, lambda k, v: v, "Drop key")
 
+    def _add_empty_public_partitions(self, col, public_partitions,
+                                     aggregator_fn):
+        self._add_report_stage(
+            "Adding empty accumulators to public partitions and joining "
+            "them with accumulators from data")
+        empty_accumulators = self._ops.map(
+            public_partitions, lambda partition_key:
+            (partition_key, aggregator_fn([])))
+
+        return self._ops.flatten(
+            col, empty_accumulators,
+            "Join public partitions with partitions from data")
+
     def _bound_contributions(self, col, max_partitions_contributed: int,
                              max_contributions_per_partition: int,
                              aggregator_fn):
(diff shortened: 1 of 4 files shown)


tests/dp_engine_test.py
```diff
@@ -400,7 +400,7 @@ def test_aggregate_private_partition_selection_drop_many(self):
         # Most partition should be dropped by private partition selection.
         # This tests is non-deterministic, but it should pass with probability
         # very close to 1.
-        self.assertLess(len(col), 5)
+        self.assertLess(5, len(col))
```
miracvbasaran, line 403: "expected" first, then "actual"

pipeline_dp/dp_engine.py
```diff
@@ -180,6 +187,19 @@ def _drop_not_public_partitions(self, col, public_partitions,
             f"Public partition selection: dropped non public partitions")
         return self._ops.map_tuple(col, lambda k, v: v, "Drop key")
 
+    def _add_empty_public_partitions(self, col, public_partitions,
+                                     aggregator_fn):
+        self._add_report_stage(
+            "Adding empty accumulators to public partitions and joining "
```
dvadym, line 193: Those reports are for API users. Accumulators are not part of the API. 
Maybe something like that
"Adding empty partitions for those public partitions which are missing in data" 

Technically speaking right here we're doing in slightly different way, but it's equavalent. WDYT?

tests/dp_engine_test.py
dvadym: Thanks!

pipeline_dp/dp_engine.py
```diff
@@ -180,6 +187,19 @@ def _drop_not_public_partitions(self, col, public_partitions,
             f"Public partition selection: dropped non public partitions")
         return self._ops.map_tuple(col, lambda k, v: v, "Drop key")
 
+    def _add_empty_public_partitions(self, col, public_partitions,
+                                     aggregator_fn):
```
dvadym, line 191: Please add description to this function

dvadym (commented):
Thanks, it looks great! I left minor comments

pipeline_dp/pipeline_operations.py
```diff
@@ -86,6 +86,14 @@ def reduce_accumulators_per_key(self, col, stage_name: str):
         """
         pass
 
+    @abc.abstractmethod
+    def flatten(self, col1, col2, stage_name: str):
```
dvadym, line 90: The problem is that there are other PipelineOperations, and w/o implementaiton this method in them, they are abstract clasess which can't be created.

Please add this method in other PipelineOperations:

In MultiProcessPipelineOperations please use implementation from LocalPipelineOpeations

In SparkPipelineOperations please add this method with implementation
`raise NotImplementedException("...") `
I'll implement it
In Spark please

dvadym (approved):

PR #141, opened 2022-01-15
Add PrivacyIdCount Transform to Beam's public API.

yuan2z:
Add PrivacyIdCount Transform to Beam's public API.
Make test names more descriptive and change test names to a consistent format (snake case).

## How has this been tested?
- Unit tested.


dvadym (approved):
Thanks a lot for implementing of PrivacyIdCount transform!

PR #144, opened 2022-01-17
DP Count combiner

dvadym:


pipeline_dp/combiners.py
```diff
@@ -25,3 +30,49 @@ def merge_accumulators(self, accumulator1, accumulator2):
     def compute_metrics(self, accumulator: 'Accumulator'):
         """Computes and returns the result of aggregation."""
         pass
+
+
+class CombinerParams:
+    """Parameters for an combiner.
```
sushkoy, line 36: nit: a combiner

pipeline_dp/combiners.py
```diff
@@ -25,3 +30,49 @@ def merge_accumulators(self, accumulator1, accumulator2):
...
+    def __init__(self, params: CombinerParams):
+        self._params = params
+
+    def create_accumulator(self, values) -> int:
+        return len(values)
+
+    def merge_accumulators(self, accumulator1: int, accumulator2: int):
+        return accumulator1 + accumulator2
+
+    def compute_metrics(self, accumulator: int) -> float:
```
sushkoy, line 76: The parent declaration uses type 'Accumulator', but here we say 'int'. Should we remove 'Accumulator' as the type from the parent class, and just say that any class can be used here?

pipeline_dp/combiners.py
```diff
@@ -25,3 +30,49 @@ def merge_accumulators(self, accumulator1, accumulator2):
...
+            self.aggregate_params.noise_kind)
+
+
+class CountCombiner(Combiner):
+
+    def __init__(self, params: CombinerParams):
+        self._params = params
+
+    def create_accumulator(self, values) -> int:
+        return len(values)
```
sushkoy, line 71: Why does this return length of the array? Maybe we can add some docs to the parent method to document the expectation for the return type.

pipeline_dp/combiners.py
```diff
@@ -25,3 +30,49 @@ def merge_accumulators(self, accumulator1, accumulator2):
     def compute_metrics(self, accumulator: 'Accumulator'):
         """Computes and returns the result of aggregation."""
         pass
+
+
+class CombinerParams:
+    """Parameters for an combiner.
+
+    Wraps epsilon and delta from the MechanismSpec which are lazily loaded.
```
sushkoy, line 38: nit: I'm likely missing some context.. why is that lazily loaded, and where is this loading happens?

nit2: Maybe we can stay less technical in the class doc, and add more comments to the code. Maybe we can say here something like "wraps all the information needed by the combiner to do the differentially-private computation, e.g. privacy budget and mechanism.

WDYT?

pipeline_dp/combiners.py
dvadym: I've added more comments on Combiner based class on how the Combiner framework works and on CountCombiner class. Please check whether it becomes more clear

pipeline_dp/combiners.py
dvadym: Thanks, I've removed type declarations from base class, they are incorrect

dvadym (commented):
Thanks for comments! PTAL

pipeline_dp/combiners.py
```diff
@@ -9,6 +14,18 @@ class Combiner(abc.ABC):
     aggregation state. Combiners contain logic, while accumulators contain data.
     The API of combiners are inspired by Apache Beam CombineFn class.
     https://beam.apache.org/documentation/transforms/python/aggregation/combineperkey/#example-5-combining-with-a-combinefn
+
+    Let we have some dataset X to aggregate. The workflow of running an
```
sushkoy, line 18: (nit) Here's how PipelineDP uses combiners to performs an aggregation on some dataset X:
1. ....

pipeline_dp/combiners.py
```diff
@@ -9,6 +14,18 @@ class Combiner(abc.ABC):
...
+    aggregation with combiners on X is the following:
+    1.Split dataset X on sub-datasets which might be kept in memory.
+    2.Call create_accumulators() for each sub-dataset and keep resulting accumulators.
+    3.Choosing any pair of accumulators and merge them by calling merge_accumulators().
+    4.Continue 3 until 1 accumulator is left.
+    5.Call compute_metrics() for the accumulator that left.
+
+    Assumption: merge_accumulators is associative binary operation.
+
+    The type of the accumulator is specific for each concrete Combiner.
```
sushkoy, line 28: nit: 

The type of accumulator depends on the aggregation performed by this Combiner. For example, this can be a primitive type (e.g. int for a simple "count" aggregation) or a more complex structure (e.g. for "mean")

pipeline_dp/combiners.py
```diff
@@ -9,6 +14,18 @@ class Combiner(abc.ABC):
     aggregation state. Combiners contain logic, while accumulators contain data.
     The API of combiners are inspired by Apache Beam CombineFn class.
     https://beam.apache.org/documentation/transforms/python/aggregation/combineperkey/#example-5-combining-with-a-combinefn
+
+    Let we have some dataset X to aggregate. The workflow of running an
+    aggregation with combiners on X is the following:
+    1.Split dataset X on sub-datasets which might be kept in memory.
```
sushkoy, line 20: supernit: Space after dots

pipeline_dp/combiners.py
sushkoy: Yes that's much clearer, thank you!

sushkoy (approved):

pipeline_dp/combiners.py
dvadym: Thanks!

pipeline_dp/combiners.py
dvadym: Thanks!

PR #154, opened 2022-01-19
Implement MeanCombiner

miracvbasaran:
## Description
- Implement MeanCombiner. Remaning work involves dealing with converting the output of MeanCombiner to the appropriate format depending on what aggregations (mean, sum, count) are requested. First, DPEngine would need to use the new combiners though.
- Also make some improvements to how mean is computed in dp_computations.

## How has this been tested?
- Unit tests

## Checklist


pipeline_dp/combiners.py
```diff
@@ -97,3 +99,29 @@ def merge_accumulators(self, count1: int, count2: int):
...
+
+    def __init__(self, params: CombinerParams):
+        self._params = params
+
+    def create_accumulator(self, values) -> (int, float):
+        return len(values), np.clip(values, self._params.aggregate_params.low,
+                                    self._params.aggregate_params.high).sum()
+
+    def merge_accumulators(self, accum1: tuple, accum2: tuple):
+        return accum1[0] + accum2[0], accum1[1] + accum2[1]
```
dvadym, line 121: Nit:
what do you think about the following for readbility:
count1, sum1 = accum1
count2, sum2 = accum2
and then use count, sum?
and the same for compute_metrics

pipeline_dp/combiners.py
```diff
@@ -97,3 +99,29 @@ def merge_accumulators(self, count1: int, count2: int):
...
+
+    def compute_metrics(self, accum: tuple) -> namedtuple:
+        noisy_count, noisy_sum, noisy_mean = dp_computations.compute_dp_mean(
+            accum[0], accum[1], self._params.mean_var_params)
+        MeanTuple = namedtuple('MeanTuple', ['count', 'sum', 'mean'])
```
dvadym, line 126: Please move `MeanTuple` definition before `class MeanCombiner(Combiner):`
(it's more effective to run definition once than on each compute_metrics

pipeline_dp/dp_computations.py
```diff
@@ -33,6 +33,8 @@ def squares_interval(self):
 
 def compute_middle(low: float, high: float):
     """"Returns the middle point of the interval [low, high]."""
+    # (low + high) / 2 may cause an overflow or loss of precision if low and
+    # high are large.
```
dvadym, line 37: thanks for adding comments!

pipeline_dp/dp_computations.py
```diff
@@ -304,7 +306,11 @@ def _compute_mean(
     dp_normalized_sum = _add_random_noise(normalized_sum, eps, delta,
                                           l0_sensitivity, linf_sensitivity,
                                           noise_kind)
-    return dp_normalized_sum / dp_count + middle
+    # Clamps dp_count to 1.0. We know that actual count > 1 except when the
+    # input set is empty, in which case it shouldn't matter much what the
+    # denominator is.
+    dp_count_clamped = max(1.0, dp_count)
```
dvadym, line 312: Thanks!

pipeline_dp/dp_computations.py
```diff
@@ -347,7 +353,10 @@ def compute_dp_mean(count: int, sum: float, dp_params: MeanVarParams):
         dp_params.max_contributions_per_partition,
         dp_params.noise_kind,
     )
-    return dp_count, dp_mean * dp_count, dp_mean
+
+    # Clamps dp_mean to be in the range [low, high].
```
dvadym, line 357: Atm PipelineDP doesn't do any postprocessing on DP results. The idea is that in case of API users would like to do statistical analysis, postprocessing would make things harder. 

We can introduce some standard postprocessing (like clipping to high, low or rounding), but there should be way to turn off it and we probably would like to introduce post-processing layer, separate from core DP layer. So I'd suggest not to clip mean for now and then to think about post-processing. WDYT?

dvadym (commented):

XinyuYe:
Hi Mirac, 
Some changes about the Combiners here https://github.com/OpenMined/PipelineDP/commit/0d92d0d44bfb5f88f7fc0929dfda6efc7e020e30 may influence some of your change. You may want to consider rebasing on the current head. Sorry for the inconvenience.

dvadym:
Could you please add creation of MeanCombiner in 
[create_compound_combiner](https://github.com/OpenMined/PipelineDP/blob/main/pipeline_dp/combiners.py#L199), which is a factory method for creating all concrete combiners ?

pipeline_dp/combiners.py
miracvbasaran: Makes sense, done.

pipeline_dp/dp_computations.py
miracvbasaran: Separating the post-processing from the core DP layer makes sense. We can do this at the dp_engine layer (in `aggregate()`) perhaps.

Removing this for now.

PR #155, opened 2022-01-19
Fix and test Beam count  and PrivacyIdCount 

Kulankhina:
Make public partitions optional
Delete value_extractor because it isn't used
Clean-up CountParams comments
Test that Count result is sensible

### DIFF
--- a/pipeline_dp/aggregate_params.py
+++ b/pipeline_dp/aggregate_params.py
@@ -104,7 +104,7 @@ class SumParams:
             partition.
         public_partitions: A collection of partition keys that will be present in
             the result.
-        partition_extractor: A function for partition id extraction from an RDD record.
+        partition_extractor: A function for partition id extraction from a collection record.
         value_extractor: A function for extraction of value
             for which the sum will be calculated.
   """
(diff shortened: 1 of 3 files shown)


pipeline_dp/aggregate_params.py
```diff
@@ -129,21 +129,20 @@ class CountParams:
             unit of privacy (e.g., a user) can participate.
         max_contributions_per_partition: Bounds the number of times one unit of
             privacy (e.g. a user) can contribute to a partition.
+        partition_extractor: A function for partition id extraction from an RDD record.
```
dvadym, line 132: please RDD record -> record or "collection record"
 everywhere in this file

Explanation: RDD is a Spark term, this class was introduced initially for Spark only and than used for Beam, but the comment stale

dvadym (commented):
Thanks it looks great! I left minor comment

pipeline_dp/aggregate_params.py
Kulankhina: Thanks! fixed.

pipeline_dp/aggregate_params.py
```diff
@@ -104,7 +104,7 @@ class SumParams:
             partition.
         public_partitions: A collection of partition keys that will be present in
             the result.
-        partition_extractor: A function for partition id extraction from an RDD record.
+        partition_extractor: A function for partition id extraction from an collection record.
```
dvadym, line 107: an collection -> a collection (and in other places)

Kulankhina: Thanks! Done.

dvadym (approved):
Thanks!

PR #158, opened 2022-01-20
Test Beam Count, Sum, PIdCoun with pub partitions

Kulankhina:
... return sensible result

plus delete a forgotten TODO


dvadym (approved):
Thanks a lot of creating tests!

PR #156, opened 2022-01-20
Uses combiners in dp engine.

XinyuYe:
## Description
Uses combiners in dp engine. Please take a look to see if things look good. Will fix tests later. 

There is a potential clean up to use named tuple in CompoundCombiners to make code cleaner. Will do that after the logics look correct.

## Affected Dependencies
Could affect all places using dp aggregations.

## How has this been tested?
Most unit tests except for two. One is still missing mean combiners and the other uses mock of Accumulator. Will fix after the logics look correct.

PR #159, opened 2022-01-20
Combiner tests

dvadym:
This PR:

1.Fixes dp_engine_test, which were broken after [PR](https://github.com/OpenMined/PipelineDP/pull/156) which introduced usage of CombinerFramework instead of Accumulator Framework
2.Adds tests for PrivacyIdCombiner and CreateCompoundAccumulator

### DIFF
--- a/pipeline_dp/combiners.py
+++ b/pipeline_dp/combiners.py
@@ -8,27 +8,6 @@ from pipeline_dp import dp_computations
 from pipeline_dp import budget_accounting
 
 
-def create_compound_combiners(
-    aggregation_params: pipeline_dp.AggregateParams,
-    budget_accountant: budget_accounting.BudgetAccountant
-) -> 'CompoundCombiner':
-    combiners = []
-    mechanism_type = aggregation_params.noise_kind.convert_to_mechanism_type()
-    if pipeline_dp.Metrics.COUNT in aggregation_params.metrics:
-        budget = budget_accountant.request_budget(mechanism_type)
-        combiners.append(
-            CountCombiner(CombinerParams(budget, aggregation_params)))
-    if pipeline_dp.Metrics.SUM in aggregation_params.metrics:
-        budget = budget_accountant.request_budget(mechanism_type)
-        combiners.append(SumCombiner(CombinerParams(budget,
-                                                    aggregation_params)))
-    if pipeline_dp.Metrics.PRIVACY_ID_COUNT in aggregation_params.metrics:
-        budget = budget_accountant.request_budget(mechanism_type)
-        combiners.append(
-            PrivacyIdCountCombiner(CombinerParams(budget, aggregation_params)))
-    return CompoundCombiner(combiners)
-
-
 class Combiner(abc.ABC):
     """Base class for all combiners.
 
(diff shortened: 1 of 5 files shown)


pipeline_dp/combiners.py
```diff
@@ -208,3 +187,29 @@ def compute_metrics(
         for combiner, acc in zip(self._combiners, accumulator):
             metrics.append(combiner.compute_metrics(acc))
         return (privacy_id_count, metrics)
+
+
+def create_compound_combiners(
+        aggregate_params: pipeline_dp.AggregateParams,
+        budget_accountant: budget_accounting.BudgetAccountant
+) -> CompoundCombiner:
```
dvadym, line 195: Just move (it's better to have it after CompoundCombiner definition)

tests/dp_engine_test.py
```diff
@@ -363,7 +362,7 @@ def test_aggregate_report(self, mock_create_accumulator_factories_function):
             "Differentially private: Computing <Metrics: ['privacy_id_count', 'count', 'mean']>"
             "\n1. Per-partition contribution bounding: randomly selected not more than 2 contributions"
             "\n2. Cross-partition contribution bounding: randomly selected not more than 3 partitions per user"
-            "\n3. Private Partition selection: using Truncated Geometric method with (eps= 0.5, delta = 5e-11)"
+            "\n3. Private Partition selection: using Truncated Geometric method with (eps= 0.1111111111111111, delta = 1.1111111111111111e-11)"
```
sushkoy, line 365: Do you know why this has changed?

pipeline_dp/combiners.py
```diff
@@ -208,3 +187,29 @@ def compute_metrics(
         for combiner, acc in zip(self._combiners, accumulator):
             metrics.append(combiner.compute_metrics(acc))
         return (privacy_id_count, metrics)
+
+
+def create_compound_combiners(
```
sushkoy, line 192: create_compound_combiner (without "s")? It looks like the method returns a single compound combiner (even if there are multiple child combiners)

pipeline_dp/combiners.py
```diff
@@ -208,3 +187,29 @@ def compute_metrics(
...
+        aggregate_params: pipeline_dp.AggregateParams,
+        budget_accountant: budget_accounting.BudgetAccountant
+) -> CompoundCombiner:
+    combiners = []
+    mechanism_type = aggregate_params.noise_kind.convert_to_mechanism_type()
+
+    if pipeline_dp.Metrics.COUNT in aggregate_params.metrics:
```
sushkoy, line 199: Are all metrics in aggregate_params.metrics expected to have a combiner?

If yes, it might be safer to iterate through the list, and then throw an exception if there's a metric for which we don't support a combiner. 

for metric in aggregate_params.metric:
  if metric == pipeline_dp.Metrics.COUNT:
    ..
  elseif metric = pipeline_dp.Metrics.SUM:
    ..

  else:
     throw Error ("Unsupporter metric:" + metric)


WDYT?

pipeline_dp/combiners.py
dvadym: Thanks, correct!

pipeline_dp/combiners.py
dvadym: Actually it's not 1:1 correspondence, 1 combiner can have more than 1 metrics to compute. E.g. MeanCombiner (which is currently WIP), this Combiner along the way computes Count and Sum. So with adding MeanCombiner this function needs to be significantly extended.

tests/dp_engine_test.py
dvadym: Yes, previously there was no aggregation, because it was mocked. Now we don't need to aggregate, that's why real combiner was created and it requests for budget, that's why less budget on partition selection

pipeline_dp/combiners.py
sushkoy: Thanks for the explanation! I'm still a bit unsure how the mean combiner will look here, but maybe we can defer this until we have the mean combiner. Thanks!

sushkoy:
LGTM.

PR #160, opened 2022-01-21
Fix select partitions

dvadym:
This PR contains:
  1. Fix for sampling inside inside select_private_partition (more details in comments in code) which fixes running select partitions in Beam
  2. Fixes in stage names
  3. A simple fix for private_spark tests (removed setting not existed anymore variable)

### DIFF
--- a/pipeline_dp/dp_engine.py
+++ b/pipeline_dp/dp_engine.py
@@ -78,6 +78,7 @@ class DPEngine:
             col = self._drop_not_public_partitions(col,
                                                    params.public_partitions,
                                                    data_extractors)
+
         # Extract the columns.
         col = self._ops.map(
             col, lambda row: (data_extractors.privacy_id_extractor(row),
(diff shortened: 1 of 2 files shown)


pipeline_dp/dp_engine.py
```diff
@@ -207,14 +208,27 @@ def select_private_partitions(self, col,
...
+                # 1. Apache Beam can not serialize numpy types.
+                # 2. It might lead for losing precision (e.g. arbitrary
+                # precision int is converted to int64).
+                # So np.random.choice should not be applied directly to
+                # 'unique_pks'. It is better to apply it to indices.
+                sampled_indices = np.random.choice(np.arange(len(unique_pks)),
+                                                   max_partitions_contributed,
+                                                   replace=False)
+
+                sampled_elements = [unique_pks[i] for i in sampled_indices]
```
sushkoy, line 226: How about sampled_elements = unique_pks[sampled_indices]?

sushkoy (approved):

pipeline_dp/dp_engine.py
dvadym: using arrays for indexing works only for numpy arrays, unique_pks is a list and we don't want to convert it to numpy array because of types conversion :)

sushkoy: Sounds good, thanks for clarifying!

PR #163, opened 2022-01-22
Spark privacy id count

yuan2z:
## Description
Add Privacy ID Count Transform to Spark's public API.
- use f-string.

## How has this been tested?
Unit tested.


tests/private_spark_test.py
```diff
@@ -73,7 +73,7 @@ def privacy_id_extractor(x):
             high=5,
             budget_weight=1,
             public_partitions=None,
-            partition_extractor=lambda x: "pk" + str(x // 10),
+            partition_extractor=lambda x: f"pk:{x // 10}",
```
dvadym, line 76: Thanks for improving style!

dvadym (approved):
Thanks!

PR #169, opened 2022-01-24
Don't return privacy id count from CompoundCombiner

dvadym:


pipeline_dp/__init__.py
```diff
@@ -14,4 +14,3 @@
 from pipeline_dp.pipeline_backend import SparkRDDBackend
 
 __version__ = '0.0.1rc0'
-
```
miracvbasaran, line 17: Is this intended?

miracvbasaran (approved):

pipeline_dp/__init__.py
dvadym: Yeah, this is because of "make format" and for some reason that fixed linting on github workflow, which is great.

PR #172, opened 2022-01-25
Fix type error in beams tests on Windows.

XinyuYe:
## Description
There seems to be some bugs around long values in pyspark on older python versions. This PR replaces the numbers in the broken tests with float

```
E   RuntimeError: VarLong too long.
```


dvadym (approved):
Wow, thanks a lot of fixing this!

PR #175, opened 2022-01-25
Revamp the colab

Kulankhina:
Completely rewrote the colab:
- added DP description, example description, etc
- added Core API examples
- added Beam API examples


dvadym (approved):
Thanks a lot for creating such great colab!

PR #179, opened 2022-01-25
Update of readme and a local run example added

dvadym:


README.md
```diff
@@ -16,13 +16,14 @@ The project is in the early development stage. More description will be added la
 
 ## Getting started
 
-Take a look on examples of how to run PipelineDP Apache Spark or Beam:
+Take a look on examples of how to run PipelineDP:
 
 * [Apache Spark example](examples/movie_view_ratings_spark.py)
 * [Apache Beam example](examples/movie_view_ratings_beam.py)
-* [Local execution example](examples/movie_view_ratings.py)
+* [W/o frameworks example](examples/movie_view_ratings_local.py)
```
sushkoy, line 23: nit: "Framework-free example" or "Example without using any frameworks" (which is similar to the next point)

examples/movie_view_ratings_local.py
```diff
@@ -0,0 +1,76 @@
...
+
+FLAGS = flags.FLAGS
+flags.DEFINE_string('input_file', None, 'The file with the movie view data')
+flags.DEFINE_string('output_file', None, 'Output file')
+
+
+def main(unused_argv):
+    # Here, we use a local backend for computations. Which does not depend on
+    # any pipeline framework and it is implemented in pure Python in
+    # PipelineDP. It keeps all data in memory and not optimized for large data.
```
sushkoy, line 18: ..and [is] not optimised..

examples/movie_view_ratings_local.py
```diff
@@ -0,0 +1,76 @@
+""" Demo of PipelineDP with Apache Beam.
```
sushkoy, line 1: "locally, without any external data processing framework"

examples/movie_view_ratings_local.py
```diff
@@ -0,0 +1,76 @@
...
+FLAGS = flags.FLAGS
+flags.DEFINE_string('input_file', None, 'The file with the movie view data')
+flags.DEFINE_string('output_file', None, 'Output file')
+
+
+def main(unused_argv):
+    # Here, we use a local backend for computations. Which does not depend on
+    # any pipeline framework and it is implemented in pure Python in
+    # PipelineDP. It keeps all data in memory and not optimized for large data.
+    # For datasets till tens of megabytes it is faster than running Beam or Spark
```
sushkoy, line 19: For datasets smaller than ~tens of megabytes, local execution without any framework is faster than local mode with Beam or Spark.

examples/movie_view_ratings_local.py
```diff
@@ -0,0 +1,76 @@
...
+
+    # Create a computational graph for the aggregation.
+    # All computations are lazy. dp_result is iterable, but iterating it would
+    # fail until budget computed (below).
+    # It’s possible to call DPEngine.aggregate multiple times with different
+    # metrics to compute.
+    dp_result = dp_engine.aggregate(movie_views, params, data_extractors)
+
+    # Make budget computations.
+    budget_accountant.compute_budgets()
```
sushkoy, line 62: Reading the example makes me think (separately from this CL):

Maybe it'd be easier to have:
dp_engine.finalize() 
instead of
?

The dp_engine.finalize() could simply call budget_accountant.compute_budgets internally (or, if needed, do something more). For the developer, the important part is that dp_engine needs to be finalized before the computation can begin.

budget_accountant.compute_budgets might be too technical from the API perspective. Developers may not understand why such computation is needed, and they don't need to.

The typical flow would look like:
dp_engine.aggregate(..)
dp_engine.aggregate(..)
dp_engine.aggregate(..)
dp_engine.finalize()

Another optional idea on top of this (might need more discussion):
dp_engine.finalize() can actually initiate lazy computations. That's a bit more explicit and less surprising than initiating computations during iteration or converting to a list.

WDYT?

examples/movie_view_ratings_local.py
```diff
@@ -0,0 +1,76 @@
...
+
+    # Initiate computations.
```
sushkoy, line 64: It's a bit unclear why transforming to list initiates computations.

Maybe:
"Here's where the lazy iterator initiates computations and gets transformed into actual results."

sushkoy (approved):

examples/movie_view_ratings_local.py
```diff
@@ -0,0 +1,76 @@
...
+                                                          total_delta=1e-6)
+
+    # Load and parse input data
+    movie_views = parse_file(FLAGS.input_file)
+
+    # Create a DPEngine instance.
+    dp_engine = pipeline_dp.DPEngine(budget_accountant, backend)
+
+    params = pipeline_dp.AggregateParams(
+        noise_kind=pipeline_dp.NoiseKind.LAPLACE,
```
sushkoy, line 34: To make this simpler, can we use the default noise_kind?

examples/movie_view_ratings_local.py
dvadym: Thanks!

dvadym (commented):
Thanks for reivew!

examples/movie_view_ratings_local.py
```diff
@@ -0,0 +1,76 @@
...
+        metrics=[
+            # we can compute multiple metrics at once.
+            pipeline_dp.Metrics.COUNT,
+            pipeline_dp.Metrics.SUM,
+            pipeline_dp.Metrics.PRIVACY_ID_COUNT
+        ],
+        # Limit contributions of each user.
```
sushkoy, line 40: We could also ay something more specific, e.g.:

....
# Maximum movies rated per user
max_partitions_contributed=2,

# At most one rating per user per movie
max_contributions_per_partition=1,

# Minimal movie rating 
min_value = 1,

# Maximal movie rating 
max_value = 5)

PR #187, opened 2022-01-27
Fixes for utility analysis to include metrics_names.

XinyuYe:
## Description
Fixes for utility analysis for namedtuple change of the metrics in https://github.com/OpenMined/PipelineDP/commit/8f730292f45628c46949eff16d190365a5e3d42b. Because the metrics of the utility analysis are not directly exposed to customers (they are used in visualization). We don't change compute_metrics for now.

### DIFF
--- a/utility_analysis/non_private_combiners.py
+++ b/utility_analysis/non_private_combiners.py
@@ -5,7 +5,7 @@ noises and computation.
 """
 import abc
 import copy
-from typing import Iterable, Sized, Tuple
+from typing import Iterable, Sized, Tuple, List
 
 import pipeline_dp
 import numpy as np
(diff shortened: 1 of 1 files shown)


utility_analysis/non_private_combiners.py
```diff
@@ -109,6 +121,14 @@ class CompoundCombiner(pipeline_dp.combiners.Combiner):
 
     def __init__(self, combiners: Iterable['Combiner']):
         self._combiners = combiners
+        self._metrics_to_compute = []
```
dvadym, line 124: Can we use here CompoundCombiner from combiners.py?

dvadym (approved):
Thanks! Changes itself LGTM. I left a question (which is independent of this CL)

PR #204, opened 2022-01-27
Fixes accumulator tests.

XinyuYe:
## Description
Quick fixes accumulator tests. Accumulators shouldn't matter too much in our code base going forward. Probably we want to remove them altogether later.


dvadym (approved):
Thank you!

PR #213, opened 2022-01-28
Adds min_sum_per_partition and max_sum_per_partition

XinyuYe:
## Description
Adds min_sum_per_partition and max_sum_per_partition. These two parameters only works for sum. 
Only 1 pair of (min_value, max_value) and (min_sum_per_partition, max_sum_per_partition) might set
if (min_sum_per_partition, max_sum_per_partition) is set Metrics should be equal to [SUM]


dvadym:
It was already [implemented](https://github.com/OpenMined/PipelineDP/blob/4540083dfbab6377b28c3912dfc5ef70a8141075/pipeline_dp/aggregate_params.py#L103) on different PRs

PR #241, opened 2022-02-03
Private Beam: CombineFn API for making custom DP aggregation

dvadym:


pipeline_dp/private_beam.py
```diff
@@ -308,3 +315,117 @@ def __init__(self, fn: Callable, label: Optional[str] = None):
     def expand(self, pcol: pvalue.PCollection):
         return pcol | "flatten values" >> beam.ParDo(
             FlatMap._FlattenValues(map_fn=self._fn))
+
+
+class PrivateCombineFn(beam.CombineFn):
+
+    @abc.abstractmethod
+    def add_private_input(self, accumulator, input):
```
sushkoy, line 323: As this is developer-facing, could we add some docs: what is this method supposed to do?

add_private_input implies that the input variable should already be protected by some privacy, or should this be just "add_input"?

pipeline_dp/private_beam.py
```diff
@@ -308,3 +315,117 @@ def __init__(self, fn: Callable, label: Optional[str] = None):
...
+        pass
+
+    @abc.abstractmethod
+    def extract_private_output(self, accumulator):
```
sushkoy, line 327: ditto, could we add some some instructions for developers who will be implementing this method?

pipeline_dp/private_beam.py
```diff
@@ -308,3 +315,117 @@ def __init__(self, fn: Callable, label: Optional[str] = None):
...
+        live in the driver process.
+        """
+        pass
+
+    @abc.abstractmethod
+    def set_aggregate_params(self,
+                             aggregate_params: pipeline_dp.AggregateParams):
+        """Sets aggregate parameters
+
+        The custom combiner can optionally use it for own DP parameter
```
sushkoy, line 351: Just a few clarifications:

1) I'm wondering about "optionally". Can we think of a case where data in AggregateParams (e.g. sensitivity params) will *not* be used?

2) Why aren't we just storing aggregate_params here instead of letting clients to override this method? Is that because the clients may want to store only *some* parameters? In this case we could possibly clarify:

"Use this to store the parameters that are needed to ensure differential privacy for your custom combiner (e.g. sensitivity parameters)"

pipeline_dp/private_beam.py
```diff
@@ -308,3 +315,117 @@ def __init__(self, fn: Callable, label: Optional[str] = None):
     def expand(self, pcol: pvalue.PCollection):
         return pcol | "flatten values" >> beam.ParDo(
             FlatMap._FlattenValues(map_fn=self._fn))
+
+
+class PrivateCombineFn(beam.CombineFn):
```
sushkoy, line 320: IIUC, the developers implementing this method will need to understand differential privacy quite a lot, and will need to do things like clamping and noising. Maybe we can add some documentation to make it clear that this is an advanced feature that can break differential privacy guarantees if not implemented correctly, wdyt? 

Related: what is the distribution of responsibilities between the custom CombineFn and the rest of PipelineDp? I see that the custom code needs to take care of clamping and noising. What are the guarantees enforced by the PipelineDP, does it do cross-partition contribution bounding and partition selection?

(Also a few related questions below)

examples/experimental/beam_combine_fn.py
```diff
@@ -0,0 +1,120 @@
...
+class DPSumCombineFn(pbeam.PrivateCombineFn):
+
+    def __init__(self, min_value, max_value):
+        self._min_value = min_value
+        self._max_value = max_value
+
+    def create_accumulator(self):
+        return 0
+
+    def add_private_input(self, accumulator, input):
```
sushkoy, line 45: Should this now be "add_input_for_private_output"?

examples/experimental/beam_combine_fn.py
```diff
@@ -0,0 +1,120 @@
...
+    def extract_private_output(self, accumulator):
+        # Simple implementation of Laplace mechanism.
+        max_abs_value = np.maximum(np.abs(self._min_value),
+                                   np.abs(self._max_value))
+        sensitivity = self._aggregate_params.max_contributions_per_partition * \
+                      self._aggregate_params.max_partitions_contributed * max_abs_value
+        eps = self._budget.eps
+        laplace_b = sensitivity / eps
+
+        # Warning: using a standard laplace noise is done only for simplicity, don't use it in production.
```
sushkoy, line 60: nit: "This is a demo, so for simplicity we use unsafe Laplace noise, which isn't hardened against the floating point attacks. In real implementations, you should use safe noise generators (e.g. ...provide links...)"

On a second thought: if we use this as a demo, maybe we should provide an example that's using a safe noise, to show that custom implementations need to be aware of this complexity, and provide examples of how to use Laplace noise (from the pydp library).

examples/experimental/beam_combine_fn.py
```diff
@@ -0,0 +1,120 @@
...
+        # Better it's to use a standard PipelineDP metric Sum.
+        return np.random.laplace(accumulator, laplace_b)
+
+    def request_budget(self, budget_accountant):
+        self._budget = budget_accountant.request_budget(
```
sushkoy, line 65: Optional: to make this more fool proof and put less burden on developers, we could have this method return the budget, and have the upstream code store this in a variable and pass this to extract_private_output as an argument.

```
def request_budget(self, budget_accountant):
        return budget_accountant.request_budget(..)

def extract_private_output(self, accumulator, budget)
   ...
```

pipeline_dp/private_beam.py
```diff
@@ -308,3 +315,146 @@ def __init__(self, fn: Callable, label: Optional[str] = None):
...
+    def request_budget(self,
+                       budget_accountant: budget_accounting.BudgetAccountant):
+        self._private_combine_fn.request_budget(budget_accountant)
+
+    def set_aggregate_params(self, aggregate_params):
+        self._private_combine_fn.set_aggregate_params(aggregate_params)
+
+
+@dataclasses.dataclass
+class CombinePerKeyParams:
```
sushkoy, line 417: nit: Add docs for the class and fields, similar to other classes, e.g.: https://github.com/OpenMined/PipelineDP/blob/583aeed06dfe5cc8d49bf4fc74552cf41628cfbc/pipeline_dp/aggregate_params.py#L227

pipeline_dp/private_beam.py
```diff
@@ -308,3 +315,146 @@ def __init__(self, fn: Callable, label: Optional[str] = None):
...
+        combiner = _CombineFnCombiner(self._combine_fn)
+        aggregate_params = pipeline_dp.AggregateParams(
+            metrics=None,
+            max_partitions_contributed=self._params.max_partitions_contributed,
+            max_contributions_per_partition=self._params.
+            max_contributions_per_partition,
+            custom_combiners=[combiner])
+
+        backend, dp_engine = self._create_dp_engine()
+        # Assumed elements format: (privacy_id, (partition_key, value))
```
sushkoy, line 445: Should we document that input/output assumptions in the top-level doc for CombinePerKey?

dvadym (commented):
Thanks for comments! I've addressed them

PR #266, opened 2022-04-20
Computing Epsilon/Delta for BudgetScope (obsolete)

dvadym:


tests/budget_accounting_test.py
```diff
@@ -143,6 +143,9 @@ def test_request_after_compute_raise_exception(self):
             budget_accountant.request_budget(
                 mechanism_type=MechanismType.LAPLACE)
 
+    def test_n_aggregations(self):
```
miracvbasaran: Did you mean to add a test here?

pipeline_dp/budget_accounting.py
```diff
@@ -128,23 +134,35 @@ def request_budget(
...
+            aggregation_scope: if True, this is a high-level scope which
+             corresponds to one aggregation.
 
         Returns:
-            the scope that should be used in a "with" block enclosing the operations consuming the budget.
+            the scope that should be used in a "with" block enclosing the
+            operations consuming the budget.
         """
-        return BudgetAccountantScope(self, weight)
+        aggregation_index = None
```
miracvbasaran, line 162: I am not sure if I perfectly understand this aggregation_scope parameter. Why do we have separate logic when it is set and when it is not?

pipeline_dp/budget_accounting.py
```diff
@@ -196,11 +247,29 @@ def _normalise_mechanism_weights(self):
         for mechanism in self.mechanisms:
             mechanism.weight *= normalisation_factor
 
+    def _compute_budget_for_aggregation_scope(self):
+        n_aggregations = self.accountant._n_aggregations
+        weights = self.accountant._aggregations_weights
+        if n_aggregations is not None:
+            budget_ratio = 1 / n_aggregations
+        elif weights is not None:
+            sum_weights = sum(weights)
```
miracvbasaran, line 255: Are you still able to provide weights when calling `aggregate()`? If so, what happens when that weight is different from the original weight provided to `BudgetAccountant`?

pipeline_dp/budget_accounting.py
```diff
@@ -128,23 +134,35 @@ def request_budget(
     def compute_budgets(self):
         pass
 
-    def scope(self, weight: float):
-        """Defines a scope for DP operations that should consume no more than "weight" proportion of the budget
-        of the parent scope.
+    def scope(self,
+              weight: float,
+              aggregation_scope=False) -> 'BudgetAccountantScope':
```
chinmayshah99, line 139: nit: add type hint `bool` for this

pipeline_dp/budget_accounting.py
```diff
@@ -169,24 +187,56 @@ def _finalize(self):
...
+                 accountant: BudgetAccountant,
+                 weight: float,
+                 index_aggregation_scope: Optional[int] = None):
         self.weight = weight
         self.accountant = accountant
+        self._index_aggregation_scope = index_aggregation_scope
+        self._epsilon = None
+        self._delta = None
+        if index_aggregation_scope is not None:
+            self._compute_budget_for_aggregation_scope()
```
chinmayshah99, line 208: do we not set `self._index_aggregation_scope` after the computation? 
This is referenced in `_validate_epsilon_delta()`

pipeline_dp/budget_accounting.py
```diff
@@ -196,11 +246,29 @@ def _normalise_mechanism_weights(self):
         for mechanism in self.mechanisms:
             mechanism.weight *= normalisation_factor
 
+    def _compute_budget_for_aggregation_scope(self):
+        n_aggregations = self.accountant._n_aggregations
+        weights = self.accountant._aggregations_weights
+        if n_aggregations is not None:
+            budget_ratio = 1 / n_aggregations
```
chinmayshah99, line 253: need to add a check for if `_aggregations` is 0

chinmayshah99 (changes requested):

pipeline_dp/budget_accounting.py
```diff
@@ -128,23 +134,35 @@ def request_budget(
...
+        The accountant will automatically scale the budgets of all
+        sub-operations accordingly.
 
         Example usage:
           with accountant.scope(weight = 0.5):
              ... some code that consumes DP budget ...
 
         Args:
-            weight: budget weight of all operations made within this scope as compared to.
+            weight: budget weight of all operations made within this scope as
```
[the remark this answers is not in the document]
dvadym: Updated, please check

pipeline_dp/budget_accounting.py
dvadym: There were no yet comments about aggregation_scope, please check now

pipeline_dp/budget_accounting.py
dvadym: Now, it set 3 lines above

pipeline_dp/budget_accounting.py
dvadym: It's checked in the contrustor, that it should be > 0

pipeline_dp/budget_accounting.py
dvadym: yes, now (previously PR wasn't ready yet) they are checked `BudgetAccount.scope()`, that those weights should be the same

dvadym (commented):
PTAL

pipeline_dp/budget_accounting.py
miracvbasaran: Looks much better, thanks!

pipeline_dp/budget_accounting.py
```diff
@@ -151,17 +158,49 @@ def scope(self,
...
-        if aggregation_scope:
+        if aggregation_scope and (self._n_aggregations or
+                                  self._aggregation_weights):
+            if self._inside_aggregation_scope:
+                raise ValueError("Aggregation scopes can not be nested.")
             aggregation_index = self._get_next_aggregation_index()
+            if self._n_aggregations:
+                if aggregation_index >= self._n_aggregations:
+                    raise ValueError("Exceeded the number of allowed "
+                                     "aggregations. If you need more update "
```
miracvbasaran, line 181: supernit: comma(,) after "more"

pipeline_dp/budget_accounting.py
```diff
@@ -151,17 +158,49 @@ def scope(self,
...
+                                     "BudgetAccountant")
+                if weight != 1:
+                    raise ValueError(f"When 'n_aggregations' is set in the "
+                                     f"constructor of BudgetAccountant, all "
+                                     f"aggregation weights have to be 1, but "
+                                     f"weight={weight}.")
+            elif self._aggregation_weights != None:
+                if aggregation_index >= len(self._aggregation_weights):
+                    raise ValueError("Exceeded the number of allowed "
+                                     "aggregations. If you need more update "
```
miracvbasaran, line 192: supernit: comma(,) after "more"

pipeline_dp/budget_accounting.py
```diff
@@ -151,17 +158,49 @@ def scope(self,
...
+                                     "of BudgetAccountant")
+                expected_weight = self._aggregation_weights[aggregation_index]
+                if weight != expected_weight:
+                    raise ValueError(
+                        f"The provided weight for the aggregation "
+                        f"with index={aggregation_index} is {weight}, "
+                        f"but 'aggregation_weights' in the constructor "
+                        "of BudgetAccountant contains {expected_weight}")
+            else:
+                assert False, "It should not happen."
```
miracvbasaran, line 203: Is it idiomatic to throw an error like this instead of `raise XError("It should not happen.")` in Python?

pipeline_dp/budget_accounting.py
```diff
@@ -151,17 +158,49 @@ def scope(self,
...
+                                     f"aggregation weights have to be 1, but "
+                                     f"weight={weight}.")
+            elif self._aggregation_weights != None:
+                if aggregation_index >= len(self._aggregation_weights):
+                    raise ValueError("Exceeded the number of allowed "
+                                     "aggregations. If you need more update "
+                                     "'aggregation_weights' in the constructor "
+                                     "of BudgetAccountant")
+                expected_weight = self._aggregation_weights[aggregation_index]
+                if weight != expected_weight:
```
miracvbasaran, line 196: If they already specify `expected_weight`, maybe they shouldn't need to specity weight here? It feels not great from a usability perspective to ask for the same information twice?

Although I could see it being useful for making it less error-prone.

tests/budget_accounting_test.py
```diff
@@ -160,6 +160,33 @@ def test_n_aggregations(self, n_aggregations):
 
         budget_accountant.compute_budgets()
 
+    def test_aggregation_weights(self):
```
miracvbasaran, line 163: Do we also test automatic budget splitting, i.e. equal weights when weights are not specified, somewhere?

miracvbasaran (changes requested):

tests/budget_accounting_test.py
dvadym: `weights` in BudgetAccountant is not optional parameter, it has to be specified

dvadym (commented):

dvadym:
Refined version of this feature was submitted on [PR](https://github.com/OpenMined/PipelineDP/pull/269)

PR #274, opened 2022-05-11
Support for DP without identifiers. Experimental, untested code.

sushkoy:


pipeline_dp/dp_engine.py
```diff
@@ -338,12 +361,18 @@ def _select_private_partitions_internal(self, col,
...
                        combiners.CompoundCombiner.AccumulatorType]) -> bool:
             """Lazily creates a partition selection strategy and uses it to determine which
             partitions to keep."""
-            privacy_id_count, _ = row[1]
+            row_count, _ = row[1]
+
+            # A conservative (lower) estimate of how many privacy IDs contributed to
+            # this partition. This estimate is only needed when privacy IDs are not available
+            # in the original dataset.
+            privacy_id_count = row_count // max_privacy_ids_per_row
```
dvadym, line 374: `//` rounds down, but we can round up, example

`max_privacy_ids_per_row = 2`, `row_count = 3`, then there is at least 2 privacy ids.

There is a trick for making round up for integers:
`(a + b -1) // b`.
Probably better to extract it to a separate method.

dvadym (commented):
It looks very cool.

pipeline_dp/dp_engine.py
```diff
@@ -122,9 +134,19 @@ def _aggregate(self, col, params: pipeline_dp.AggregateParams,
             col, combiner, "Reduce accumulators per partition key")
         # col : (partition_key, accumulator)
 
+
         if public_partitions is None:
+            max_privacy_ids_per_row = 1
```
dvadym, line 139: Isn't `max_rows_per_privacy_id`?

pipeline_dp/combiners.py
```diff
@@ -502,6 +504,9 @@ def create_compound_combiner(
             combiners.append(
                 SumCombiner(CombinerParams(budget_sum, aggregate_params)))
     if pipeline_dp.Metrics.PRIVACY_ID_COUNT in aggregate_params.metrics:
+        if aggregate_params.contribution_bounds_already_enforced:
```
dvadym, line 507: It's better to fail earlier, it look like here a better place is `AggregateParams` validation [here](https://github.com/OpenMined/PipelineDP/blob/3a7b87a5d943ea6b083c45934c8bb1110cbe5766/pipeline_dp/aggregate_params.py#L86)

pipeline_dp/combiners.py
```diff
@@ -502,6 +504,9 @@ def create_compound_combiner(
...
+            raise 'Cannot calculate PRIVACY_ID_COUNT when ' \
```
dvadym, line 508: Nit: raise `ValueError('...')`

dvadym (approved):
Thank you!

PR #290, opened 2022-06-02
Improving ggregate_params validation

dvadym:


pipeline_dp/aggregate_params.py
```diff
@@ -160,23 +150,25 @@ def __post_init__(self):
...
-                 self.max_contributions_per_partition <= 0)):
+            if (self.max_partitions_contributed is None or
+                    self.max_contributions_per_partition is None):
                 raise ValueError(
-                    "One amongst params.max_contributions or "
-                    "(params.max_partitions_contributed or "
-                    "params.max_contributions_per_partition) must be set to a "
-                    "positive value.")
+                    "Either params.max_contributions must be set or "
+                    "(params.max_partitions_contributed and "
```
chinmayshah99, line 166: It would be nice if we got a separate error for these cases. ex. a user may just set `max_partitions_contributed` and not `max_contributions_per_partition`. Maybe raise a separate error?

dvadym: Thanks, good point. Done

pipeline_dp/aggregate_params.py
```diff
@@ -160,23 +150,31 @@ def __post_init__(self):
...
-                    "Only one in params.max_contributions or "
-                    "(params.max_partitions_contributed and "
-                    "params.max_contributions_per_partition) must be set")
-        else:
-            if ((self.max_partitions_contributed is None or
-                 self.max_partitions_contributed <= 0) and
-                (self.max_contributions_per_partition is None or
-                 self.max_contributions_per_partition <= 0)):
+                    "AggregateParams: only one in max_contributions or "
+                    "(max_partitions_contributed and "
```
chinmayshah99, line 159: nit: 
`Both (max_partitions_contributed and "max_contributions_per_partition) must be set"`

pipeline_dp/aggregate_params.py
```diff
@@ -160,23 +150,31 @@ def __post_init__(self):
...
+                    "max_contributions_per_partition) must be set")
+        else:  # self.max_contributions is None
+            n_not_none = _count_not_none(self.max_partitions_contributed,
+                                         self.max_contributions_per_partition)
+            if n_not_none == 0:
+                raise ValueError(
+                    "AggregateParams: either max_contributions must be set or "
+                    "(max_partitions_contributed and "
```
chinmayshah99, line 167: nit:
`Both (max_partitions_contributed and "max_contributions_per_partition) must be set"`

chinmayshah99 (changes requested):

pipeline_dp/aggregate_params.py
dvadym: Thanks,done

pipeline_dp/aggregate_params.py
dvadym: thanks, done

PR #296, opened 2022-06-14
Multi aggregate

dvadym:


pipeline_dp/private_beam.py
```diff
@@ -534,3 +535,96 @@ def expand(self, pcol: pvalue.PCollection):
...
+    if named_tuple is None:
+        named_tuple = collections.namedtuple(type_name, field_names)
+        named_tuple.__reduce__ = lambda self: (_create_named_tuple_instance,
+                                               (type_name, field_names,
+                                                tuple(self)))
+        _agg_named_tuple_cache[cache_key] = named_tuple
+    return named_tuple
+
+
+def _create_named_tuple_instance(type_name: str, field_names: tuple, values):
```
chinmayshah99, line 562: return Type is missing

chinmayshah99 (changes requested):

pipeline_dp/private_beam.py
dvadym: actually it's tricky to specify type here, since this function dynamically creates named_tuple :)

This PR is WIP and it will change much :)

Issue #287, opened 2022-06-01
Improve numerical stabililty of DP mean and variance computations


dvadym:
# Context
Let we have a set of real numbers `{x_1, ... x_n}`, such that `min_value <= x_i <= max_value` (`min_value` and `max_value` are known),  and each user can contribute only 1 data point.

A simple way to compute DP mean is `(sum+noise)/(count+noise)`. The sensitivity of the `sum` then is `max(|min_value|, |max_value|)`.
There is a normalization trick, to decrease the amount of noise in the nominator, namely
`normalized_sum = (x_1-mid)+...+(x_n-mid)=(x_1+..x_n)-mid*n, mid = (min_value+max_value)/2`.
and
`dp_mean = (normalized_sum+noise)/(count+noise) + mid`.
We call `x_i-mid` normalization.

The win here, is because `sensitivity(normalized_sum) = (max_value-min_value) <=max(|min_value|, |max_value|)`.

[Here](https://github.com/OpenMined/PipelineDP/blob/c9ed8157b73d9f9f86baead2487c9737f334e88a/pipeline_dp/dp_computations.py#L320) is the implementation of this in PipelineDP.

## What can be done better 
We can normalize elements, when we do summation,  now `sum` is normalized.

**Explanation:** When `max_value - min_value << max_value`, the elements might be large, while normalized elements are small, so `normalized_sum` is small, but when `sum` is computed, the precision is lost because of rounding.

# Goals
To implement normalization of each element, which will improve numerical stability.

It will consist of 2 parts:
1. Change computing mean aggregation in [MeanCombiner](https://github.com/OpenMined/PipelineDP/blob/c9ed8157b73d9f9f86baead2487c9737f334e88a/pipeline_dp/combiners.py#L253).
2. Change [_compute_mean](https://github.com/OpenMined/PipelineDP/blob/c9ed8157b73d9f9f86baead2487c9737f334e88a/pipeline_dp/dp_computations.py#L287).

**Note**: the next step is to improve variance

PR #301, opened 2022-07-10
Improvement: numerical stability of DP mean and variance computations as per #287

OmarElbaroudy:
# Description
This PR introduces an improvement for the numerical stability of DP mean and variance computations as per #287.
This is done by computing the normalized sum in the accumulator of the mean combiner.  
For computing variance the sum and the squares sum are both normalized in the accumulator of the variance combiner.

## Affected Dependencies
N/A

## How has this been tested?
- Existing tests for the mean and variance combiners where refactored to test the normalized sum instead of sum
- Existing tests in `dp_computations` were refactored to call `compute_dp_mean` and `compute_dp_var` methods correctly with normalized sum instead of total sum

## Checklist

### DIFF
--- a/pipeline_dp/combiners.py
+++ b/pipeline_dp/combiners.py
@@ -253,8 +253,8 @@ class SumCombiner(Combiner):
 class MeanCombiner(Combiner):
     """Combiner for computing DP Mean. Also returns sum and count in addition to
     the mean.
-    The type of the accumulator is a tuple(count: int, sum: float) that holds
-    the count and sum of elements in the dataset for which this accumulator is
+    The type of the accumulator is a tuple(count: int, normalized_sum: float) that holds
+    the count and normalized sum of elements in the dataset for which this accumulator is
     computed.
     """
     AccumulatorType = Tuple[int, float]
@@ -274,9 +274,12 @@ class MeanCombiner(Combiner):
         self._metrics_to_compute = metrics_to_compute
 
     def create_accumulator(self, values: Iterable[float]) -> AccumulatorType:
-        return len(values), np.clip(
-            values, self._params.aggregate_params.min_value,
-            self._params.aggregate_params.max_value).sum()
+        min_value = self._params.aggregate_params.min_value
+        max_value = self._params.aggregate_params.max_value
+        middle = dp_computations.compute_middle(min_value, max_value)
+        normalized_values = np.clip(values, min_value, max_value) - middle
+
+        return len(values), normalized_values.sum()
 
     def merge_accumulators(self, accum1: AccumulatorType,
                            accum2: AccumulatorType):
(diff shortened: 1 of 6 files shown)


pipeline_dp/combiners.py
```diff
@@ -324,10 +328,20 @@ def __init__(self, params: CombinerParams,
...
+
+        middle = dp_computations.compute_middle(min_value, max_value)
+        squares_middle = dp_computations.compute_middle(squares_min_value, squares_max_value)
         clipped_values = np.clip(values,
                                  self._params.aggregate_params.min_value,
                                  self._params.aggregate_params.max_value)
-        return len(values), clipped_values.sum(), (clipped_values**2).sum()
+
+        normalized_values = clipped_values - middle
+        normalized_squared_values = (clipped_values**2) - squares_middle
```
dvadym, line 342: Variance of `x_1, ...x_n` does not change when the same number is added to all `x_i`, so variance is not changed with applying normalization. So we can do normalization as `normalized_squared_values = normalized_values**2` and as a result the values will be smaller -> so less rounding, better accuracy. WDYT?

dvadym (commented):
Thanks, it looks great! I left some comments.

Could you run `make format` from sources root in order to make formatting

pipeline_dp/combiners.py
```diff
@@ -274,9 +274,13 @@ def __init__(self, params: CombinerParams,
         self._metrics_to_compute = metrics_to_compute
 
     def create_accumulator(self, values: Iterable[float]) -> AccumulatorType:
-        return len(values), np.clip(
+        middle = dp_computations.compute_middle(self._params.aggregate_params.min_value,
```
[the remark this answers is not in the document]
OmarElbaroudy: Done.

pipeline_dp/combiners.py
OmarElbaroudy: Good point. Done.

pipeline_dp/dp_computations.py
```diff
@@ -285,9 +286,8 @@ def compute_dp_sum(sum: float, dp_params: MeanVarParams):
 
 
 def _compute_mean(
-    count: float,
     dp_count: float,
-    sum: float,
+    normalized_sum: float,
```
dvadym, line 290: We changed behaviour of this function and expectations on the arguments. Let's reflect that in function signature/comments.

1. Maybe something like `_compute_mean_for_normalized_sum` (and to name the argument `sum`)?
2. After normalization all elements are in the symmetric interval, maybe to use one arg `max_abs_value` instead of `min_value/max_value`

dvadym (commented):
It looks great, I left 1 comment

dvadym (approved):
Thanks!

PR #302, opened 2022-07-21
Contribution bounding refactoring

dvadym:
This PR extracts Contribution bounding to `ContributionBounder` class. There is refactoring, no changes in functionality.

### DIFF
--- a/pipeline_dp/contribution_bounders.py
+++ b/pipeline_dp/contribution_bounders.py
@@ -82,13 +82,12 @@ class SamplingCrossAndPerPartitionContributionBounder(ContributionBounder):
             f"{max_partitions_contributed}) partitions")
 
         # (privacy_id, [(partition_key, accumulator)])
-        def unnest_cross_partition_bound_sampled_per_key(pid_pk_v):
+        def rekey_by_privacy_id_and_unnest(pid_pk_v):
             pid, pk_values = pid_pk_v
             return (((pid, pk), v) for (pk, v) in pk_values)
 
-        return backend.flat_map(col,
-                                unnest_cross_partition_bound_sampled_per_key,
-                                "Unnest")
+        return backend.flat_map(col, rekey_by_privacy_id_and_unnest,
+                                "Rekey by privacy id and unnest")
 
 
 class SamplingPerPrivacyIdContributionBounder(ContributionBounder):
--- a/pipeline_dp/contribution_bounders.py
+++ b/pipeline_dp/contribution_bounders.py
@@ -77,8 +77,8 @@ class SamplingCrossAndPerPartitionContributionBounder(ContributionBounder):
                                            "Sample per privacy_id")
 
         report_generator.add_stage(
-            f"Cross-partition contribution bounding: for eacy privacy id "
-            f" randomly select max(actual_partition_contributed, "
+            f"Cross-partition contribution bounding: for each privacy id "
+            f"randomly select max(actual_partition_contributed, "
             f"{max_partitions_contributed}) partitions")
 
         # (privacy_id, [(partition_key, accumulator)])
@@ -97,10 +97,6 @@ class SamplingPerPrivacyIdContributionBounder(ContributionBounder):
     If a privacy_id contributes more than max_contributions, then
     max_contributions contributions are uniformly sampled, otherwise all
     contributions are kept.
-
-    Warning: This contribution bounder does group_by by privacy_id and processes
-    each privacy_id's data on one worker. In case if there are some privacy_ids
-    which contributes too much, it might lead to stale worker jobs.
     """
 
     def bound_contributions(self, col, params, backend, report_generator,
--- a/pipeline_dp/contribution_bounders.py
+++ b/pipeline_dp/contribution_bounders.py
@@ -1,0 +1,145 @@
+import abc
+import collections
+from typing import Callable
+
+import pipeline_dp
+from pipeline_dp import pipeline_backend
+
+
+class ContributionBounder(abc.ABC):
+    """Interface for objects which perform contribution bounding."""
+
+    @abc.abstractmethod
+    def bound_contributions(self, col, params: pipeline_dp.AggregateParams,
+                            backend: pipeline_backend.PipelineBackend,
+                            aggregate_fn: Callable):
+        """Bound contributions of privacy id.
+
+        Contribution bounding is performed to ensure that sensitivity of the
+        query to anonymize is limited. There are many ways how to perform
+        contribution bounding. Sub-classes implement specific contribution
+        strategies.
+
+        Also this function performs the aggregation step per
+        (privacy_id, partition_key) by calling aggregate_fn on data points
+        corresponding to each (privacy_id, partition_key).
+
+        Args:
+          col: collection, with types of each element: (privacy_id,
+            partition_key, value).
+          params: contains parameters needed for contribution bounding..
+          backend: pipeline backend for performing operations on collections.
+          aggregate_fn: function that takes a list of values and returns an
+            aggregator object.
+        Returns:
+          collection with elements ((privacy_id, partition_key),
+              accumulator).
+
+        """
+
+
+class SamplingCrossAndPerPartitionContributionBounder(ContributionBounder):
+    """Bounds the contribution by privacy_id per and cross partitions.
+
+    It ensures by performing sampling if needed that each privacy_id contributes
+    not more than to max_partitions_contributed partitions (cross-partition
+    contribution bounding) and not more than max_contributions_per_partition to
+    each contributed partition (per partition contribution bounding).
+    """
+
+    def bound_contributions(self, col, params, backend, report_generator,
+                            aggregate_fn):
+        """See docstrings for this class and for the base class bound_contributions."""
+        max_partitions_contributed = params.max_partitions_contributed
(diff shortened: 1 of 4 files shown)


pipeline_dp/contribution_bounders.py
```diff
@@ -0,0 +1,145 @@
...
+        col = backend.sample_fixed_per_key(col, max_contributions,
+                                           "Sample per privacy_id")
+        report_generator.add_stage(
+            f"User contributions bounding: randomly selected not "
+            f"more than {max_contributions} contributions")
+
+        # (privacy_id, [(partition_key, value)])
+
+        # Convert the per privacy id list into a dict with key as partition_key
+        # and value to be list of input values.
```
miracvbasaran, line 119: nit: 'and values as the list of input values'

pipeline_dp/contribution_bounders.py
```diff
@@ -0,0 +1,145 @@
...
+        col = backend.map_tuple(
+            col, lambda pid, pk, v: (pid, (pk, v)),
+            "Rekey to ((privacy_id), (partition_key, value))")
+        col = backend.sample_fixed_per_key(col, max_contributions,
+                                           "Sample per privacy_id")
+        report_generator.add_stage(
+            f"User contributions bounding: randomly selected not "
+            f"more than {max_contributions} contributions")
+
+        # (privacy_id, [(partition_key, value)])
```
miracvbasaran, line 116: nit: I would delete the newline before this line (and below on line 130)

pipeline_dp/contribution_bounders.py
```diff
@@ -0,0 +1,145 @@
...
+    """
+
+    def bound_contributions(self, col, params, backend, report_generator,
+                            aggregate_fn):
+        """See docstrings for this class and for the base class bound_contributions."""
+        max_contributions = params.max_contributions
+        col = backend.map_tuple(
+            col, lambda pid, pk, v: (pid, (pk, v)),
+            "Rekey to ((privacy_id), (partition_key, value))")
+        col = backend.sample_fixed_per_key(col, max_contributions,
```
miracvbasaran, line 110: This wouldn't work for Sums if we want to do a per-partition/per-user sum before doing contribution bounding like we do in go, right?

pipeline_dp/contribution_bounders.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+class SamplingPerPrivacyIdContributionBounder(ContributionBounder):
+    """Bounds the total contributions of a privacy_id.
+
+    If a privacy_id contributes more than max_contributions, then
+    max_contributions contributions are uniformly sampled, otherwise all
+    contributions are kept.
+
+    Warning: This contribution bounder does group_by by privacy_id and processes
```
miracvbasaran, line 98: I am not sure if that'll be a problem actually. If runtime of backend.sample_fixed_per_key does not scale by number of elements per key (and it would be possible to implement it as such), then after that step we only have max_contributions amount of data per privacy id, which should be fairly low.

I thinkg processing of all data per privacy_id is required if we want to do a scaling-based sampling (i.e. found total number of contributions, scale each contribution by that amount).

miracvbasaran (changes requested):
Thanks, looks good overall!

pipeline_dp/contribution_bounders.py
```diff
@@ -0,0 +1,145 @@
...
+        Also this function performs the aggregation step per
+        (privacy_id, partition_key) by calling aggregate_fn on data points
+        corresponding to each (privacy_id, partition_key).
+
+        Args:
+          col: collection, with types of each element: (privacy_id,
+            partition_key, value).
+          params: contains parameters needed for contribution bounding..
+          backend: pipeline backend for performing operations on collections.
+          aggregate_fn: function that takes a list of values and returns an
```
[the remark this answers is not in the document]
dvadym: There are multiple reasons:
1. Sometimes we can't separate bound contributions from aggregation, for example for Sum with `max_sum_per_partition`
2. It's better for performance to start aggregation during contribution bounding - anyway we group by `(privacy_id, partition_key)`, so we have on one worker all contributions from `(privacy_id, partition_key)`

In order to separate contribution bounding from aggregation, aggregation is injected as a function, so contribution bounding and aggregation is loosely coupled.

pipeline_dp/contribution_bounders.py
```diff
@@ -0,0 +1,145 @@
...
+        query to anonymize is limited. There are many ways how to perform
+        contribution bounding. Sub-classes implement specific contribution
+        strategies.
+
+        Also this function performs the aggregation step per
+        (privacy_id, partition_key) by calling aggregate_fn on data points
+        corresponding to each (privacy_id, partition_key).
+
+        Args:
+          col: collection, with types of each element: (privacy_id,
```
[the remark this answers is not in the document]
dvadym: For simplicity of the code it has to have. Since `aggregate()` can process any metrics, it's attempted to minimize divergence. In case of count and if the user doesn't specify `value_extractor`, value will be `None`.

pipeline_dp/contribution_bounders.py
```diff
@@ -0,0 +1,145 @@
...
+          collection with elements ((privacy_id, partition_key),
+              accumulator).
+
+        """
+
+
+class SamplingCrossAndPerPartitionContributionBounder(ContributionBounder):
+    """Bounds the contribution by privacy_id per and cross partitions.
+
+    It ensures by performing sampling if needed that each privacy_id contributes
```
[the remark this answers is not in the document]
dvadym: thanks!

pipeline_dp/contribution_bounders.py
```diff
@@ -0,0 +1,145 @@
...
+    not more than to max_partitions_contributed partitions (cross-partition
+    contribution bounding) and not more than max_contributions_per_partition to
+    each contributed partition (per partition contribution bounding).
+    """
+
+    def bound_contributions(self, col, params, backend, report_generator,
+                            aggregate_fn):
+        """See docstrings for this class and for the base class bound_contributions."""
```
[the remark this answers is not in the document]
dvadym: Thanks, I've used 2nd one.

pipeline_dp/contribution_bounders.py
```diff
@@ -0,0 +1,145 @@
...
+        max_partitions_contributed = params.max_partitions_contributed
+        max_contributions_per_partition = params.max_contributions_per_partition
+        col = backend.map_tuple(
+            col, lambda pid, pk, v: ((pid, pk), v),
+            "Rekey to ( (privacy_id, partition_key), value))")
```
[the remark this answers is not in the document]
dvadym: Interesting point. For some reason it was chosen as this ([web-page](https://pipelinedp.io/key-definitions/)). We should discuss it

pipeline_dp/contribution_bounders.py
```diff
@@ -0,0 +1,145 @@
...
+        col = backend.sample_fixed_per_key(
+            col, params.max_contributions_per_partition,
+            "Sample per (privacy_id, partition_key)")
+        report_generator.add_stage(
+            f"Per-partition contribution bounding: randomly selected not "
```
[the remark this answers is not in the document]
dvadym: Thanks, good point. PTAL

pipeline_dp/contribution_bounders.py
```diff
@@ -0,0 +1,145 @@
...
+        col = backend.sample_fixed_per_key(col,
+                                           params.max_partitions_contributed,
+                                           "Sample per privacy_id")
+
+        report_generator.add_stage(
+            f"Cross-partition contribution bounding: randomly selected not more"
+            f" than {max_partitions_contributed} partitions per privacy id")
+
+        # (privacy_id, [(partition_key, accumulator)])
+        def unnest_cross_partition_bound_sampled_per_key(pid_pk_v):
```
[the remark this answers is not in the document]
dvadym: Do you mean to rename this function? it's not just rekey, the input collection has list as value, we need to unnest it.

pipeline_dp/contribution_bounders.py
dvadym: Good catch, correct, it wouldn't work for sum and `max_sum_per_partition`, not all ContributorBound are compatible with all Combiners. `max_sum_per_partition` it's not yet implement, it will require a separate ContributionBounder

pipeline_dp/contribution_bounders.py
dvadym: I would do this as well, but the formatter returns it back :(

dvadym (commented):

miracvbasaran (changes requested):

tests/contribution_bounders_test.py
```diff
@@ -62,7 +62,7 @@ def test_contribution_bounding_per_partition_bounding_applied(self):
             input, max_partitions_contributed, max_contributions_per_partition)
 
         self.assertEqual(3, len(bound_result))
-        # Check contributions per partitions
+        # Check contribution limits per-partition are enforced.
```
miracvbasaran, line 65: Could you please make this change in other places? (e.g. line 83 but others as well)

miracvbasaran (approved):

tests/contribution_bounders_test.py
dvadym: Done

PR #303, opened 2022-07-27
Implemented compute_count_noise_params 

dvadym:
Implement function which returns noise type and standard deviation from `AggregateParams`

### DIFF
--- a/pipeline_dp/dp_computations.py
+++ b/pipeline_dp/dp_computations.py
@@ -14,6 +14,7 @@
 """Differential privacy computing of count, sum, mean, variance."""
 
 import numpy as np
+from typing import Tuple
 import pipeline_dp
 # TODO: import only modules https://google.github.io/styleguide/pyguide.html#22-imports
 from pipeline_dp.aggregate_params import NoiseKind
@@ -435,3 +436,29 @@ def compute_dp_var(count: int, normalized_sum: float,
         dp_mean += compute_middle(dp_params.min_value, dp_params.max_value)
 
     return dp_count, dp_mean * dp_count, dp_mean, dp_var
+
+
+def compute_count_noise_params(dp_params: MeanVarParams) -> Tuple[str, float]:
+    """Computes noise type and scale for DP count.
+
+    Args:
+        dp_params: parameters for DP count aggregation.
+
+    Returns:
+        Tuple (noise type ('laplace' or 'gaussian'), noise scale).
+    """
+    l0_sensitivity = dp_params.l0_sensitivity()
+    linf_sensitivity = dp_params.max_contributions_per_partition
+
+    if dp_params.noise_kind == NoiseKind.LAPLACE:
+        l1_sensitivity = compute_l1_sensitivity(l0_sensitivity,
+                                                linf_sensitivity)
+        mechanism = dp_mechanisms.LaplaceMechanism(epsilon=dp_params.eps,
+                                                   sensitivity=l1_sensitivity)
+        return "laplace", mechanism.diversity * np.sqrt(2)
+    if dp_params.noise_kind == NoiseKind.GAUSSIAN:
+        l2_sensitivity = compute_l2_sensitivity(l0_sensitivity,
+                                                linf_sensitivity)
+        return "gaussian", compute_sigma(dp_params.eps, dp_params.delta,
+                                         l2_sensitivity)
+    assert "Only Laplace and Gaussian noise is supported."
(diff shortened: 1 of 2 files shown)


pipeline_dp/dp_computations.py
```diff
@@ -435,3 +436,30 @@ def compute_dp_var(count: int, normalized_sum: float,
         dp_mean += compute_middle(dp_params.min_value, dp_params.max_value)
 
     return dp_count, dp_mean * dp_count, dp_mean, dp_var
+
+
+def compute_dp_count_noise_params(
+        dp_params: MeanVarParams) -> Tuple[str, float]:
```
miracvbasaran, line 442: Should we return the enum instead of str?

pipeline_dp/dp_computations.py
```diff
@@ -435,3 +436,30 @@ def compute_dp_var(count: int, normalized_sum: float,
...
+    """Computes noise type and standard deviation for DP count.
+
+    Args:
+        dp_params: parameters for DP count aggregation.
+
+    Returns:
+        Tuple (noise type ('laplace' or 'gaussian'), noise std).
```
miracvbasaran, line 449: nit: "stdv".

Also, I thought we were going with scale instead of stdv?

pipeline_dp/dp_computations.py
```diff
@@ -435,3 +436,30 @@ def compute_dp_var(count: int, normalized_sum: float,
...
+    """
+    l0_sensitivity = dp_params.l0_sensitivity()
```
miracvbasaran, line 451: I don't like that this logic lives in PipelineDp. It feels to me Building Blocks Library would be the ideal place for it. All this computation is already done there, no?

tests/dp_computations_test.py
```diff
@@ -495,6 +497,66 @@ def test_compute_dp_var(self):
...
+
+        noise_kind, scale = dp_computations.compute_dp_count_noise_params(
+            params)
+        self.assertEqual("gaussian", noise_kind)
+        self.assertAlmostEqual(scale, expected_scale)
+
 
 if __name__ == '__main__':
-    unittest.main()
+    absltest.main()
```
miracvbasaran, line 562: Is this intentional?

pipeline_dp/dp_computations.py
dvadym: Somewhere in `Count` algorithm, but it's not used in PipelineDP.

And that sensitivity logic is that simple, that we shouldn't hide it inside the library.  And the using the library for this makes things more complicated :).

tests/dp_computations_test.py
dvadym: yes, parametrized tests are in absltest, and not available in unittest, so in this file it's replaced using unittest to absltest

pipeline_dp/dp_computations.py
dvadym: good point! Starting from that, I've rewritten this function only to return std :). It will provide simpler API and anyway noise_kind (as enum) as a  property of `dp_params`, so the caller of this function it's trivial to get `noise_kind`

pipeline_dp/dp_computations.py
dvadym: Docstring is changed

dvadym (commented):

pipeline_dp/dp_computations.py
miracvbasaran: Sure, it's simple but I think it'd be better to not duplicate code as much as possible. I'd vote for being mindful of complexity of maintaining all places with DP logic

miracvbasaran (approved):

pipeline_dp/dp_computations.py
dvadym: Duplicating code is only one dimension of simplicity of maintenance :) and there are always tradeoff :). 

For example here:  
1. Creating `Count`/`BoundedSum` objects are slow in Python (only ~30-50K/sec).
2. `Count` object has not very friendly API for getting the amount of noise :)
3. PyDP `Count` wrapper doesn't support Gaussian noise

And independently of Count. PipelineDP has vector aggregation, we need to compute sensitivities/amount of noise for it, and it's more complicated then in scalar case, it would be weird when simple logic is outsourced to the library, while complex is in PipelineDP :)

miracvbasaran: IMO, problems 2 & 3 (and 4, meaining your last point about complexity and vector aggregations) are things that can be solved. I understand that the way C++ library and by extension Python library are implemented are not very compatible with  what we'd like to do here, but I do think there is a lot of benefit to migrating as much DP logic as we can to the libraries if we can do it in a nice and seamless way.

But we can discuss that later on, it's not specific to this PR :)

dvadym: Sure, It would be great to have all logic in one place :). Let's discuss later :)

PR #304, opened 2022-07-27
Implement DpEngine and ContributionBounder for Utility Analysis

miracvbasaran:


utility_analysis_new/dpengine.py
```diff
@@ -0,0 +1,145 @@
...
+    def _aggregate(self, col, params: pipeline_dp.AggregateParams,
+                   data_extractors: DataExtractors, public_partitions):
+        assert params.custom_combiners is None, "Cumtom combiners are not supported"
+        assert params.metrics == [
+            pipeline_dp.Metrics.COUNT
+        ], f"Supported only count metrics, metrics={params.metrics}"
+        assert public_partitions is not None, "Only public partitions supported"
+        assert params.contribution_bounds_already_enforced is False, "Utility Analysis when contribution bounds are already enforced is not supported"
+
+        # TODO: Implement custom combiner
```
dvadym, line 49: Custom combiners are combiners that can be implemented by PipelineDP users. Here it's probably meant utility analysis combiners.

utility_analysis_new/dpengine.py
```diff
@@ -0,0 +1,145 @@
...
+        combiner = self._create_compound_combiner(params)
+        # Extract the columns.
+        col = self._backend.map(
```
dvadym, line 52: it would be ideal to use` _aggregate` from DPEngine with injecting ContributionBounders and Combiners, such that it will be the same computational graph as in production. Would it work?

utility_analysis_new/dpengine.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+class UtilityAnalysisSamplingCrossAndPerPartitionContributionBounder(
+        ContributionBounder):
+    """'Bounds' the contribution by privacy_id per and cross partitions.
+
+  Because this is for Utility Analysis, it doesn't actually ensure that
+  contribution bounds are enforced. Instead, it keeps track of probabilities
+  different data points are kept under a per-partition and cross-partition
+  contribution bounder.
```
dvadym, line 89: Typo: bounding

utility_analysis_new/dpengine.py
```diff
@@ -0,0 +1,145 @@
...
+        return col
+
+    def _create_contribution_bounder(
+        self, params: pipeline_dp.AggregateParams
+    ) -> contribution_bounders.ContributionBounder:
+        """Creates Utility Analysis ContributionBounder."""
+        return UtilityAnalysisSamplingCrossAndPerPartitionContributionBounder()
+
+
+class UtilityAnalysisSamplingCrossAndPerPartitionContributionBounder(
```
dvadym, line 82: 1. Let's move it utility_analysis_new/contribution_bounders.py
2. `UtilityAnalysisSamplingCrossAndPerPartitionContributionBounder` feels too long, it seems better to drop UtiliatyAnalysis from the name. All classes are referenced with module name, so we can write like here
`import utility_analysis.contribution_bounders as utility_contribution_bounders

bounder = utility_contribution_bounders.SamplingCrossAndPerPartitionContributionBounder()
`

utility_analysis_new/dpengine.py
```diff
@@ -0,0 +1,145 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""DPEngine and ContributionBounder for utility analysis."""
+import collections
+
+import pipeline_dp
+from pipeline_dp import DPEngine
```
dvadym, line 18: Nit: by Google Python coding style modules should be imported not classes function. Please use `pipeline_dp.DPEngin`e in code. 

It's possible to use aliases for modules, like
from <module> import <sub-module> as <short name>

utility_analysis_new/dpengine.py
miracvbasaran: Yeah, that's what I meant to say. I'll be more clear

utility_analysis_new/dpengine.py
miracvbasaran: Hmm, I think the partition-selection bit would be problematic. Right now, for utility analysis, public_partitions is technically always set to None (but conceptually we are only doing public partitions), but that'd mean we have to go through the private partition selection logic of _aggregate of DPEngine, which drops some partitions. If we set public_partitions to something other than None, we have to make sure that additional public-partitions-related logic (adding partitions, dropping partitions) do not interfere with the utility analysis.

I don't see an obvious way to make it work without overriding `_aggregate()`. WDYT?

utility_analysis_new/dpengine.py
miracvbasaran: This applies to other classes as well, right? E.g. ContributionBounder or DataExtractor?

utility_analysis_new/dpengine.py
dvadym: no worries about private partition selection for now, I have ideas how to introduce them into `DPEngine.aggregate_` computation graph, I'll need to refine/experiment more with them, but it's doable.

We should use provided `public_partitions` from the user and do normal processing - filtering non public partitions and adding 0-results for missing partitions. And `DPEngine._aggregate` does that (adding of 0-results will be automatic, since `UtilityAnalysisCombiner.create_accumulator `will be called for empty set, we just need to be sure that it returns correct `ErrorPrimitive`.

Does it make sense?

utility_analysis_new/dpengine.py
dvadym: Yes

pipeline_dp/pipeline_backend.py
```diff
@@ -190,11 +190,12 @@ def group_by_key(self, col, stage_name: str):
         """Group the values for each key in the PCollection into a single sequence.
 
         Args:
-          col: input collection
+          col: input collection (k, v)
```
dvadym, line 193: Nit: with elements (key, value)

pipeline_dp/pipeline_backend.py
```diff
@@ -190,11 +190,12 @@ def group_by_key(self, col, stage_name: str):
...
           stage_name: name of the stage
 
         Returns:
-          An PCollection of tuples in which the type of the second item is list.
+          A PCollection of tuples in which the type of the second item is an
```
dvadym, line 197: Thanks for updating comments!

i.e. (key, Iterable[value])

**Explanation**: there is no `Iterable` interface, but there is [typing](https://docs.python.org/3/library/typing.html) annotations for Python. and typing.Iterable is annotations for iterable :), let's use it in this comment.

utility_analysis_new/contribution_bounders.py
```diff
@@ -0,0 +1,73 @@
...
+    def bound_contributions(self, col, params, backend, report_generator,
+                            aggregate_fn):
+        """See docstrings for this class and the base class."""
+        col = backend.map_tuple(
+            col, lambda pid, pk, v: (pid, (pk, v)),
+            "Rekey to ((privacy_id), (partition_key, value))")
+        col = backend.group_by_key(
+            col, "Group by key to get (privacy_id, [(partition_key, value)])")
+        report_generator.add_stage(
+            f"Converting ((privacy_id), (partition_key, value)) to"
```
dvadym, line 41: Explain computations reports are for users. So it should contain some high-level information, understandable for users. This looks pretty low-level, please drop it.

utility_analysis_new/contribution_bounders.py
```diff
@@ -0,0 +1,73 @@
...
+        def rekey_per_privacy_id_per_partition_key_and_unnest(pid_pk_dict):
+            privacy_id, partition_dict = pid_pk_dict
+            num_partitions_contributed = len(partition_dict)
+            for partition_key, values in partition_dict.items():
+                yield (privacy_id, partition_key), (values,
+                                                    num_partitions_contributed)
+
+        # Unnest the list per privacy id.
+        col = backend.flat_map(
+            col, rekey_per_privacy_id_per_partition_key_and_unnest, "Unnest")
```
dvadym, line 69: Nit: "Unnest" -> "Unnest per privacy id"

utility_analysis_new/dpengine.py
```diff
@@ -0,0 +1,67 @@
...
+    assert params.custom_combiners is None, "Custom combiners are not supported"
+    assert params.metrics == [
+        pipeline_dp.Metrics.COUNT
+    ], f"Supported only count metrics, metrics={params.metrics}"
+    assert public_partitions is not None, "Only public partitions supported"
+    assert params.contribution_bounds_already_enforced is False, "Utility Analysis when contribution bounds are already enforced is not supported"
+
+
+@dataclass
+class ErrorPrimitives:
```
dvadym, line 56: Nit: drop this class from here, it will be introduced in UtilityAnalysis Combiner PR

utility_analysis_new/tests/contribution_bounders_test.py
```diff
@@ -0,0 +1,95 @@
...
+        max_contributions_per_partition = 5
+
+        bound_result = self._run_contribution_bounding(
+            input, max_partitions_contributed, max_contributions_per_partition)
+
+        expected_result = [(('pid1', 'pk1'), 2), (('pid1', 'pk2'), 3),
+                           (('pid1', 'pk3'), 1), (('pid1', 'pk4'), 1),
+                           (('pid2', 'pk4'), 1)]
+        # Check per- and cross-partition contribution limits are not enforced.
+        self.assertEqual(set(expected_result), set(bound_result))
```
dvadym, line 95: Please add 

```
if __name__ == '__main__':
    absltest.main()
```

[example](https://github.com/OpenMined/PipelineDP/blob/dee896300bcf67b48363a41170e9acfefe94c77d/tests/combiners_test.py#L529)

Otherwise tests wouldn't run.

utility_analysis_new/dpengine.py
```diff
@@ -0,0 +1,67 @@
+# Copyright 2022 OpenMined.
```
dvadym, line 1: Nit:` dpengine.py` ->` dp_engine.py` and with tests.

You can use `git mv  dpengine.py dp_engine.py`

utility_analysis_new/dpengine.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def __init__(self, budget_accountant: 'BudgetAccountant',
+                 backend: 'PipelineBackend'):
+        super().__init__(budget_accountant, backend)
+
+    def aggregate(self,
+                  col,
+                  params: pipeline_dp.AggregateParams,
+                  data_extractors: pipeline_dp.DataExtractors,
+                  public_partitions=None):
```
dvadym, line 32: Please add docstring

utility_analysis_new/dpengine.py
```diff
@@ -0,0 +1,67 @@
...
+        self, params: pipeline_dp.AggregateParams
+    ) -> contribution_bounders.ContributionBounder:
+        """Creates Utility Analysis ContributionBounder."""
+        return utility_contribution_bounders.SamplingCrossAndPerPartitionContributionBounder(
+        )
+
+
+def _check_utility_analysis_params(params: pipeline_dp.AggregateParams,
+                                   public_partitions=None):
+    assert params.custom_combiners is None, "Custom combiners are not supported"
```
dvadym, line 47: Nit: please use `NotImplementedError` or `ValueError` as [here](https://github.com/OpenMined/PipelineDP/blob/dee896300bcf67b48363a41170e9acfefe94c77d/pipeline_dp/dp_engine.py#L374)

**Explanation**: `assert` is for checking invariants for checking correctness of the program. Here `params` comes from users. According to style we need to use other types of errors. `ValueError` for incorect arguments

utility_analysis_new/dpengine.py
```diff
@@ -0,0 +1,67 @@
...
+    assert params.metrics == [
+        pipeline_dp.Metrics.COUNT
+    ], f"Supported only count metrics, metrics={params.metrics}"
+    assert public_partitions is not None, "Only public partitions supported"
+    assert params.contribution_bounds_already_enforced is False, "Utility Analysis when contribution bounds are already enforced is not supported"
```
dvadym, line 52: Nit:
`not params.contribution_bounds_already_enforced`

utility_analysis_new/tests/dpengine_test.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def test_utility_analysis_params(self):
+        default_extractors = self._get_default_extractors()
+        default_params = pipeline_dp.AggregateParams(
+            noise_kind=pipeline_dp.NoiseKind.GAUSSIAN,
+            max_partitions_contributed=1,
+            max_contributions_per_partition=1,
+            metrics=[pipeline_dp.Metrics.COUNT])
+        params_with_custom_combiners = default_params
+        params_with_custom_combiners.custom_combiners = lambda x: sum(x)
```
dvadym, line 39: Nit: 
`params_with_custom_combiners.custom_combiners = sum`

(functions are just objects and they can be assigned :) )

utility_analysis_new/tests/dpengine_test.py
```diff
@@ -0,0 +1,91 @@
...
+        params_with_unsupported_metric = default_params
+        params_with_unsupported_metric.metrics = [pipeline_dp.Metrics.MEAN]
+        params_with_contribution_bounds_already_enforced = default_params
+        params_with_contribution_bounds_already_enforced.contribution_bounds_already_enforced = True
+
+        test_cases = [
+            {
+                "desc": "custom combiners",
```
dvadym, line 47: Nit: no need in `col` in `data_extractors` in test cases, anyway they are the same

utility_analysis_new/dpengine.py
```diff
@@ -0,0 +1,67 @@
...
+# limitations under the License.
+"""DPEngine for utility analysis."""
+
+import pipeline_dp
+from pipeline_dp import contribution_bounders
+import utility_analysis_new.contribution_bounders as utility_contribution_bounders
+from dataclasses import dataclass
+
+
+class DPEngine(pipeline_dp.DPEngine):
```
dvadym, line 22: I'm thinking, it might be confusing that both classes called `DPEngine`, maybe `UtilityAnalysisEngine`, WDYT?

utility_analysis_new/dpengine.py
```diff
@@ -0,0 +1,67 @@
...
+
```
dvadym, line 23: Please add docstring for the class

dvadym (commented):

utility_analysis_new/dpengine.py
miracvbasaran: This is the same case as ContributionBounders, no? I think we can use the package/module path to tell the difference between regular dp_engine and utility_analyis.dp_engine.

dvadym: ContributionBounders are less confusing:
1. They are not part of public API, so no user confusion
2.`class DPEngine(pipeline_dp.DPEngine):` the line is especially confusing (`DPEngine` is subclass of `DPEngine`)
So I think `UtilityAnalysisEngine` or something like that would be much better.

dvadym (approved):
Thanks!

PR #305, opened 2022-07-28
Extract choose_from_list_without_replacement to util

dvadym:


pipeline_dp/utils.py
```diff
@@ -0,0 +1,28 @@
...
+import numpy as np
+
+
+def choose_from_list_without_replacement(a: list, size: int) -> list:
+    if len(a) <= size:
+        return a
+    # np.random.choice makes casting of elements to numpy types
+    # which is undesirable by 2 reasons:
+    # 1. Apache Beam can not serialize numpy types.
+    # 2. It might lead for losing precision (e.g. arbitrary
```
miracvbasaran, line 24: nit: `It might lead to a loss of precision...`

pipeline_dp/dp_engine.py
```diff
@@ -12,16 +12,15 @@
...
 import dataclasses
 import functools
 from typing import Any, Callable, Tuple
-import numpy as np
 
 import pipeline_dp
 from pipeline_dp import combiners
 import pipeline_dp.contribution_bounders as contribution_bounders
 import pipeline_dp.report_generator as report_generator
+import pipeline_dp.utils as utils
```
miracvbasaran, line 23: I don't feel great about creating a utils module, maybe we can be more clear about what we are refactoring here. Do we expect to refactor away more statistics/probability-related functions? Then, maybe we can rename it to something like `stats_helpers`/`stats_utils`? If we don't expect any more utils, maybe we can just keep it in DPEngine?

miracvbasaran (changes requested):

pipeline_dp/dp_engine.py
dvadym: Thanks, good point! I don't expect now other functions. But there will be call of this function in contribution_bounder.py in the next PR (WIP).
I've renamed to sampling_utils.py, which is much better represents what this code doing

dvadym (commented):
Thanks for comments! PTAL

tests/sampling_utils_test.py
```diff
@@ -0,0 +1,42 @@
+from absl.testing import absltest
```
miracvbasaran, line 1: No license header here?

miracvbasaran (approved):

tests/sampling_utils_test.py
dvadym: Thanks, added.

dvadym (commented):
Thanks for review!

PR #307, opened 2022-07-29
Polishings

dvadym:
It contains:

1. Adding missing copyright headers
2. Fixing coding style: importing of modules instead of importing classes


miracvbasaran (approved):
Have you checked other files in the repository to see whether these two issues (lack of license headers and class imports) exists?

PR #308, opened 2022-07-29
Implementing min/max sum per partition contribution bounding

dvadym:
This PR implements `max/min_sum_per_partition` contribution bounding and implements `SUM` metric computations with this bound. This makes PipelineDP consistent with privacy on beam ([code](https://github.com/google/differential-privacy/blob/e130d473b6df005017545ac99f6b71c576845ca3/privacy-on-beam/pbeam/sum.go#L72)).

**Note:** tests and validation of parameters will be implemented soon.


pipeline_dp/dp_computations.py
```diff
@@ -23,21 +23,35 @@
...
+    def __post_init__(self):
+        assert self.min_value is None == self.max_value is None, "min_value and max_value should be or both set or both None."
+        assert self.min_sum_per_partition is None == self.max_sum_per_partition is None, "min_sum_per_partition and max_sum_per_partition should be or both set or both None."
+
     def l0_sensitivity(self):
         """"Returns the L0 sensitivity of the parameters."""
         return self.max_partitions_contributed
 
+    @property
+    def bounds_per_value_are_set(self) -> bool:
```
miracvbasaran, line 48: What if client has two metrics in one AggregateParams, mean with min_value & max_value and sum with min_sum_per_partition and max_sum_per_partition?

pipeline_dp/contribution_bounders.py
```diff
@@ -141,3 +142,73 @@ def rekey_per_privacy_id_per_partition_key(pid_pk_dict):
...
+        col = backend.flat_map(col, rekey_per_privacy_id_per_partition_key,
+                               "Unnest per privacy id")
+        # ((privacy_id, partition_key), [value])
+
+        return backend.map_values(
+            col, aggregate_fn,
+            "Apply aggregate_fn after per privacy id contribution bounding")
+
+
+def collect_values_per_partition_key_per_privacy_id(
```
miracvbasaran, line 188: There seems to be a naming collision with an inline function for SamplingPerPrivacyId, which is not great also because that means we are not able to effectively explain the difference between these two very similar functions. I think we should

1. Name them differently
2. Consider if we can refactor to potentially reuse some code? Maybe SamplingPerPrivacyId could be made to work with this new function. Going even further, we have [similar](https://github.com/google/differential-privacy/blob/main/privacy-on-beam/pbeam/aggregations.go#L870)* helper functions in Privacy on Beam that are shared throughout the codebase. Maybe there is an opportunity to do this for python as well?

* Functions that convert different permutations of (privacy_id, partition_key, value) to other permutation of (privacy_id, partition_key, value), expanding values to lists or unnesting values...

pipeline_dp/contribution_bounders.py
```diff
@@ -141,3 +142,73 @@ def rekey_per_privacy_id_per_partition_key(pid_pk_dict):
...
+        col, backend: pipeline_backend.PipelineBackend):
+    """Groups per privacy id and collects values per partition key.
+
+    The output collection is a mapping from privacy_id (i.e. each privacy_id
+     from 'col' occurs exactly once) to a list [(partition_key, [values]].
+    And for any privacy_id, each partition_key for which it contributes,
```
miracvbasaran, line 194: This sentence sounds a bit odd. Maybe something along the lines of `For any privacy_id, each partition_key it contributes to appears exactly once.` could be better?

pipeline_dp/contribution_bounders.py
dvadym: Thanks, that's very good point. I've changed `collect_values_per_partition_key_per_privacy_id` to be used in `SamplingPerPrivacyId`. Also I made usage of this function in utility analysis contribution bounder as well. PTAL

dvadym (commented):
Thanks for comments Mirac! I've addressed them. PTAL

Tests are in progress ))

pipeline_dp/dp_computations.py
```diff
@@ -35,8 +35,12 @@ class ScalarNoiseParams:
     noise_kind: pipeline_dp.NoiseKind  # Laplace or Gaussian
 
     def __post_init__(self):
-        assert self.min_value is None == self.max_value is None, "min_value and max_value should be or both set or both None."
-        assert self.min_sum_per_partition is None == self.max_sum_per_partition is None, "min_sum_per_partition and max_sum_per_partition should be or both set or both None."
+        assert (self.min_value is None) == (
```
miracvbasaran, line 38: There should be a newline before the function, no?

pipeline_dp/contribution_bounders.py
```diff
@@ -192,31 +186,32 @@ def rekey_per_privacy_id_per_partition_key(pid_pk_v_values):
...
         return backend.map_values(
             col, aggregate_fn,
-            "Apply aggregate_fn after per privacy id contribution bounding")
+            "Apply aggregate_fn after cross-partition contribution bounding")
 
 
 def collect_values_per_partition_key_per_privacy_id(
         col, backend: pipeline_backend.PipelineBackend):
-    """Groups per privacy id and collects values per partition key.
+    """Collects values per partition key in grouped by privacy_id collection.
```
miracvbasaran, line 199: nit: Not sure what this means. `Collects values per partition key keyed by privacy_id`?

pipeline_dp/contribution_bounders.py
```diff
@@ -192,31 +186,32 @@ def rekey_per_privacy_id_per_partition_key(pid_pk_v_values):
...
 
     Args:
-        col: collection with elements (privacy_id, (partition_key, value)).
+        col: collection with elements (privacy_id,
+        Iterable[(partition_key, value)]). It's assumed that each privacy_id
+        occurs only once.
         backend: pipeline backend for performing operations on collections.
 
     Returns:
         collection with elements (privacy_id, [partition_key, [values]).
```
miracvbasaran, line 213: nit: `Iterable[values]` instead of `[values]`?

pipeline_dp/aggregate_params.py
```diff
@@ -113,45 +113,83 @@ def metrics_str(self) -> str:
             return f"custom combiners={[c.metrics_names() for c in self.custom_combiners]}"
         return f"metrics={[m.value for m in self.metrics]}"
 
+    @property
+    def bounds_per_value_are_set(self) -> bool:
```
miracvbasaran, line 117: Optional: Maybe `bounds_per_contribution_are_set`? In one case, we are bounding each contribution, in the other we are bounding total contribution to a partition.

pipeline_dp/aggregate_params.py
```diff
@@ -113,45 +113,83 @@ def metrics_str(self) -> str:
             return f"custom combiners={[c.metrics_names() for c in self.custom_combiners]}"
         return f"metrics={[m.value for m in self.metrics]}"
 
+    @property
```
miracvbasaran, line 116: What does `@property` do exactly?

pipeline_dp/aggregate_params.py
```diff
@@ -113,45 +113,83 @@ def metrics_str(self) -> str:
...
-                    "finite numbers")
-            if needs_min_max_value and self.max_value < self.min_value:
-                raise ValueError(
-                    "AggregateParams: max_value must be equal to or greater than"
-                    " min_value")
-            if Metrics.VECTOR_SUM in self.metrics and \
-            (Metrics.SUM in self.metrics or \
-            Metrics.MEAN in self.metrics or \
-            Metrics.VARIANCE in self.metrics):
+        if (self.min_value is None) != (self.max_value is None):
```
miracvbasaran, line 131: It seems we have logic in multiple places to check this (and `min/max_sum_per_partition`), can we consolidate them somehow?

pipeline_dp/aggregate_params.py
```diff
@@ -113,45 +113,83 @@ def metrics_str(self) -> str:
...
+            raise ValueError(
+                "AggregateParams: min_value and max_value should be or both set or both None."
```
miracvbasaran, line 133: nit: extra `or`

pipeline_dp/aggregate_params.py
```diff
@@ -113,45 +113,83 @@ def metrics_str(self) -> str:
...
+            )
+        if (self.min_sum_per_partition is None) != (self.max_sum_per_partition
+                                                    is None):
+            raise ValueError(
+                "AggregateParams: min_sum_per_partition and max_sum_per_partition should be or both set or both None."
```
miracvbasaran, line 138: nit: extra `or`

pipeline_dp/aggregate_params.py
```diff
@@ -113,45 +113,83 @@ def metrics_str(self) -> str:
...
+            )
+
+        value_bound = self.min_value is not None
+        if value_bound:
+            self._check_range_correctness("min_value", "max_value")
+            if self.max_contributions_per_partition is None:
```
miracvbasaran, line 144: Hmm, this is getting complex. Maybe a switch-case on the parameters could be better for this type of parameter checking? E.g. something like https://github.com/google/differential-privacy/blob/main/privacy-on-beam/pbeam/pardo.go#L177

pipeline_dp/aggregate_params.py
```diff
@@ -113,45 +113,83 @@ def metrics_str(self) -> str:
...
+        if value_bound and partition_bound:
+            raise ValueError(
+                "min_value and min_sum_per_partition can not be both set.")
+
+        if self.metrics:
+            if Metrics.VECTOR_SUM in self.metrics:
+                if Metrics.SUM in self.metrics or Metrics.MEAN in self.metrics or Metrics.VARIANCE in self.metrics:
+                    raise ValueError(
+                        "AggregateParams: vector sum can not be computed together"
+                        " with scalar metrics, like sum, mean etc")
```
miracvbasaran, line 163: nit: `with scalar metrics such as sum, mean, etc.`

pipeline_dp/aggregate_params.py
```diff
@@ -113,45 +113,83 @@ def metrics_str(self) -> str:
...
+            elif partition_bound:
+                ALL_ALLOWED_METRICS = set(
+                    [Metrics.SUM, Metrics.PRIVACY_ID_COUNT, Metrics.COUNT])
+                not_allowed_metrics = set(
```
miracvbasaran, line 167: Why is this lowercase?

pipeline_dp/aggregate_params.py
```diff
@@ -113,45 +113,83 @@ def metrics_str(self) -> str:
...
-                    "with scalar metrics, like sum, mean etc")
+                    "AggregateParams: max_contributions_per_partition must be "
+                    "set when min_value/max_value are set.")
+
+        partition_bound = self.min_sum_per_partition is not None
+        if partition_bound:
+            self._check_range_correctness("min_sum_per_partition",
+                                          "max_sum_per_partition")
+
+        if value_bound and partition_bound:
```
miracvbasaran, line 154: I would move this before `if value_bound: check_range_correctness` since if you supply all 4 parameters (but incorrectly), you'll first get an error for that the range is not correct; then after you fix it you'll get an error that you can't supply value and partition bound at the same time.

pipeline_dp/contribution_bounders.py
miracvbasaran: Looks great! One minor comment: We are now calling the pipeline_dp.contribution_bounders from utility_analysis_new.contribution_bounders. Do you think there is another place more fitting for such a shared function?

pipeline_dp/aggregate_params.py
```diff
@@ -113,45 +113,83 @@ def metrics_str(self) -> str:
...
+            elif not partition_bound and not value_bound:
+                ALL_ALLOWED_METRICS = set(
+                    [Metrics.PRIVACY_ID_COUNT, Metrics.COUNT])
+                not_allowed_metrics = set(
+                    self.metrics).difference(ALL_ALLOWED_METRICS)
+                if not_allowed_metrics:
+                    raise ValueError(
+                        f"AggregateParams: for metrics {not_allowed_metrics} "
+                        f"bounds per partition are required (e.g. min_value,"
+                        f"max_value).")
```
miracvbasaran, line 183: Should we mention `min/max_sum_per_partition` here?

pipeline_dp/aggregate_params.py
```diff
@@ -187,7 +225,20 @@ def __post_init__(self):
...
-        #TODO(this PR): validation that min/max_sum_per_partition is used only for SUM and min_value, max_value are not set.
+
+    def _check_range_correctness(self, min_property_name: str,
+                                 max_property_name: str):
+        for property_name in [min_property_name, max_property_name]:
+            value = getattr(self, property_name)
+            if _not_a_proper_number(value):
+                raise ValueError(
+                    f"AggregateParams: {property_name} must be a finite number")
+        min_value = getattr(self, min_property_name)
```
miracvbasaran, line 236: What does this do if min_property_name doesn't exists?

pipeline_dp/aggregate_params.py
```diff
@@ -187,7 +225,20 @@ def __post_init__(self):
...
+        max_value = getattr(self, max_property_name)
+        if min_value > max_value:
+            raise ValueError(
+                "AggregateParams: max_value must be equal to or greater than"
```
miracvbasaran, line 240: nit: `max_property_name` instead of `max_value`? (also for min)

pipeline_dp/dp_computations.py
miracvbasaran: Why is that? Can we not support them both?

miracvbasaran (commented):
`aggregate_params_test.py` seems to be added but it's empty. Was that intentional?

pipeline_dp/contribution_bounders.py
dvadym: Updated docstring PTAL

dvadym (commented):
Thanks a lot for comments! I've addressed them, PTAL

Tests are added

pipeline_dp/contribution_bounders.py
miracvbasaran: SG, thanks!

pipeline_dp/aggregate_params.py
miracvbasaran: Is it OK to use lowercase for a const, then, for consistency? I also don't think it is a const conceptually, it takes different values in different branches.

pipeline_dp/aggregate_params.py
miracvbasaran: SG

tests/dp_engine_test.py
```diff
@@ -133,38 +133,42 @@ def test_aggregate_none(self):
...
-        dict(testcase_name='unspecified max_value MEAN',
-             error_msg='min_value and max_value must be set',
-             min_value=1,
-             max_value=None,
-             max_partitions_contributed=1,
-             max_contributions_per_partition=1,
-             max_contributions=None,
-             metrics=[pipeline_dp.Metrics.MEAN]),
+        dict(
+            testcase_name='min_value is not set, max_valus is set SUM',
```
miracvbasaran, line 137: nit: `max_value`

pipeline_dp/aggregate_params.py
miracvbasaran: So switch cases are not supported in Python?

pipeline_dp/aggregate_params.py
miracvbasaran: SG, thanks!

pipeline_dp/aggregate_params.py
dvadym: Correct, switch is not supported ((

dvadym (commented):
Thanks, I've addressed comments! PTAL

dvadym:
I've added a comment that `max/min_sum_per_partition` can't be used with `min_value/max_value` in `AggregateParams`.

`SumParams` is used for Beam/Spark API, it will be done in the subsequent PR.
Update of the documentation as well in the next PRs.

miracvbasaran (changes requested):
LGTM module typo with `max_value` and making const lowercase

miracvbasaran (approved):

PR #311, opened 2022-08-03
Return negative expected errors in utility analysis combiner

miracvbasaran:
If we define Absolute Error to be y - x where y is the DP count and x is the true count, then the per- and cross-partition errors are always nonpositive in expectation. IMO this will make the downstream computations easier.


dvadym (approved):
Thanks, it makes sense to have them negative!

PR #312, opened 2022-08-04
Use UtilityAnalysisCombiner in UtilityAnalysisDPEngine

miracvbasaran:


dvadym (commented):
Thanks, it looks good!

Could you please add 1 E2E test, with local backend, here is an [example](https://github.com/OpenMined/PipelineDP/blob/fa64ca73e2ae818fef2cd025d421c15ed59e6a5d/tests/dp_engine_test.py#L454)

dvadym (approved):
Thanks!

PR #316, opened 2022-08-11
Poisson binomial distribution computations

dvadym:
Those computations will be used in partition selection utility analysis

### DIFF
--- a/utility_analysis_new/poisson_binomial.py
+++ b/utility_analysis_new/poisson_binomial.py
@@ -22,7 +22,7 @@ from typing import Sequence, Tuple
 
 
 def compute_pmf(probabilities: Sequence[float]) -> np.ndarray:
-    """Computes probability mass function of poisson binomial distribution."""
+    """Computes probability mass function of Poisson binomial distribution."""
     # Compute coefficients of Probability Generating Function (PGF), which
     # equals to
     # PGF(x) = \product_{p\in ps}(1-p + p*x)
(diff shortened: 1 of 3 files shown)


utility_analysis_new/poisson_binomial.py
```diff
@@ -0,0 +1,55 @@
...
+
+def compute_pmf_approximation(mean: float, sigma: float, skewness: float,
+                              n: int):
+    """Computes approximate probability mass function of poisson binomial distribution.
+
+    The computation is based on paper chapter 3.3 (refined normal approximation)
+    https://www.researchgate.net/publication/257017356_On_computing_the_distribution_function_for_the_Poisson_binomial_distribution
+    """
+    G = lambda x: (norm.cdf(x) + skewness * (1 - x * x) * norm.pdf(x) / 6)
+    xs = np.arange(-0.5, n + 1, 1)
```
sushkoy, line 53: optional: wdyt on using the formula from the Chapter G((k + 0.5 − mean) / sigma) with 
k = [0, 1, 2, ..]

utility_analysis_new/poisson_binomial.py
```diff
@@ -0,0 +1,55 @@
...
+    cdf_values = G((xs - mean) / sigma)
```
sushkoy, line 54: Do we need to clamp CDF values to [0, 1] as the article suggests?

sushkoy (approved):

utility_analysis_new/poisson_binomial.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def compute_exp_std_skewness(ps: Sequence[float]) -> Tuple[float, float, float]:
+    exp = np.sum(ps)
+    std = np.sqrt(np.sum([p * (1 - p) for p in ps]))
+    skewness = np.sum([p * (1 - p) * (1 - 2 * p) for p in ps]) / std**3
+    return exp, std, skewness
+
+
+def compute_pmf_approximation(mean: float, sigma: float, skewness: float,
```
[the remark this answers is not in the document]
dvadym: No, mean, sigma, skewness need to be computed distributely, from data on many machines, we can't assume that probabilities will always fit in memory of one machine, so we need to compute mean, std, skewness with combiners in scalable way.

utility_analysis_new/poisson_binomial.py
```diff
@@ -0,0 +1,55 @@
...
+                              n: int):
+    """Computes approximate probability mass function of poisson binomial distribution.
```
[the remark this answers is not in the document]
dvadym: No :(

utility_analysis_new/poisson_binomial.py
dvadym: Good point, done!

utility_analysis_new/poisson_binomial.py
dvadym: Thanks, yeah it's better to clip

dvadym (commented):
Thanks for very useful comments!

PR #317, opened 2022-08-11
Implementation of partition selection accumulator

dvadym:
This accumulator is needed for computation of probability that the partition is kept. Computations of that probability will be in the subsequent PRs.

### DIFF
--- a/utility_analysis_new/combiners.py
+++ b/utility_analysis_new/combiners.py
@@ -98,20 +98,21 @@ class PartitionSelectionAccumulator:
             None), "Only one of probabilities and moments must be set."
 
     def __add__(self, other):
-        probs1 = self.probabilities
-        probs2 = other.probabilities
-        if probs1 and probs2 and len(probs1) + len(
-                probs2) <= MAX_PROBABILITIES_IN_ACCUMULATOR:
-            return PartitionSelectionAccumulator(probs1 + probs2)
-        moments1 = self.moments
-        if moments1 is None:
-            moments1 = _probabilities_to_moments(probs1)
+        probs_self = self.probabilities
+        probs_other = other.probabilities
+        if probs_self and probs_other and len(probs_self) + len(
+                probs_other) <= MAX_PROBABILITIES_IN_ACCUMULATOR:
+            return PartitionSelectionAccumulator(probs_self + probs_other)
+        moments_self = self.moments
+        if moments_self is None:
+            moments_self = _probabilities_to_moments(probs_self)
 
-        moments2 = other.moments
-        if moments2 is None:
-            moments2 = _probabilities_to_moments(probs2)
+        moments_other = other.moments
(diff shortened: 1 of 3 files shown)


utility_analysis_new/combiners.py
```diff
@@ -41,20 +43,107 @@ class CountUtilityAnalysisMetrics:
...
+    """Stores data needed for computing probability of keeping the partition.
+
+    Args:
+        probabilities: probabilities that each specific user contributes to the
+        partition after contribution bounding.
+        moments: contains moments of the sum of independent random
+        variables, which represent whether user contributes to the partition.
+
+    Those variables are set mutually exclusive. If len(probabilities) <= 
+    MAX_PROBABILITIES_IN_ACCUMULATOR then 'probabilities' are used otherwise
```
jspacek, line 86: Just a thought - MAX_PROBABILITIES_IN_ACCUMULATOR is kind of like an enable flag in a sense. Do you think the client may want to have more control to tune this or turn on/off? They could override it?

utility_analysis_new/combiners.py
```diff
@@ -41,20 +43,107 @@ class CountUtilityAnalysisMetrics:
...
+    moments: Optional[SumOfRandomVariablesMoments] = None
+
+    def __post_init__(self):
+        assert (self.probabilities is None) != (
+            self.moments is
+            None), "Only one of probabilities and moments must be set."
+
+    def __add__(self, other):
+        probs1 = self.probabilities
+        probs2 = other.probabilities
```
jspacek, line 102: Nit: maybe probs_self/other moments_self/_other?

utility_analysis_new/combiners.py
```diff
@@ -81,21 +170,19 @@ def create_accumulator(self, data: Tuple[Sized, int]) -> AccumulatorType:
...
-        return UtilityAnalysisCountAccumulator(
-            count=acc1.count + acc2.count,
-            per_partition_contribution_error=acc1.
-            per_partition_contribution_error +
-            acc2.per_partition_contribution_error,
-            expected_cross_partition_error=acc1.expected_cross_partition_error +
-            acc2.expected_cross_partition_error,
-            var_cross_partition_error=acc1.var_cross_partition_error +
-            acc2.var_cross_partition_error)
+        return acc1 + acc2
```
jspacek, line 185: 🎉

jspacek:
LGTM, some nits to make dp_engine_test pass 👍

jspacek (approved):

utility_analysis_new/combiners.py
dvadym: This const looks too implementation specific, it's unlikely useful for clients. If some clients will want this, we can expose as a flag, otherwise let's keep API small :)

dvadym (commented):
Thanks for review!

PR #319, opened 2022-08-12
Sum Utility Analysis combiner and tests.

jspacek:
## Description
Add a sum combiner for utility analysis.

## Affected Dependencies
None

## How has this been tested?
- Unit tests.

## Checklist

### DIFF
--- a/utility_analysis_new/combiners.py
+++ b/utility_analysis_new/combiners.py
@@ -11,10 +11,10 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""UtilityAnalysisCountCombiner."""
+"""Utility Analysis Combiners."""
 
 from dataclasses import dataclass
-from typing import List, Optional, Sized, Tuple
+from typing import List, Optional, Sequence, Sized, Tuple
 import numpy as np
 
 import pipeline_dp
@@ -147,7 +147,7 @@ class UtilityAnalysisCountCombiner(pipeline_dp.Combiner):
         self._is_public_partitions = is_public_partitions
 
     def create_accumulator(self, data: Tuple[Sized, int]) -> AccumulatorType:
-        """Create an accumulator for data.
+        """Creates an accumulator for data.
 
         Args:
             data is a Tuple containing; 1) a list of the user's contributions for a single partition, and 2) the total
@@ -182,12 +182,12 @@ class UtilityAnalysisCountCombiner(pipeline_dp.Combiner):
             ps_accumulator)
 
     def merge_accumulators(self, acc1: AccumulatorType, acc2: AccumulatorType):
-        """Merge two accumulators together additively."""
+        """Merges two accumulators together additively."""
         return acc1 + acc2
 
     def compute_metrics(self,
                         acc: AccumulatorType) -> CountUtilityAnalysisMetrics:
-        """Compute metrics based on the accumulator properties.
+        """Computes metrics based on the accumulator properties.
 
         Args:
             acc: the accumulator to compute from.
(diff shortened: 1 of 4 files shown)


utility_analysis_new/combiners.py
```diff
@@ -215,3 +215,130 @@ def metrics_names(self) -> List[str]:
...
+        Returns:
+            An accumulator with the sum of contributions and the contribution error.
+        """
+        if not data or not len(data[0]):
+            return UtilityAnalysisSumAccumulator(0, 0, 0, 0, None)
+        values, n_partitions = data
+        max_partitions = self._params.aggregate_params.max_partitions_contributed
+        prob_keep_partition = min(1, max_partitions /
+                                  n_partitions) if n_partitions > 0 else 0
+        partition_sum = np.sum(values)
```
jspacek, line 287: @dvadym I'm curious about an issue with this line which gives rounding error sometimes. I tried `math.fsum` which is reported to sometimes fix the issue, but it is still a problem even with 2 small numbers in the simple tests. Do we have a precision standard so it could be rounded? I think the default is 17 digits. Would you suggest using `Decimal` type or are we OK with some rounding?

utility_analysis_new/tests/combiners_test.py
```diff
@@ -158,8 +158,106 @@ def test_add_accumulators_moments(self):
...
+                 expected_cross_partition_error=0,
+                 std_cross_partition_error=0,
+                 std_noise=7.46484375,
+                 noise_kind=pipeline_dp.NoiseKind.GAUSSIAN)),
+        dict(testcase_name='one_partition_zero_partition_error',
+             num_partitions=1,
+             contribution_values=(1.1, 2.2),
+             params=_create_combiner_params_for_sum(0, 3.4),
+             expected_metrics=combiners.SumUtilityAnalysisMetrics(
+                 sum=3.3000000000000003,
```
jspacek, line 204: @dvadym This is the inaccurate test 😢 🐼

utility_analysis_new/combiners.py
```diff
@@ -215,3 +215,130 @@ def metrics_names(self) -> List[str]:
...
+                probabilities=[prob_keep_partition])
+
+        return UtilityAnalysisSumAccumulator(partition_sum,
+                                             per_partition_contribution_error,
+                                             expected_cross_partition_error,
+                                             var_cross_partition_error,
+                                             ps_accumulator)
+
+    def merge_accumulators(self, acc1: AccumulatorType, acc2: AccumulatorType):
+        """Merge two accumulators together additively."""
```
dvadym, line 312: Nit: Merge**s** (and in other places)

utility_analysis_new/combiners.py
```diff
@@ -215,3 +215,130 @@ def metrics_names(self) -> List[str]:
...
+
+    def compute_metrics(self,
+                        acc: AccumulatorType) -> SumUtilityAnalysisMetrics:
+        """Compute metrics based on the accumulator properties.
+
+        Args:
+            acc: the accumulator to compute from.
+
+        Returns:
+            A SumUtilityAnalysisMetrics object with computed metrics.
```
dvadym, line 323: no need in the type of output, it's already indicated in definition, after removing that, it looks like 1 line comment is enough (Args and Returns follow from definition)
Nit: leave only 1 line comment

utility_analysis_new/tests/combiners_test.py
dvadym: Comparing the floating point number is tricky. Only if a floating point number is binary rational (i.e. can be presented in binary system with the number of digits not more than float64, e.g. `3`, `0.5`, `1/256`) the operation is precise. The usual approach in tests not to rely on the exact answer of floating point (which can be different on different CPUs), but to make approximate comparison.

Please compare each floating point field with `assertAlmostEqual`. Here is an
[example](https://github.com/OpenMined/PipelineDP/blob/2940e2ba8825321c6c8c8b5a0f2f6dc290477d2f/tests/dp_computations_test.py#L57) in code of using `self.assertAlmostEqual`,

dvadym (commented):
Thanks, it looks good! 

Having SumCombiner, basically finishes implementation of Sum utility analysis, there are left, only 2 steps:

1. Allowing Sum metric in `UtilityAnalysisEngine`, [here](https://github.com/OpenMined/PipelineDP/blob/2940e2ba8825321c6c8c8b5a0f2f6dc290477d2f/utility_analysis_new/dp_engine.py#L69)
2. Creating `SumCombiner` [here](https://github.com/OpenMined/PipelineDP/blob/2940e2ba8825321c6c8c8b5a0f2f6dc290477d2f/utility_analysis_new/dp_engine.py#L58)  ([if aggregate_params.metrics = [Sum]](https://github.com/OpenMined/PipelineDP/blob/2940e2ba8825321c6c8c8b5a0f2f6dc290477d2f/utility_analysis_new/dp_engine.py#L58)).

Could you please add that?

utility_analysis_new/combiners.py
```diff
@@ -215,3 +215,130 @@ def metrics_names(self) -> List[str]:
...
+            return UtilityAnalysisSumAccumulator(0, 0, 0, 0, None)
+        values, n_partitions = data
+        max_partitions = self._params.aggregate_params.max_partitions_contributed
+        prob_keep_partition = min(1, max_partitions /
+                                  n_partitions) if n_partitions > 0 else 0
+        partition_sum = np.sum(values)
+        clipped_sum = np.clip(
+            partition_sum, self._params.aggregate_params.min_sum_per_partition,
+            self._params.aggregate_params.max_sum_per_partition)
+        per_partition_contribution_error = partition_sum - clipped_sum
```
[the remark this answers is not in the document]
jspacek: I've added `per_partition_contribution_error_min` and `per_partition_contribution_error_max` to hold the errors due to min and max clipping. I wasn't sure if the user would prefer absolute value for the min, so I left that as negative. PTAL, thank you!

utility_analysis_new/combiners.py
jspacek: Thanks for this, it works!

utility_analysis_new/combiners.py
```diff
@@ -215,3 +215,130 @@ def metrics_names(self) -> List[str]:
...
+        count = len(values)
```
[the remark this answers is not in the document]
jspacek: Ah yes, I get it! Now the cross partition errors make sense, eg. if the clipping is larger than so should these errors be. Thank you for pointing this out! PLMK if more changes are needed!

utility_analysis_new/combiners.py
jspacek: Changed the language here and in the other places, thank you!

utility_analysis_new/tests/combiners_test.py
jspacek: This works, thank you for the suggestion! I have left it at default precision (7 digits). Each field of SumUtilityAnalysisMetrics is now compared with almost equal, except noise type 😄 PLMK if there should be modifications. Thank you!

utility_analysis_new/dp_engine.py
```diff
@@ -54,21 +54,29 @@ def _create_compound_combiner(
         mechanism_type = aggregate_params.noise_kind.convert_to_mechanism_type()
         budget = self._budget_accountant.request_budget(
             mechanism_type, weight=aggregate_params.budget_weight)
-        return combiners.CompoundCombiner([
-            utility_analysis_combiners.UtilityAnalysisCountCombiner(
-                combiners.CombinerParams(budget, aggregate_params),
-                self._is_public_partitions)
-        ],
+        compound_combiners = []
+        if set(aggregate_params.metrics).issubset({pipeline_dp.Metrics.COUNT}):
```
dvadym, line 58: Nit:` if pipeline_dp.Metrics.COUNT in aggregate_params.metrics`

utility_analysis_new/dp_engine.py
```diff
@@ -54,21 +54,29 @@ def _create_compound_combiner(
...
+        return combiners.CompoundCombiner(compound_combiners,
                                           return_named_tuple=False)
 
 
 def _check_utility_analysis_params(params: pipeline_dp.AggregateParams,
                                    public_partitions=None):
     if params.custom_combiners is not None:
         raise NotImplementedError("custom combiners are not supported")
-    if params.metrics != [pipeline_dp.Metrics.COUNT]:
+    if not (set(params.metrics).issubset(
```
dvadym, line 76: I like that approach!

utility_analysis_new/dp_engine.py
```diff
@@ -54,21 +54,29 @@ def _create_compound_combiner(
...
+        {pipeline_dp.Metrics.COUNT, pipeline_dp.Metrics.SUM})):
         raise NotImplementedError(
-            f"supported only count metrics, metrics={params.metrics}")
+            f"unsupported metric in metrics={params.metrics}")
```
dvadym, line 79: To help users to understand which metrics are not supported, it's better to present only not supported, for example

```
not_supported_metrcs = list(set(params.metrics).difference({pipeline_dp.Metrics.COUNT, pipeline_dp.Metrics.SUM}))
raise NotImplementedError(
            f"unsupported metric in metrics={not_supported_metrcs}")

```

utility_analysis_new/combiners.py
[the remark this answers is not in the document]
dvadym: Thanks, I like `_min, _max` more than positive and negative!

dvadym (commented):
Thanks, it looks great! Just a few minor comments!

utility_analysis_new/dp_engine.py
jspacek: 🚀  This is cool, thank you!

utility_analysis_new/dp_engine.py
jspacek: Yay 🐍  Python functions!

utility_analysis_new/dp_engine.py
jspacek: Thanks for the catch, changed!

jspacek:
Added the migrations, it's rebased and ready for review. Thank you!

dvadym (approved):
Thanks!

PR #320, opened 2022-08-16
Fix ordering error in compound combiners.

jspacek:
## Description
Fix the ordering in the if statement so that the collection is tested against the value.

## Affected Dependencies
N/A

## How has this been tested?
- Testing will follow when compound combiners is fully implemented.

## Checklist

### DIFF
--- a/utility_analysis_new/dp_engine.py
+++ b/utility_analysis_new/dp_engine.py
@@ -55,12 +55,12 @@ class UtilityAnalysisEngine(pipeline_dp.DPEngine):
         budget = self._budget_accountant.request_budget(
             mechanism_type, weight=aggregate_params.budget_weight)
         compound_combiners = []
-        if [pipeline_dp.Metrics.COUNT] in aggregate_params.metrics:
+        if pipeline_dp.Metrics.COUNT in aggregate_params.metrics:
             compound_combiners.append(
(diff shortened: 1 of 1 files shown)


utility_analysis_new/dp_engine.py
```diff
@@ -55,12 +55,12 @@ def _create_compound_combiner(
         budget = self._budget_accountant.request_budget(
             mechanism_type, weight=aggregate_params.budget_weight)
         compound_combiners = []
-        if aggregate_params.metrics in [pipeline_dp.Metrics.COUNT]:
+        if [pipeline_dp.Metrics.COUNT] in aggregate_params.metrics:
```
dvadym, line 58: It should be 
`if pipeline_dp.Metrics.COUNT in aggregate_params.metrics:`

if we use [] around COUNT, the expression can't be true in our case:
operator in it's operator of element inclusion, e.g.

 `2 in [1,2,3]` is true, but 

`[2] in [1,2,3]` is false, because [2] is not in [1,2,3], while 

`[2] in [1, [2], 3])` is true

jspacek: Got it, thank you for the explanation!

dvadym (approved):

PR #322, opened 2022-08-18
Utility analysis for private partition selection

dvadym:


utility_analysis_new/combiners.py
```diff
@@ -114,37 +118,96 @@ def __add__(self, other):
         return PartitionSelectionAccumulator(moments=moments_self +
                                              moments_other)
 
+    def compute_probability_to_keep(self, eps: float, delta: float,
+                                    max_partitions_contributed: int) -> float:
+        """Computes the probability that this partition is kept."""
```
jspacek, line 123: WDYT about adding a few more comments here like to describe that if probabilities aren't passed, there is an approximation?

utility_analysis_new/combiners.py
```diff
@@ -114,37 +118,96 @@ def __add__(self, other):
...
+        pass
+
 
 @dataclass
 class UtilityAnalysisCountAccumulator:
     count: int
     per_partition_contribution_error: int
     expected_cross_partition_error: float
     var_cross_partition_error: float
-    partition_selection_accumulator: Optional[PartitionSelectionAccumulator]
```
jspacek, line 124: 🎉 I like this restructuring, thank you!

utility_analysis_new/dp_engine.py
```diff
@@ -80,8 +92,6 @@ def _check_utility_analysis_params(params: pipeline_dp.AggregateParams,
                 {pipeline_dp.Metrics.COUNT, pipeline_dp.Metrics.SUM}))
         raise NotImplementedError(
             f"unsupported metric in metrics={not_supported_metrics}")
-    if public_partitions is None:
```
jspacek, line 83: The test for this can now be removed in dp_engine_test, it will fail on expected error

jspacek (changes requested):
🎉 Looks great, just failing on the dp_engine_test

utility_analysis_new/dp_engine.py
dvadym: Thanks, the test is removed

utility_analysis_new/combiners.py
dvadym: Thanks, it's a good point. Done

dvadym (commented):

jspacek (approved):
LGTM! 🚀

utility_analysis_new/combiners.py
jspacek: 🚀  Thank you!

utility_analysis_new/dp_engine.py
jspacek: 👍  ✂️

PR #327, opened 2022-08-29
Convert public partitions to PCollection

dvadym:
This PR fixes intricated problem inside Beam code. The idea is that Beam PTransforms should be applied to beam PCollection. While applying them to Python list is possible, it might lead to exceptions in complicated pipelines.


jspacek (approved):
Looks great, thanks for this!

PR #328, opened 2022-09-02
contribution histogram

dvadym:


utility_analysis_new/parameter_tuning.py
```diff
@@ -0,0 +1,82 @@
...
+
+def _to_bin_lower(n: int) -> int:
+    """Finds the lower bound of the histogram bin which contains ."""
+    # For scalability reasons bins can not be all width=1. For the goals of
+    # contribution computations it is ok to make bins of large values, to be
+    # larger.
+    # Here, the following strategy is used: n is rounded down, such that only 3
+    # left-most digits of n is kept, e.g. 123->123, 1234->1230, 12345->12300.
+    bound = 1000
+    while n > bound:
```
miracvbasaran, line 40: I believe this could be optimized with bitwise operations but maybe we don't need that for now.

utility_analysis_new/parameter_tuning.py
```diff
@@ -0,0 +1,82 @@
...
+        bound *= 10
+
+    round_base = bound // 1000
+    return n // round_base * round_base
+
+
+def _compute_frequency_histogram(
+        col, backend: pipeline_dp.pipeline_backend.PipelineBackend):
+    """Computes histogram of elements frequencies in collection.
```
miracvbasaran, line 49: nit: `element` or `frequencies of elements in collection`

utility_analysis_new/parameter_tuning.py
```diff
@@ -0,0 +1,82 @@
...
+
+    Args:
+        col: collection with positive integers.
+        backend: PipelineBackend to run operations on the collection.
+    Returns:
+        1 element collection, which contains a list of FrequencyBin sorted by
```
miracvbasaran, line 55: nit: sorted in which order? Would be nice to include that here.

utility_analysis_new/parameter_tuning.py
```diff
@@ -0,0 +1,82 @@
...
+        'lower' attribute.
+    """
+
+    col = backend.count_per_element(col, "Frequency of elements")
+
+    # Combiner elements to histogram buckets of increasing sizes. Having buckets
+    # of size = 1 is not scalable.
```
miracvbasaran, line 62: nit: `width` instead of `size` for consistency.

utility_analysis_new/parameter_tuning.py
```diff
@@ -0,0 +1,82 @@
...
+        col, lambda n, f:
+        (_to_bin_lower(n),
+         FrequencyBin(lower=_to_bin_lower(n), count=f, sum=f * n, max=n)),
+        "To FrequencyBin")
+
+    # (lower_bin_value, FrequencyBin)
+    col = backend.combine_per_key(col, operator.add, "Combine FrequencyBins")
+    # (lower_bin_value, FrequencyBin)
+    col = backend.values(col, "To FrequencyBin")
+    # (FrequencyBin,)
```
miracvbasaran, line 73: nit: Delete `,`

utility_analysis_new/parameter_tuning.py
```diff
@@ -0,0 +1,82 @@
...
+
+    # Combiner elements to histogram buckets of increasing sizes. Having buckets
+    # of size = 1 is not scalable.
+    col = backend.map_tuple(
+        col, lambda n, f:
+        (_to_bin_lower(n),
+         FrequencyBin(lower=_to_bin_lower(n), count=f, sum=f * n, max=n)),
+        "To FrequencyBin")
+
+    # (lower_bin_value, FrequencyBin)
```
miracvbasaran, line 69: We can't have this comment line be under the previous code line again? Maybe having a nextline after this comment helps?

utility_analysis_new/parameter_tuning.py
```diff
@@ -0,0 +1,82 @@
...
+from dataclasses import dataclass
+import operator
+
+
+@dataclass
+class FrequencyBin:
+    """Represents 1 bin of the histogram.
+
+    The bin represents integers between 'lower' (inclusive) and 'upper'
+     (exclusive, not stored in this class).
```
miracvbasaran, line 11: This is a bit ominous, maybe we could briefly mentiond that 'upper' is a deterministic function of 'lower', and we can mention what that function is.

miracvbasaran (changes requested):

pipeline_dp/pipeline_backend.py
```diff
@@ -468,9 +499,16 @@ def merge_accumulators(accumulators):
...
         return self.map_values(self.group_by_key(col), merge_accumulators)
 
+    def combine_per_key(self, col, combine_fn: Callable, stage_name: str):
+        combine = lambda elements: functools.reduce(combine_fn, elements)
+        return self.map_values(self.group_by_key(col), combine)
+
     def flatten(self, col1, col2, stage_name: str = None):
         return itertools.chain(col1, col2)
 
+    def to_list(self, col, stage_name: str):
```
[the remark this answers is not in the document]
dvadym: It's not no-op for local backend, it converts
[1,2,3,4] -> [[1,2,3,4]]

utility_analysis_new/parameter_tuning.py
dvadym: It makes sense about having deterministic function, added. Which function exactly, it's independent of the class itself, since there might be different conversions from n to bin

utility_analysis_new/parameter_tuning.py
```diff
@@ -0,0 +1,82 @@
...
+                            self.sum + other.sum, max(self.max, other.max))
+
+    def __eq__(self, other):
+        return self.lower == other.lower and self.count == other.count and self.sum == other.sum and self.max == other.max
+
+
+def _to_bin_lower(n: int) -> int:
+    """Finds the lower bound of the histogram bin which contains ."""
+    # For scalability reasons bins can not be all width=1. For the goals of
+    # contribution computations it is ok to make bins of large values, to be
```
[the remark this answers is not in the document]
dvadym: thanks, done

utility_analysis_new/parameter_tuning.py
dvadym: Yeah, it will be unlikely to be a bottleneck, so no need to optimize. Here it was chosen decimal digits instead of binary for making it readable, rather than fast :)

dvadym (commented):
Thanks for comments! PTAL

miracvbasaran (approved):

PR #334, opened 2022-09-12
Minor improvements

dvadym:


utility_analysis_new/combiners.py
```diff
@@ -125,8 +124,7 @@ def compute_probability_to_keep(self, eps: float, delta: float,
             eps, delta, max_partitions_contributed)
         probability = 0
         for i, prob in enumerate(pmf):
-            # TODO: Replace 0.5 with ps_strategy.probability_of_keep(i)
-            probability += prob * 0.5
+            probability += prob * ps_strategy.probability_of_keep(i)
```
dvadym, line 127: Otherwise partition selection utility analysis doesn't work

utility_analysis_new/dp_engine.py
```diff
@@ -84,6 +84,9 @@ def _create_compound_combiner(
     def _select_private_partitions_internal(self, col,
                                             max_partitions_contributed: int,
                                             max_rows_per_privacy_id: int):
+        # Utility analysis of private partition selection is performed in a
```
dvadym, line 87: I've added that comments, otherwise it's not clear why this function is no-op

utility_analysis_new/dp_engine.py
```diff
@@ -84,6 +84,9 @@ def _create_compound_combiner(
...
+        # corresponding combiners (unlike DP computations). So this function is
```
miracvbasaran, line 88: Optional: `unlike actual DP computations` for emphasis

miracvbasaran (approved):

PR #336, opened 2022-09-15
Implementation of choosing Partition selection strategy

dvadym:


pipeline_dp/aggregate_params.py
```diff
@@ -89,6 +95,8 @@ class AggregateParams:
          max_contributions_per_partition. This option can be used if the dataset
          does not contain any identifiers that can be used to enforce
          contribution bounds automatically.
+        partition_selection_strategy: which strategy to use for private
+         partition selection. It is ignored when public paritions are used.
```
miracvbasaran, line 99: nit: typo in `partitions`

pipeline_dp/aggregate_params.py
```diff
@@ -108,6 +116,7 @@ class AggregateParams:
     vector_max_norm: Optional[float] = None
     vector_size: Optional[int] = None
     contribution_bounds_already_enforced: bool = False
+    partition_selection_strategy: PartitionSelectionStrategy = PartitionSelectionStrategy.TRUNCATED_GEOMETRIC
```
miracvbasaran, line 119: I think it makes sense to do the optimization in Go here:
if max_partitions_contributed > 3
partition_selection_strategy = GAUSSIAN_THRESHOLDING

(also below for select_partitions)

pipeline_dp/aggregate_params.py
```diff
@@ -53,6 +53,12 @@ class NormKind(Enum):
     L2 = "l2"
 
 
+class PartitionSelectionStrategy(Enum):
```
miracvbasaran, line 56: Idle thought: I wonder if having a PUBLIC_PARTITIONS option for PartitionSelectionStrategy would help with code clarity. We have many checks (`if public_partitions is not None`) for public_partitions that could be replaced with the enum. But that might require some refactoring. WDYT?

miracvbasaran (approved):
LG, thanks!

pipeline_dp/aggregate_params.py
dvadym: it's an option, but it would require users to set this parameter directly:
  - to set `PUBLIC_PARTITIONS` when public partitions are used
  - to set some other method otherwise
  
 while it's more preferable that the users change this parameter only if needed (it's definitely a complicated parameter and it's hard for most users to understand educated decision)

pipeline_dp/aggregate_params.py
dvadym: I'm thinking to make at some point `auto`, which would do something like that. But to make it later, when PLD accounting is ready.

It can be more complicated if advanced composition is used, since for Gaussian_Thresholding advance composition is used automatically (just by computing sensitivity), but for TRUNCATED_GEOMETRIC not

pipeline_dp/aggregate_params.py
miracvbasaran: The users would not necessarily have to set it themselves though, we could set it for them once if public_partitions are supplied.

PR #337, opened 2022-09-16
Compute aggregate metrics for partition selection

miracvbasaran:
Aggregate metric computation for count now also takes into account probability that partition is kept.


utility_analysis_new/combiners.py
```diff
@@ -18,6 +18,7 @@
 import numpy as np
 import math
 import scipy
+from numpy import ndarray
```
dvadym, line 21: Google Python coding style, import modules, not classes
Nit: remove this import and use `np.ndarray`

utility_analysis_new/combiners.py
```diff
@@ -552,22 +574,26 @@ def __init__(self, params: pipeline_dp.combiners.CombinerParams,
...
-    def create_accumulator(
-            self, metrics: CountUtilityAnalysisMetrics) -> AccumulatorType:
+    def create_accumulator(self,
+                           metrics: CountUtilityAnalysisMetrics,
+                           probability_to_keep: float = 1) -> AccumulatorType:
         """Creates an accumulator for metrics."""
         # Absolute error metrics
-        abs_error_expected = metrics.per_partition_error + metrics.expected_cross_partition_error
-        abs_error_variance = metrics.std_cross_partition_error**2 + metrics.std_noise**2
+        abs_error_expected = probability_to_keep * (
```
dvadym, line 582: We need to compute weighted average. The weighted average of `a1,..,an` with weight `w1,..,wn` equals to
`(w1a1 + ... + wnan)/(w1+...+ wn)`.
Here weights are `probability_to_keep`, so we need to compute sum of weights and divide by sum of weight, not on `num_partitions` (this sum is equal to num_partition when all `probability_to_keep=1`)

dvadym (commented):
It looks great overall. I've left some comments

dvadym (approved):
It looks great. There are some minor comments, I'll adress them in the subsequent PR

PR #348, opened 2022-09-30
Setting 0 for relative errors when actual value = 0

dvadym:
Setting errors to inf, make the result inf, i.e. for any public partitions the result is inf. While it's a good question how to count relative error for added empty partitions, returning inf doesn't sound as a good option


dvadym:
Thanks for review!

At some moment of time we definitely need to think how to deal with empty partitions and also how to compute relative error in a better way. 
This PR basically excludes empty partitions from relative error computations.

PR #350, opened 2022-10-21
Refactoring: Using tuples instead of dataclasses for Utility Analysis accumulators.

dvadym:
This is done for performance reasons - it's much faster to work with tuples and serialized size of tuples is ~10 times smaller.

It turned out that the changes are pretty small and the resulting code structuring is nice:

1. Before:
```
def create_accumulators():
  ...
  return Accumulator(...)

```
After
```
def create_accumulators():
  ...
  return (...)  # the same code, just Accumulator name dropped

```

2. Before
```
  def compute_metrics(acc)
      # using acc.field1, acc.field2 etc
```
After
```
  def compute_metrics(acc)
     field1, field2.. = acc 
     # replace acc.field1 -> field1 ...
```

Also, it's introduced a base class for UtilityAnalysisCombiners.

### DIFF
--- a/utility_analysis_new/combiners.py
+++ b/utility_analysis_new/combiners.py
@@ -288,27 +288,9 @@ class SumUtilityAnalysisMetrics:
     noise_kind: pipeline_dp.NoiseKind
 
 
-@dataclass
-class UtilityAnalysisSumAccumulator:
-    sum: float
-    per_partition_error_min: float
-    per_partition_error_max: float
-    expected_cross_partition_error: float
-    var_cross_partition_error: float
-
-    def __add__(self, other):
-        return UtilityAnalysisSumAccumulator(
-            self.sum + other.sum,
-            self.per_partition_error_min + other.per_partition_error_min,
-            self.per_partition_error_max + other.per_partition_error_max,
-            self.expected_cross_partition_error +
-            other.expected_cross_partition_error,
-            self.var_cross_partition_error + other.var_cross_partition_error)
-
-
 class UtilityAnalysisSumCombiner(pipeline_dp.Combiner):
     """A combiner for utility analysis sums."""
-    AccumulatorType = UtilityAnalysisSumAccumulator
+    AccumulatorType = Tuple[float, float, float, float, float]
 
     def __init__(self, params: pipeline_dp.combiners.CombinerParams):
         self._params = params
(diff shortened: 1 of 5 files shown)


Kulankhina (approved):

utility_analysis_new/combiners.py
```diff
@@ -28,6 +29,29 @@
...
+    def create_accumulator(self, data: Tuple[Sized, int]):
+        """Creates an accumulator for data.
+
+        Args:
+            data is a Tuple containing:
+              1) a list of the user's contributions for a single partition
+              2) the total number of partitions a user contributed to.
+
+        Returns:
+            A tuple which is accumulator.
```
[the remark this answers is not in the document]
dvadym: Done

miracvbasaran, line 44: nit: `an accumulator`

utility_analysis_new/combiners.py
```diff
@@ -201,25 +225,9 @@ class CountUtilityAnalysisMetrics:
...
-            self.expected_cross_partition_error +
-            other.expected_cross_partition_error,
-            self.var_cross_partition_error + other.var_cross_partition_error)
-
-
-class UtilityAnalysisCountCombiner(pipeline_dp.Combiner):
+class UtilityAnalysisCountCombiner(UtilityAnalysisCombiner):
     """A combiner for utility analysis counts."""
-    AccumulatorType = UtilityAnalysisCountAccumulator
+    AccumulatorType = Tuple[int, int, float, float]
```
miracvbasaran, line 230: I think it is important to document what those ints and floats mean somewhere (e.g. maybe here), so something like:

`[count, per_partition_error, expected_cross_partition_error, var_cross_partition_error]`

(also for other aggregations)

utility_analysis_new/combiners.py
```diff
@@ -416,39 +372,16 @@ class PrivacyIdCountUtilityAnalysisMetrics:
...
-    def __add__(self, other):
-        return UtilityAnalysisPrivacyIdCountAccumulator(
-            self.privacy_id_count + other.privacy_id_count,
-            self.expected_cross_partition_error +
-            other.expected_cross_partition_error,
-            self.var_cross_partition_error + other.var_cross_partition_error)
-
-
-class UtilityAnalysisPrivacyIdCountCombiner(pipeline_dp.Combiner):
+class UtilityAnalysisPrivacyIdCountCombiner(UtilityAnalysisCombiner):
```
miracvbasaran, line 375: nit: What do you think about dropping `UtilityAnalysis` from some of the class names here? E.g. this could be `PrivacyIdCountCombiner(UtilityAnalysisCombiner)`, repeating `UtilityAnalysis` twice seems reduntant.

Similarly, `SumUtilityAnalysisMetrics` could be `SumMetrics` since the module/package(?) name already has utility_analysis in it, but there I am not as sure since there could still be confusion if you import without the module/package name.

miracvbasaran (changes requested):

miracvbasaran (approved):

PR #351, opened 2022-10-21
DP percentiles

dvadym:
This PR implements DP percentile computations.

API:
It adds metric `metrics.Percentile`, so now it's possible to compute percentiles with `DPEngine.aggregate`
   
`dp_engine.aggregate(... metrics=[metrics.Count, metrics.Percentile(50), metrics.Percentile(90)] ...)`

The result looks like:
(8, MetricsTuple(count=14532.95396910912, percentile_50=2.9999810443476416, percentile_90=4.99996013795033))

Implementation:
1. `Metrics`: previously it was `enum`, but metrics.Percentile(50) can't be implemented with enum, so `Metric` class was implemented.
2. `QuantileCombiner` is implemented. It uses `QuantileTree` object which is wrapper from PyDP for [the object](https://github.com/google/differential-privacy/blob/main/cc/algorithms/quantile-tree.h) from Google C++ building block library

### DIFF
--- a/pipeline_dp/aggregate_params.py
+++ b/pipeline_dp/aggregate_params.py
@@ -37,6 +37,9 @@ class Metric:
     def __repr__(self):
         return self.__str__()
 
+    def __hash__(self):
+        return hash(str(self))
+
     @property
     def is_percentile(self):
         return self.name == 'PERCENTILE'
--- a/pipeline_dp/aggregate_params.py
+++ b/pipeline_dp/aggregate_params.py
@@ -20,13 +20,39 @@ import math
 import logging
 
 
-class Metrics(Enum):
-    COUNT = 'count'
-    PRIVACY_ID_COUNT = 'privacy_id_count'
-    SUM = 'sum'
-    MEAN = 'mean'
-    VARIANCE = 'variance'
-    VECTOR_SUM = 'vector_sum'
+class Metric:
+
+    def __init__(self, name: str, parameter: Optional[Any] = None):
+        self.name = name
+        self.parameter = parameter
+
+    def __eq__(self, other: 'Metric') -> bool:
+        return self.name == other.name and self.parameter == other.parameter
+
+    def __str__(self):
+        if self.parameter is None:
+            return self.name
+        return f'{self.name}({self.parameter})'
+
+    def __repr__(self):
+        return self.__str__()
+
+    @property
+    def is_percentile(self):
+        return self.name == 'PERCENTILE'
+
+
+class Metrics:
+    COUNT = Metric('COUNT')
+    PRIVACY_ID_COUNT = Metric('PRIVACY_ID_COUNT')
+    SUM = Metric('SUM')
+    MEAN = Metric('MEAN')
+    VARIANCE = Metric('VARIANCE')
+    VECTOR_SUM = Metric('VECTOR_SUM')
+
+    @classmethod
+    def PERCENTILE(cls, value: float):
+        return Metric('PERCENTILE', value)
 
 
 class NoiseKind(Enum):
--- a/pipeline_dp/aggregate_params.py
+++ b/pipeline_dp/aggregate_params.py
@@ -20,11 +20,17 @@ import math
 import logging
 
 
+@dataclass
 class Metric:
+    """Represents DP metrics.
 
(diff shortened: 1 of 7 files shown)


pipeline_dp/combiners.py
```diff
@@ -274,11 +276,13 @@ def explain_computation(self) -> ExplainComputationReport:
 
 
 class MeanCombiner(Combiner):
-    """Combiner for computing DP Mean. Also returns sum and count in addition to
-    the mean.
-    The type of the accumulator is a tuple(count: int, normalized_sum: float) that holds
-    the count and normalized sum of elements in the dataset for which this accumulator is
-    computed.
+    """Combiner for computing DP Mean.
```
dvadym, line 279: it's just formatting

pipeline_dp/combiners.py
```diff
@@ -392,6 +399,80 @@ def explain_computation(self) -> ExplainComputationReport:
...
+    It uses QuantileTree algorithm from Google C++ DP library
+    https://github.com/google/differential-privacy/blob/main/cc/algorithms/quantile-tree.h
+    with the wrapper from PyDP.
+
+    The accumulator is QuantileTree object serialized to string.
+    """
+
+    AccumulatorType = str
+
+    def __init__(self, params, percentiles: List[float]):
```
sushkoy, line 415: Can we call this "percentiles_to_compute" to differentiate from actual percentiles? Also for quantiles, and in other places.

It's a bit longer but less confusing, especially that you use "quantiles" to mean something else right below (line 438).

pipeline_dp/combiners.py
```diff
@@ -392,6 +399,80 @@ def explain_computation(self) -> ExplainComputationReport:
...
+        # https://github.com/google/differential-privacy/blob/605ec87bcbd4a536995b611132dbf4d341d2e91d/cc/algorithms/quantile-tree.h#L47
+        DEFAULT_TREE_HEIGHT = 4
+        DEFAULT_BRANCHING_FACTOR = 16
+        return quantile_tree.QuantileTree(
+            self._params.aggregate_params.min_value,
+            self._params.aggregate_params.max_value, DEFAULT_TREE_HEIGHT,
+            DEFAULT_BRANCHING_FACTOR)
+
+    def _noise_type(self) -> str:
+        return "laplace" if self._params.aggregate_params.noise_kind == pipeline_dp.NoiseKind.LAPLACE else "gaussian"
```
sushkoy, line 473: This assumes there are no other noise types in future.
Can we use something more stable, e.g. 
`return self._params.aggregate_params.noise_kind.value`?

pipeline_dp/aggregate_params.py
```diff
@@ -20,13 +20,49 @@
...
-class Metrics(Enum):
-    COUNT = 'count'
-    PRIVACY_ID_COUNT = 'privacy_id_count'
-    SUM = 'sum'
-    MEAN = 'mean'
-    VARIANCE = 'variance'
-    VECTOR_SUM = 'vector_sum'
+@dataclass
+class Metric:
+    """Represents DP metrics.
```
sushkoy, line 25: nit: Represents a DP metric.

pipeline_dp/aggregate_params.py
```diff
@@ -20,13 +20,49 @@
...
+
+    Attributes:
+        name: the name of the metric, like 'COUNT', 'PERCENTILE'.
+        parameter: parameter of the metric, e.g. for 90th percentile,
```
sushkoy, line 29: nit: an optional parameter ...

pipeline_dp/aggregate_params.py
```diff
@@ -20,13 +20,49 @@
...
+    """Contains all supported DP metrics."""
+    COUNT = Metric('COUNT')
+    PRIVACY_ID_COUNT = Metric('PRIVACY_ID_COUNT')
+    SUM = Metric('SUM')
+    MEAN = Metric('MEAN')
+    VARIANCE = Metric('VARIANCE')
+    VECTOR_SUM = Metric('VECTOR_SUM')
+
+    @classmethod
+    def PERCENTILE(cls, value: float):
```
sushkoy, line 64: value -> percentile_to_compute, to be more explicit?

sushkoy (approved):

PR #352, opened 2022-10-21
Refactoring: Move histograms code to a separate file

dvadym:
This PR is a move, no changes in logic. Histogram logic is moved from `parameter_tuning.py` to `histograms.py`.

Histograms logic got pretty complicated (and will be even more). It makes sense to put it in a separate file.


utility_analysis_new/histograms.py
```diff
@@ -0,0 +1,243 @@
...
+    """Represents 1 bin of the histogram.
+
+  The bin represents integers between 'lower' (inclusive) and 'upper'
+   (exclusive, not stored in this class, but uniquely determined by 'lower').
+
+  Attributes:
+      lower: the lower bound of the bin.
+      count: the number of elements in the bin.
+      sum: the sum of elements in the bin.
+      max: the maximum element in the bin (which might be smaller than upper-1).
```
Kulankhina, line 33: nit: would it make sense to say "smaller or equal to the upper-1" or not?

Kulankhina (approved):

utility_analysis_new/histograms.py
dvadym: yes, it makes sense.

PR #354, opened 2022-10-24
Refactoring: Using tuples instead of dataclasses for Partition Selection Utility Analysis accumulator

dvadym:
This is done for performance reasons - it's much faster to work with tuples and serialized size of tuples is ~10 times smaller.

### DIFF
--- a/utility_analysis_new/combiners.py
+++ b/utility_analysis_new/combiners.py
@@ -288,27 +288,9 @@ class SumUtilityAnalysisMetrics:
     noise_kind: pipeline_dp.NoiseKind
 
 
-@dataclass
-class UtilityAnalysisSumAccumulator:
-    sum: float
-    per_partition_error_min: float
-    per_partition_error_max: float
-    expected_cross_partition_error: float
-    var_cross_partition_error: float
-
-    def __add__(self, other):
-        return UtilityAnalysisSumAccumulator(
-            self.sum + other.sum,
-            self.per_partition_error_min + other.per_partition_error_min,
-            self.per_partition_error_max + other.per_partition_error_max,
-            self.expected_cross_partition_error +
-            other.expected_cross_partition_error,
-            self.var_cross_partition_error + other.var_cross_partition_error)
-
-
 class UtilityAnalysisSumCombiner(pipeline_dp.Combiner):
     """A combiner for utility analysis sums."""
-    AccumulatorType = UtilityAnalysisSumAccumulator
+    AccumulatorType = Tuple[float, float, float, float, float]
 
     def __init__(self, params: pipeline_dp.combiners.CombinerParams):
         self._params = params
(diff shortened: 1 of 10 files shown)


utility_analysis_new/combiners.py
```diff
@@ -82,8 +82,8 @@ def _probabilities_to_moments(
 
 
 @dataclass
```
miracvbasaran, line 84: This is still a dataclass, do we want to keep it that way?

utility_analysis_new/combiners.py
```diff
@@ -158,49 +142,65 @@ def compute_pmf(self) -> np.ndarray:
         return pmf
 
 
-class PartitionSelectionCombiner(pipeline_dp.Combiner):
+# PartitionSelectionAccumulator = (probabilities, moments). They are mutually
```
miracvbasaran, line 145: nit: Maybe `These two`/`The two elements` instead of `They` is more descriptive, I could see someone being confused about what `They` refers to here.

utility_analysis_new/combiners.py
```diff
@@ -158,49 +142,65 @@ def compute_pmf(self) -> np.ndarray:
...
-        Args:
-            data is a Tuple containing; 1) a list of the user's contributions
-            for a single partition, and 2) the total number of partitions a user
-            contributed to.
-
-        Returns:
-            An accumulator for computing probability of selecting partition.
-        """
+    def create_accumulator(
+            self, data: Tuple[Sized, int]) -> PartitionSelectionAccumulator:
```
miracvbasaran, line 178: Python question: Does it make a performance difference to use `PartitionSelectionAccumulator` instead of Tuple[Optional[Tuple[float]], Optional[SumOfRandomVariablesMoments]] directly here?

utility_analysis_new/combiners.py
```diff
@@ -601,22 +601,28 @@ def __init__(self, params: pipeline_dp.combiners.CombinerParams,
         self._params = params
         self._error_quantiles = error_quantiles
 
-    def create_accumulator(self, prob_to_keep: float) -> AccumulatorType:
+    def create_accumulator(
+            self, prob_to_keep: float) -> PartitionSelectionAccumulator:
         """Creates an accumulator for metrics."""
-        return PartitionSelectionAccumulator(probabilities=[prob_to_keep])
+        return ((prob_to_keep,), None)
```
miracvbasaran, line 607: What does the `,` after `prob_to_keep` do?

(Also exists in some places in the tests)

miracvbasaran (approved):

utility_analysis_new/combiners.py
dvadym: Type annotations doesn't have any performance impact, as for me `Tuple[Optional[Tuple[float]], Optional[SumOfRandomVariablesMoments]]` is just too complicated :)

utility_analysis_new/combiners.py
dvadym: `(10)` is an expression which equals `10` (just by arithmetic rules, as in any programming language)
`(10,)` is a tuple with 1 element `10`

dvadym (commented):
Thanks for review!

PR #355, opened 2022-10-25
Refactoring: Moving utility analysis metrics dataclasses to a separate file

dvadym:


utility_analysis_new/metrics.py
```diff
@@ -0,0 +1,114 @@
...
+"""Dataclasses with Utility Analysis result metrics."""
+
+import pipeline_dp
+from dataclasses import dataclass
+from typing import List, Optional
+import math
+
+
+@dataclass
+class CountMetrics:
```
Kulankhina, line 23: nit: I would have a very slight preference for calling this "CountUtilityMetrics" since there are so many metrics that we typically discuss (including count being a metric itself). But this is a very optional suggestion.

Kulankhina (approved):

utility_analysis_new/metrics.py
dvadym: Previously this class has a name: `CountUtilityAnalysisMetrics`, a few days ago in discussion with Mirac, we've decided to drop `UtilityAnalysis`, since anyway those metrics are inside `utility_analysis` package and it doesn't make sense to duplicate utility.

PR #358, opened 2022-10-26
Sampling per partitions

dvadym:
This PR implements sub-sampling of partitions for utility analysis.
It works the following:
1.`UtilityAnalysisEngine.aggregate` has sampling_probability
2. Utility analysis ContributionBounder can perform sub-sampling of partitions using `ValueSampler` if sampling probability < 1.0
3. `ValueSampler` computes hash of partition key and if the hash (which is assumed to return uniform values) is less than sampling_rate*max_hash_value, the partition_key is kept.

### DIFF
--- a/pipeline_dp/sampling_utils.py
+++ b/pipeline_dp/sampling_utils.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import numpy as np
+import hashlib
 
 
 def choose_from_list_without_replacement(a: list, size: int) -> list:
@@ -26,3 +27,25 @@ def choose_from_list_without_replacement(a: list, size: int) -> list:
     sampled_indices = np.random.choice(np.arange(len(a)), size, replace=False)
 
     return [a[i] for i in sampled_indices]
+
+
+def _compute_64bit_hash(v) -> int:
+    m = hashlib.sha1()
+    m.update(repr(v).encode())
+    return int(m.hexdigest()[:16], 16)
+
+
+class DeterministicSampler:
+    """Hash based sampler.
+
+    For a value it returns whether this value should be kept. The keeping
+    decision is deterministic for fixed value. On average the decision to keep
+    happens in sampling_rate cases.
+    """
+
+    def __init__(self, sampling_rate: float):
+        self._sample_bound = int(round(2**64 * sampling_rate))
+
+    def keep(self, value) -> bool:
+        """Returns true if the value should be kept."""
+        return _compute_64bit_hash(value) < self._sample_bound
--- a/pipeline_dp/sampling_utils.py
+++ b/pipeline_dp/sampling_utils.py
@@ -39,7 +39,7 @@ class ValueSampler:
     """Deterministic value sampler.
 
     For a value it returns whether this value should be kept. The keeping
-    decision is deterministic for a fixed value. For randomly chosen value the
+    decision is deterministic for a fixed value. For a randomly chosen value the
     decision to keep happens with probability sampling_rate.
     """
 
--- a/pipeline_dp/sampling_utils.py
+++ b/pipeline_dp/sampling_utils.py
@@ -36,10 +36,10 @@ def _compute_64bit_hash(v) -> int:
 
 
 class DeterministicSampler:
-    """Hash based sampler.
+    """Deterministic value sampler.
 
     For a value it returns whether this value should be kept. The keeping
-    decision is deterministic for fixed value. On average the decision to keep
+    decision is deterministic for a fixed value. On average the decision to keep
     happens in sampling_rate cases.
     """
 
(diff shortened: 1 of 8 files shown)


pipeline_dp/sampling_utils.py
```diff
@@ -26,3 +27,25 @@ def choose_from_list_without_replacement(a: list, size: int) -> list:
...
+    m.update(repr(v).encode())
+    return int(m.hexdigest()[:16], 16)
+
+
+class ValueSampler:
+    """Deterministic value sampler.
+
+    For a value it returns whether this value should be kept. The keeping
+    decision is deterministic for a fixed value. On average the decision to keep
+    happens in sampling_rate cases.
```
miracvbasaran, line 43: nit: It happens in `sampling_rate*num_values` cases right? Maybe we can say instead `happens with probability sampling_rate`?

utility_analysis_new/contribution_bounders.py
```diff
@@ -13,20 +13,28 @@
...
     """'Bounds' the contribution by privacy_id per and cross partitions.
 
     Because this is for Utility Analysis, it doesn't actually ensure that
     contribution bounds are enforced. Instead, it keeps track of probabilities
-    different data points are kept under per-partition and cross-partition
-    contribution bounding.
+    different data points are kept under per-partition (Linf) and
+    cross-partition (L0) contribution bounding.
+
+    If partitions_sampling_prob < 1.0, partitions are subsampled. This sampling
```
miracvbasaran, line 28: Should we subsample on privacy IDs instead? I think that'd give a better estimate for cross-partition bounding error at the expense of a worse estimate for partition selection error. But for that we can correct by the sampling rate.

utility_analysis_new/contribution_bounders.py
```diff
@@ -42,12 +50,19 @@ def bound_contributions(self, col, params, backend, report_generator,
 
         # (privacy_id, [(partition_key, [value])])
 
-        # Rekey by (privacy_id, partition_key) and unnest values along with the
+        # Rekey by (privacy_id, partition_key) and unnest val_coues along with the
```
miracvbasaran, line 53: nit: `values`

utility_analysis_new/tests/contribution_bounders_test.py
```diff
@@ -96,6 +98,25 @@ def test_contribution_bounding_cross_partition_bounding_not_applied(self):
...
+        max_partitions_contributed = 3
+        max_contributions_per_partition = 5
+
+        bound_result = self._run_contribution_bounding(
+            input,
+            max_partitions_contributed,
+            max_contributions_per_partition,
+            partitions_sampling_prob=0.7)
+
+        # 'pk1' and 'pk2' are dropped by subsampling.
```
miracvbasaran, line 114: This is very flaky, no? Could you please compute the flakiness of this and put it as comments?

utility_analysis_new/utility_analysis.py
```diff
@@ -32,6 +32,7 @@ class UtilityAnalysisOptions:
     aggregate_params: pipeline_dp.AggregateParams
     multi_param_configuration: Optional[
         dp_engine.MultiParameterConfiguration] = None
+    partitions_sampling_prob: float = 1
 
     def __post_init__(self):
```
miracvbasaran, line 37: Could we validate `0 < partitions_sampling_prob <= 1` here?

pipeline_dp/sampling_utils.py
dvadym: It's tricky to write about probability, since "The keeping decision is deterministic for a fixed value"

I've updated the text. PTAL

utility_analysis_new/contribution_bounders.py
dvadym: This is very good question! Yes, we can, but it's more complicated to make it reliable for partition selection, that's why I chose to start with per partition selection. 

We can implement sampling per partition later

utility_analysis_new/tests/contribution_bounders_test.py
dvadym: The test is deterministic, since ValueSampler is deterministic for the same partition key

utility_analysis_new/utility_analysis.py
dvadym: Yes, thanks

dvadym (commented):
Thanks for comments! PTAL

pipeline_dp/sampling_utils.py
```diff
@@ -39,8 +39,8 @@ class ValueSampler:
     """Deterministic value sampler.
 
     For a value it returns whether this value should be kept. The keeping
-    decision is deterministic for a fixed value. On average the decision to keep
-    happens in sampling_rate cases.
+    decision is deterministic for a fixed value. For randomly chosen value the
```
miracvbasaran, line 42: supernit: `For a randomly chosen value...`

miracvbasaran (approved):

dvadym:
Thanks for review!

PR #359, opened 2022-10-27
Computing partition count histograms

dvadym:
This PR implements computing of histograms for `(partition_key, count)` (i.e. how many partitions have count=1, how many partitions have count=2, ... )   and `(partition_key, privacy_id_count)` (i.e. how many partitions have privacy_id_count=1, how many partitions have privacy_id_count=2, ... ).

Those are the same histograms which are used for computing cross and per partition contributions per privacy_id.

These histograms will be used in parameter tuning.

### DIFF
--- a/utility_analysis_new/histograms.py
+++ b/utility_analysis_new/histograms.py
@@ -47,10 +47,17 @@ class FrequencyBin:
         return self.lower == other.lower and self.count == other.count and self.sum == other.sum and self.max == other.max
 
 
+class HistogramType(enum.Enum):
+    L0_CONTRIBUTIONS = 'l0_contributions'
+    LINF_CONTRIBUTIONS = 'linf_contributions'
+    COUNT_PER_PARTITION = 'count_per_partition'
+    COUNT_PRIVACY_ID_PER_PARTITION = 'privacy_id_per_partition_count'
+
+
 @dataclass
 class Histogram:
     """Represents a histogram over integers."""
-    name: str
+    name: HistogramType
     bins: List[FrequencyBin]
 
     def total_count(self):
@@ -92,20 +99,13 @@ class Histogram:
         return result[::-1]
 
 
-class HistogramType(enum.Enum):
-    L0_CONTRIBUTIONS = 'l0_contributions'
-    LINF_CONTRIBUTIONS = 'linf_contributions'
-    COUNT_PER_PARTITION_COUNT = 'count_per_partition_count'
-    PRIVACY_ID_PER_PARTITION_COUNT = 'privacy_id_per_partition_count'
-
-
 @dataclass
 class ContributionHistograms:  # todo: another name?
     """Histograms of privacy id contributions."""
-    cross_partition_histogram: Histogram
-    per_partition_histogram: Histogram
-    count_per_partition_count: Histogram
-    privacy_id_per_partition_count: Histogram
+    l0_contributions_histogram: Histogram
+    linf_contributions_histogram: Histogram
+    count_per_partition_histogram: Histogram
+    count_privacy_id_per_partition: Histogram
 
 
 def _to_bin_lower(n: int) -> int:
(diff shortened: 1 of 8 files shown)


utility_analysis_new/histograms.py
```diff
@@ -152,31 +162,36 @@ def bins_to_histogram(bins):
...
-def _compute_cross_partition_histogram(
+        if histogram.name == HistogramType.L0_CONTRIBUTIONS:
+            l0_contributions = histogram
+        elif histogram.name == HistogramType.LINF_CONTRIBUTIONS:
+            linf_contributions = histogram
+        elif histogram.name == HistogramType.COUNT_PER_PARTITION:
+            count_per_partition = histogram
+        elif histogram.name == HistogramType.COUNT_PRIVACY_ID_PER_PARTITION:
+            privacy_id_per_partition_count = histogram
+    return DatasetHistograms(l0_contributions, linf_contributions,
```
miracvbasaran, line 176: This assumes all of (lo_contributions, linf_contributions, count_per_partition, privacy_id_per_partition_count) will be not None, and so is brittle. Maybe we could refactor a little to not assume this or assert that all of them are not None?

utility_analysis_new/histograms.py
```diff
@@ -229,13 +305,20 @@ def compute_contribution_histograms(
...
-    cross_partition_histogram = _compute_cross_partition_histogram(col, backend)
+    cross_partition_histogram = _compute_l0_contributions_histogram(
+        col, backend)
     # 1 element collection: ContributionHistogram
-    per_partition_histogram = _compute_per_partition_histogram(col, backend)
+    per_partition_histogram = _compute_linf_contributions_histogram(
+        col, backend)
+
+    partition_count_histogram = _compute_partition_count_histogram(col, backend)
+    partition_privacy_id_count_histogram = _compute_partition_privacy_id_count_histogram(
```
miracvbasaran, line 315: I feel like we are doing very similar computations to come up with these 4 histograms. Maybe we can refactor a bit to get rid of unnecessary computation?

That said, this might not be a bottleneck, in which case we could ignore this.

utility_analysis_new/tests/histograms_test.py
```diff
@@ -156,14 +162,102 @@ def test_compute_cross_partition_histogram(self, input, expected):
...
+            expected=[
+                FrequencyBin(lower=1, count=20, sum=20, max=1),
+                FrequencyBin(lower=2, count=5, sum=10, max=2),
+            ]),
+    )
+    def test_compute_partitions_count_histogram(self, input, expected):
+        histogram = hist._compute_partition_count_histogram(
+            input, pipeline_dp.LocalBackend())
+        histogram = list(histogram)[0]
+        self.assertEqual(hist.HistogramType.COUNT_PER_PARTITION, histogram.name)
```
miracvbasaran, line 215: Optional: You could change `expected` to be a DatasetHistogram, which would also have the name information; so you could simply call self.assertEqual in the code.

Although I am not sure how self.assertEqual handles objects with lists inside.

utility_analysis_new/tests/parameter_tuning_test.py
```diff
@@ -51,8 +51,9 @@ def test_find_candidate_parameters(
         mock_linf_histogram = histograms.Histogram(None, None)
         mock_linf_histogram.quantiles = mock.Mock(return_value=[3, 6, 6])
 
-        mock_histograms = histograms.ContributionHistograms(
-            mock_l0_histogram, mock_linf_histogram)
+        mock_histograms = histograms.DatasetHistograms(
+            mock_l0_histogram, mock_linf_histogram,
+            histograms.Histogram(None, None), histograms.Histogram(None, None))
```
miracvbasaran, line 56: Related to my comment above, here we had to create fake Histograms; whereas we shouldn't have to create fake ones if we don't need them.

miracvbasaran (commented):
This looks great, thanks Vadym!

utility_analysis_new/histograms.py
dvadym: Yeah, good point! In principle it's fine that some histograms are `None`, but we shouldn't crash :). I've initialized all variable to None and update the test.

dvadym (commented):
Thanks for comments! PTAL

miracvbasaran (approved):

utility_analysis_new/histograms.py
```diff
@@ -187,15 +189,12 @@ def _compute_l0_contributions_histogram(
...
     Args:
       col: collection with elements (privacy_id, partition_key).
+        Assumption: 'col' contains different elements!
       backend: PipelineBackend to run operations on the collection.
 
     Returns:
       1 element collection, which contains the computed Histogram.
     """
-
-    col = backend.distinct(col, "Distinct (privacy_id, partition_key)")
```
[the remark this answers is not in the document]
dvadym: It's a refactoring for performance improvements, distinct is called in the caller code, since 2 histogram computations require distinct.

PR #363, opened 2022-11-01
Speed up probabilities computations for partition selection

dvadym:
One of the step of computation of probability that a partition is kept is computing probabilities that after contribution bounding the partition will have 0, 1, ..., N privacy ids (where N is the number of privacy ids w/o contribution bounding). N can be large (billions). So computing all probabilities for i=0, ... N is non-efficient. This PR makes it faste by limiting computing probabilities to `i in (mean - 8*sigma, mean+8*sigma)`, since for all i outside of this range, probabilities are close to 0.

### DIFF
--- a/utility_analysis_new/poisson_binomial.py
+++ b/utility_analysis_new/poisson_binomial.py
@@ -24,6 +24,14 @@ from dataclasses import dataclass
 
 @dataclass
 class PMF:
+    """Represents probability mass function of finite integer value distribution.
+
+    Attributes:
+        start: the minimmum value of
+        probabilities: the i-th element represents the probability that the
+         distribution attains value i+start.
+
+    """
     start: int
     probabilities: np.ndarray
 
--- a/utility_analysis_new/poisson_binomial.py
+++ b/utility_analysis_new/poisson_binomial.py
@@ -24,10 +24,10 @@ from dataclasses import dataclass
 
 @dataclass
 class PMF:
-    """Represents probability mass function of finite integer value distribution.
+    """Represents probability mass function of a finite integer value distribution.
 
     Attributes:
-        start: the minimmum value of
+        start: the minimal value that this distribution can produce.
         probabilities: the i-th element represents the probability that the
          distribution attains value i+start.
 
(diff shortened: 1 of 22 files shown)


utility_analysis_new/poisson_binomial.py
```diff
@@ -51,8 +66,16 @@ def compute_pmf_approximation(mean: float, sigma: float, skewness: float,
     The computation is based on paper chapter 3.3 (refined normal approximation)
     https://www.researchgate.net/publication/257017356_On_computing_the_distribution_function_for_the_Poisson_binomial_distribution
```
sushkoy, line 67: Can we add the new assumption to the doc?

E.g. values with negligible probabilities will be dropped for efficiency reasons. The total probability for the dropped values is no more than 10^(-15).

utility_analysis_new/poisson_binomial.py
```diff
@@ -51,8 +66,16 @@ def compute_pmf_approximation(mean: float, sigma: float, skewness: float,
...
     """
+    if sigma == 0:
+        return PMF(int(round(mean)), np.array([1]))
     G = lambda x: (norm.cdf(x) + skewness * (1 - x * x) * norm.pdf(x) / 6)
-    xs = np.arange(-1, n + 1)
+    # n can be large. Assuming that the output distribution is close to the
+    # normal, for all i further than 8*sigma contributes less than 10^(-15).
```
sushkoy, line 73: Does this affect any DP guarantees downstream? Or does this make them only tighter?
Might be helpful to add a comment.

utility_analysis_new/poisson_binomial.py
```diff
@@ -19,6 +19,21 @@
 import numpy as np
 from scipy.stats import norm
 from typing import Sequence, Tuple
+from dataclasses import dataclass
+
+
+@dataclass
+class PMF:
+    """Represents probability mass function of finite integer value distribution.
```
sushkoy, line 27: nit: of [a] finite integer value distribution

sushkoy (approved):

PR #365, opened 2022-11-03
Implement cross-partition metrics for privacy_id_count

miracvbasaran:
Also refactor AggregateMetrics to contain multiple aggregations

DO NOT SUBMIT yet, I had to comment out some dependencies due to using 3.10. Likewise, some tests fail because I had to hardcode some values for missing dependencies.

### DIFF
--- a/pipeline_dp/dp_computations.py
+++ b/pipeline_dp/dp_computations.py
@@ -106,7 +106,6 @@ def compute_sigma(eps: float, delta: float, l2_sensitivity: float):
     # TODO: use named arguments, when argument names are added in PyDP on PR
     # https://github.com/OpenMined/PyDP/pull/398.
     return dp_mechanisms.GaussianMechanism(eps, delta, l2_sensitivity).std
-    return 1
 
 
 def apply_laplace_mechanism(value: float, eps: float, l1_sensitivity: float):
--- a/pipeline_dp/dp_computations.py
+++ b/pipeline_dp/dp_computations.py
@@ -17,7 +17,7 @@ import numpy as np
 from typing import Optional
 import pipeline_dp
 from dataclasses import dataclass
-# from pydp.algorithms import numerical_mechanisms as dp_mechanisms
+from pydp.algorithms import numerical_mechanisms as dp_mechanisms
 
 
 @dataclass
@@ -105,7 +105,7 @@ def compute_sigma(eps: float, delta: float, l2_sensitivity: float):
     """
     # TODO: use named arguments, when argument names are added in PyDP on PR
     # https://github.com/OpenMined/PyDP/pull/398.
-    # DO-NOT-SUBMIT
+    return dp_mechanisms.GaussianMechanism(eps, delta, l2_sensitivity).std
     return 1
 
 
@@ -466,9 +466,9 @@ def _compute_noise_std(linf_sensitivity: float,
     if dp_params.noise_kind == pipeline_dp.NoiseKind.LAPLACE:
         l1_sensitivity = compute_l1_sensitivity(dp_params.l0_sensitivity(),
                                                 linf_sensitivity)
-        #mechanism = dp_mechanisms.LaplaceMechanism(epsilon=dp_params.eps,
-        #                                           sensitivity=l1_sensitivity)
-        return np.sqrt(2)
+        mechanism = dp_mechanisms.LaplaceMechanism(epsilon=dp_params.eps,
+                                                   sensitivity=l1_sensitivity)
+        return mechanism.diversity * np.sqrt(2)
     if dp_params.noise_kind == pipeline_dp.NoiseKind.GAUSSIAN:
         l2_sensitivity = compute_l2_sensitivity(dp_params.l0_sensitivity(),
                                                 linf_sensitivity)
--- a/pipeline_dp/dp_computations.py
+++ b/pipeline_dp/dp_computations.py
@@ -17,7 +17,7 @@ import numpy as np
 from typing import Optional
 import pipeline_dp
 from dataclasses import dataclass
-from pydp.algorithms import numerical_mechanisms as dp_mechanisms
+# from pydp.algorithms import numerical_mechanisms as dp_mechanisms
 
 
 @dataclass
@@ -105,7 +105,8 @@ def compute_sigma(eps: float, delta: float, l2_sensitivity: float):
     """
     # TODO: use named arguments, when argument names are added in PyDP on PR
     # https://github.com/OpenMined/PyDP/pull/398.
-    return dp_mechanisms.GaussianMechanism(eps, delta, l2_sensitivity).std
+    # DO-NOT-SUBMIT
+    return 1
 
 
 def apply_laplace_mechanism(value: float, eps: float, l1_sensitivity: float):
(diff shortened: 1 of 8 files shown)


utility_analysis_new/utility_analysis.py
```diff
@@ -102,27 +102,87 @@ def perform_utility_analysis(col,
     # accumulators : (aggregate_metrics)
 
     def pack_metrics(aggregate_metrics) -> List[metrics.AggregateMetrics]:
+        if options.multi_param_configuration is None:
```
dvadym, line 105: Yeah, that's a good interview task to simplify those computations :)

What do you think about the following:

1.`options.n_parameters`returns number of parameters for which Utility analysis is computed. 
2. we can assume that `aggregate_metrics` has `options.n_parameters` groups each corresponds to 1 configuration  (this will be the only assumption on format)
3. for those `i%options.n_parameters == 0`, create a new `metrics.AggregateMetrics()`
4. for all i use `_populate_packed_metrics` for the last created `metrics.AggregateMetrics`

utility_analysis_new/combiners.py
```diff
@@ -550,6 +552,105 @@ def compute_metrics(self,
...
+        return [
+            'abs_error_expected', 'abs_error_variance', 'abs_error_quantiles',
+            'rel_error_expected', 'rel_error_variance', 'rel_error_quantiles'
+        ]
+
+    def explain_computation(self):
+        pass
+
+
+class PrivacyIdCountAggregateErrorMetricsCombiner(pipeline_dp.Combiner):
```
dvadym, line 583: What do you think about the following unifictation of count and privacy id count cross-partition metrics:
1. To Drop `PrivacyIdCountMetrics` and to use `CountMetrics`. It looks like the only different between them is that CountMetrics has   `per_partition_error` which should be set `0` for `PRIVACY_ID_COUNT`
2. To use `CountAggregateErrorMetricsCombiner` for both `COUNT` `PRIVACY_ID_COUNT` cross-partition aggregations.

utility_analysis_new/utility_analysis.py
```diff
@@ -102,27 +102,37 @@ def perform_utility_analysis(col,
...
+                _populate_packed_metrics(packed_metrics, aggregate_metrics[j])
+            return_list.append(packed_metrics)
+        return return_list
 
     return backend.map(aggregates, pack_metrics,
                        "Pack metrics from the same run")
     # (aggregate_metrics)
 
 
+# Sets the appropriate field of packed_metrics with aggregate_error_metric
```
dvadym, line 124: Nit: Python docstring style
```
def funcname() ...
"""1 line description""
```
or
```
def funcname() ...
"""1 line description

<optional long description>

Args:
  arg1: ..

Returns:
""
```

Here for me it looks like 1 line description is the best, something like
"Sets the the appropriate packed_metrics field with 'metric' according to 'metric' type. " (and to rename aggregate_error_metric -> metric)

utility_analysis_new/tests/utility_analysis_test.py
```diff
@@ -136,23 +140,35 @@ def test_w_public_partitions(self, noise_kind, abs_error_expected,
...
         self.assertLen(col, 1)
-        # Assert there is only a single AggregateErrorMetrics & no metrics for
+        # Assert there is only a single AggregateMetrics & no metrics for
         # partition selection are returned.
         self.assertLen(col[0], 1)
-        # Assert count metrics are reasonable.
+        # Assert privacy id count metrics are reasonable.
         self.assertAlmostEqual(
-            col[0][0].aggregate_error_metrics.abs_error_expected,
+            col[0][0].privacy_id_count_metrics.abs_error_expected,
```
dvadym, line 148: Nit: to avoid duplication
```
def check_metrics(metrics):
   self.assertEqual(abs_error_expected, abs_error_expected)
    ....

check_metric(col[0][0].privacy_id_count_metrics)
check_metric(col[0][0].count_metrics)
```
(AFAIU count and privacy_id count metrics are the same, since each privacy id contributes once, so no per-partition error)

dvadym (commented):
Thanks, it looks great.

utility_analysis_new/combiners.py
```diff
@@ -549,7 +520,13 @@ def merge_accumulators(self, acc1: AccumulatorType, acc2: AccumulatorType):
     def compute_metrics(self,
                         acc: AccumulatorType) -> metrics.AggregateErrorMetrics:
         """Computes metrics based on the accumulator properties."""
+        return self._compute_metrics(acc, metrics.AggregateMetricType.COUNT)
+
+    def _compute_metrics(self, acc: AccumulatorType,
```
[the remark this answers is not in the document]
miracvbasaran: Because we need to specify the `metric_type` differently for PrivacyIdCountCombiner and for CountCombiner. Is there a more Python-idiomatic way of doing this?

If `compute_metrics()` takes as input `metric_type`, I think the pipeline would fail because the collection doesn't have that.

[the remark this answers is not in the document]
dvadym: Thanks for explanation!

pipeline_dp/dp_computations.py
```diff
@@ -106,6 +106,7 @@ def compute_sigma(eps: float, delta: float, l2_sensitivity: float):
     # TODO: use named arguments, when argument names are added in PyDP on PR
     # https://github.com/OpenMined/PyDP/pull/398.
     return dp_mechanisms.GaussianMechanism(eps, delta, l2_sensitivity).std
+    return 1
```
dvadym, line 109: drop this

pipeline_dp/partition_selection.py
```diff
@@ -20,7 +20,6 @@ def create_partition_selection_strategy(
         strategy: pipeline_dp.PartitionSelectionStrategy, epsilon: float,
         delta: float,
         max_partitions_contributed: int) -> "PartitionSelectionStrategy":
-
     if strategy == pipeline_dp.PartitionSelectionStrategy.TRUNCATED_GEOMETRIC:
```
dvadym, line 23: Please return back

utility_analysis_new/combiners.py
```diff
@@ -569,14 +546,27 @@ def compute_metrics(self,
...
+            'abs_error_quantiles', 'rel_error_expected', 'rel_error_variance',
+            'rel_error_quantiles'
         ]
 
     def explain_computation(self):
         pass
 
 
+class PrivacyIdCountAggregateErrorMetricsCombiner(
+        CountAggregateErrorMetricsCombiner):
```
dvadym, line 559: I thought again. 

WDYT about avoiding class inheritance here, i.e. use `CountAggregateErrorMetricsCombiner`, and to have 
`metrics.AggregateMetricType` in the constructor of this class. It seems simpler, no additional classes.

utility_analysis_new/utility_analysis.py
```diff
@@ -102,41 +102,56 @@ def perform_utility_analysis(col,
...
+        # UtilityAnalysisEngine._create_compound_combiner().
         if pipeline_dp.Metrics.COUNT in aggregate_params.metrics:
             internal_combiners.append(
                 utility_analysis_combiners.CountAggregateErrorMetricsCombiner(
                     None, error_quantiles))
+        if pipeline_dp.Metrics.PRIVACY_ID_COUNT in aggregate_params.metrics:
+            internal_combiners.append(
+                utility_analysis_combiners.
+                PrivacyIdCountAggregateErrorMetricsCombiner(
+                    None, error_quantiles))
```
dvadym, line 155: Nit: drop params, since they are not used.

pipeline_dp/partition_selection.py
miracvbasaran: We return at the end of the function, do you mean undo deleting the newline?

dvadym: yes

dvadym (approved):
Thanks!

PR #371, opened 2022-11-16
Use float instead of np.float64

miracvbasaran:
float takes less space when serializing, leading to performance improvements as data is passed around workers.


dvadym (commented):
Could you please add verification in combiner_tests that all fields in the output are floats or list of floats?

You can do it for any dataclass by converting dataclass to tuple with [astuple](https://docs.python.org/3/library/dataclasses.html#dataclasses.astuple)

dvadym (approved):
Thanks!

PR #372, opened 2022-11-25
Implement more cross-partition metrics

miracvbasaran:
## Description
This includes noise_variance and expectations that take into account error due to dropped partitions.


utility_analysis_new/metrics.py
```diff
@@ -96,6 +100,24 @@ class AggregateErrorMetrics:
...
+    # These metrics take into account loss due to dropping partitions when
+    # computing expectation.
+    #
+    # When public partitions are used, these will be exactly equal to
+    # abs/rel_error_expected.
+    #
+    # When private partitions are used, see the following example with a single
+    # partition on how they are different:
+    #
+    # Given 1 partition with probability_to_keep=0.4, actual_count=100,
```
dvadym, line 112: Nit on comment structure
**The following error metrics ...**

\<This example\>   _(no need to mention private partition selection, formulas for those metrics independent, but keep_probability = 1 for public partitoins)_

**When the public partitions ...**  (after the example, it's more context why they are the same)

dvadym (approved):
Thanks, just minor comments!

PR #373, opened 2022-12-02
Run utility analysis on pre-aggregated data

dvadym:
This PR implements running utility analysis on pre-aggregated data (previously it was possible only on the raw data).

The implementations consists of the following:

1. The extracting of needed columns (e.g. privacy_id, partition_key) with extractors is moved to a separate `DPEngine._extract_columns` method which is overriden in UtilityAnalysisEngine (because for pre-aggregated data the different data extraction).
2. Computing of count, sum per `(privacy_id, partition_key)` was moved from `CompoundCombiner` to `ContributionBounding` (that's because count and sum is already computed in for per-aggregated data)
3. Introduced `NoOpContributionBounder` for using for pre-aggregated data (because all that needed for contribution bounding is already computed in pre-aggregted data).
4. `pre_aggregated_data` field is introduced in all structures where needed.

**Note:** Tests will be fixed and new tests will added in the following commits

### DIFF
--- a/utility_analysis_new/dp_engine.py
+++ b/utility_analysis_new/dp_engine.py
@@ -12,90 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """DPEngine for utility analysis."""
-import copy
-import dataclasses
-from typing import Callable, Iterable, Optional, Sequence, Union
+from typing import Iterable, Union
 
 import pipeline_dp
 from pipeline_dp import budget_accounting
 from pipeline_dp import combiners
 from pipeline_dp import contribution_bounders
 from pipeline_dp import pipeline_backend
+import utility_analysis_new
 import utility_analysis_new.contribution_bounders as utility_contribution_bounders
 import utility_analysis_new.combiners as utility_analysis_combiners
-
-
-@dataclasses.dataclass
-class PreAggregateExtractors:
-    partition_extractor: Callable
-    preaggregate_extractor: Callable
-
-
-@dataclasses.dataclass
-class MultiParameterConfiguration:
-    """Specifies parameters for multi-parameter Utility Analysis.
-
-    All MultiParameterConfiguration attributes corresponds to attributes in
-    pipeline_dp.AggregateParams.
-
-    UtilityAnalysisEngine can perform utility analysis for multiple sets of
-    parameters simultaneously. API for this is the following:
-    1. Specify blue-print AggregateParams instance.
-    2. Set MultiParameterConfiguration attributes (see the example below). Note
-    that each attribute is a sequence of parameter values for which the utility
-    analysis will be run. All attributes that have non-None values must have
-    the same length.
-    3. Pass the created objects to UtilityAnalysisEngine.aggregate().
-
-    Example:
-        max_partitions_contributed = [1, 2]
-        max_contributions_per_partition = [10, 11]
-
-        Then the utility analysis will be performed for
-          AggregateParams(max_partitions_contributed=1, max_contributions_per_partition=10)
-          AggregateParams(max_partitions_contributed=2, max_contributions_per_partition=11)
-    """
-    max_partitions_contributed: Sequence[int] = None
-    max_contributions_per_partition: Sequence[int] = None
-    min_sum_per_partition: Sequence[float] = None
-    max_sum_per_partition: Sequence[float] = None
-
-    def __post_init__(self):
-        attributes = dataclasses.asdict(self)
-        sizes = [len(value) for value in attributes.values() if value]
-        if not sizes:
-            raise ValueError("MultiParameterConfiguration must have at least 1"
-                             " non-empty attribute.")
-        if min(sizes) != max(sizes):
-            raise ValueError(
-                "All set attributes in MultiParameterConfiguration must have "
-                "the same length.")
-        if (self.min_sum_per_partition is None) != (self.max_sum_per_partition
-                                                    is None):
-            raise ValueError(
-                "MultiParameterConfiguration: min_sum_per_partition and "
-                "max_sum_per_partition must be both set or both None.")
-        self._size = sizes[0]
-
-    @property
-    def size(self):
-        return self._size
-
-    def get_aggregate_params(self, params: pipeline_dp.AggregateParams,
-                             index: int) -> pipeline_dp.AggregateParams:
-        """Returns AggregateParams with the index-th parameters."""
-        params = copy.copy(params)
-        if self.max_partitions_contributed:
-            params.max_partitions_contributed = self.max_partitions_contributed[
-                index]
-        if self.max_contributions_per_partition:
-            params.max_contributions_per_partition = self.max_contributions_per_partition[
-                index]
-        if self.min_sum_per_partition:
-            params.min_sum_per_partition = self.min_sum_per_partition[index]
-        if self.max_sum_per_partition:
-            params.max_sum_per_partition = self.max_sum_per_partition[index]
-        return params
 
 
 class UtilityAnalysisEngine(pipeline_dp.DPEngine):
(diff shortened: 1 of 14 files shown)


utility_analysis_new/dp_engine.py
```diff
@@ -50,11 +51,13 @@ def aggregate(self,
...
-    def analyze(self,
-                col,
-                options: utility_analysis_new.UtilityAnalysisOptions,
-                data_extractors: pipeline_dp.DataExtractors,
-                public_partitions=None):
+    def analyze(
+            self,
+            col,
+            options: utility_analysis_new.UtilityAnalysisOptions,
+            data_extractors: Union[pipeline_dp.DataExtractors,
```
monsieurmuffin, line 58: If we indeed need a dedicated data structure for PreAggregatedExtractors, please  could you update the function doc and explaining that PreAggregateExtractors should be provided when utility analysis is run on pre-aggregated data, otherwise DataExtractors should be provided.

utility_analysis_new/dp_engine.py
```diff
@@ -142,6 +147,19 @@ def _select_private_partitions_internal(
         # function is no-op.
         return col
 
+    def _extract_columns(
+        self, col,
+        data_extractors: Union[pipeline_dp.DataExtractors,
+                               utility_analysis_new.PreAggregateExtractors]):
+        """Extract columns using data_extractors."""
+        if self._options.pre_aggregated_data:
+            # todo: check data_extractors is PreAggregateExtractors
```
monsieurmuffin, line 156: nit: WDYT about switching to the Google code-style todo format (i.e., "TODO:")? (though I realize that we can't fully use the internal formatting)

utility_analysis_new/dp_engine.py
```diff
@@ -142,6 +147,19 @@ def _select_private_partitions_internal(
...
+            return self._backend.map(
```
monsieurmuffin, line 157: Can the extraction logic be moved to PreAggregateExtractors ? It seems to belong there conceptually.

utility_analysis_new/parameter_tuning.py
```diff
@@ -67,12 +67,19 @@ class TuneOptions:
           if this argument is a callable, it should take 1 argument of type
           AggregateErrorMetrics and return float.
         parameters_to_tune: specifies which parameters to tune.
+        partitions_sampling_prob: the probability with which each partition
```
monsieurmuffin, line 70: From the description it looks like this method might potentially apply sampling. We discussed that we might also want to apply sampling at pre-aggregation stage. 
Do you think it might be helpful to mention that the caller might apply sampling before passing the input dataset to this function and what kind of sampling can be used. Or is it too generic / obvious?

dvadym (commented):
Thanks a lot for comments! I've addressed them and also added tests.

utility_analysis_new/tests/dp_engine_test.py
```diff
@@ -106,29 +112,39 @@ def test_utility_analysis_params(self):
         params_with_custom_combiners.custom_combiners = sum
```
monsieurmuffin, line 112: It's a little bit difficult to understand what behavior is being tested by this test. Would you mind renaming the test and separating the arrange, act and assert blocks with blank lines or comments ?
Are we testing that "test_invalid_utility_analysis_params_throws_exception"?

utility_analysis_new/tests/dp_engine_test.py
```diff
@@ -137,18 +153,21 @@ def test_utility_analysis_params(self):
                 budget_accountant = budget_accounting.NaiveBudgetAccountant(
```
monsieurmuffin, line 153: Would it be possible to move the code that isn't expected to throw the exception outside the "with" block to make it clear what logic is being tested?

utility_analysis_new/tests/dp_engine_test.py
```diff
@@ -106,29 +112,39 @@ def test_utility_analysis_params(self):
...
+            "public_partitions": [1],
+            "pre_aggregated": False
+        }, {
+            "desc": "contribution bounds are already enforced",
+            "params": params_with_contribution_bounds_already_enforced,
+            "data_extractor": default_extractors,
+            "public_partitions": [1],
+            "pre_aggregated": False
+        }, {
+            "desc":
```
monsieurmuffin, line 138: nit: WDYT about renaming "desc" to "error_message" to make it more explicit how it's supposed to be used?

utility_analysis_new/tests/dp_engine_test.py
```diff
@@ -137,18 +153,21 @@ def test_utility_analysis_params(self):
...
                 col = [0, 1, 2]
-
                 engine.analyze(col,
                                options,
                                test_case["data_extractor"],
                                public_partitions=test_case["public_partitions"])
 
-    def test_aggregate_public_partition_e2e(self):
+    @parameterized.parameters(False, True)
+    def test_aggregate_public_partition_e2e(self, pre_aggregated: bool):
```
monsieurmuffin, line 170: nit: are we testing analyze or aggregate? Is there a naming convention for the Python test names?
Should it be "test_analyze_applies_public_partitions"?

utility_analysis_new/tests/dp_engine_test.py
```diff
@@ -137,18 +153,21 @@ def test_utility_analysis_params(self):
...
         # Arrange
```
monsieurmuffin, line 171: nit: please could you also add an "Act" comment? I am not sure what method is being tested, is it "DPEngine.analyze"?

utility_analysis_new/tests/dp_engine_test.py
```diff
@@ -183,7 +211,8 @@ def test_aggregate_public_partition_e2e(self):
         self.assertLen(col, 3)
```
monsieurmuffin, line 211: Are we reusing the same variable for input and result? I can't find any guidance on this but we never  do this in Java) Is it acceptable for Python code style?

utility_analysis_new/tests/dp_engine_test.py
```diff
@@ -160,20 +179,29 @@ def test_aggregate_public_partition_e2e(self):
         # Input collection has 100 elements, such that each privacy id
```
monsieurmuffin, line 179: nit: I recently learned that giving partition keys and simila elements meaningful names can help readability. E.g., we could call "pk101" something like "empty_public_partition" to make it more explicit how it's being used.

utility_analysis_new/tests/dp_engine_test.py
```diff
@@ -183,7 +211,8 @@ def test_aggregate_public_partition_e2e(self):
         self.assertLen(col, 3)
         self.assertTrue(any(v[0] == 'pk101' for v in col))
 
-    def test_aggregate_error_metrics(self):
+    @parameterized.parameters(False, True)
+    def test_aggregate_error_metrics(self, pre_aggregated: bool):
```
monsieurmuffin, line 215: nit: It's a bit difficult to guess what's being tested by looking at the test name.

utility_analysis_new/dp_engine.py
```diff
@@ -142,9 +148,47 @@ def _select_private_partitions_internal(
...
+                "Extract (partition_key, preaggregate_data))")
+        return super()._extract_columns(col, data_extractors)
+
+    def _check_aggregate_params(
+        self, col, params: pipeline_dp.AggregateParams,
+        data_extractors: Union[pipeline_dp.DataExtractors,
+                               utility_analysis_new.PreAggregateExtractors]):
+        # Do not check data_extractors. It is already checked in the base class
+        # does not support PreAggregateExtractors.
+        super()._check_aggregate_params(col,
```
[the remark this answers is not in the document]
dvadym: I don't like adding boolean here either, but having 2 methods is more complicated and less flexible (e.g. we can't control the order in which `DataExtractors` exception is called, only in the begining or in the end). Also I'd like to minimize changes in production code which are needed to be introduced with UtilityAnalysis implementation. So I'd prefer to keep this approach

dvadym (commented):
Thanks a lot for comments! I've addressed them. PTAL

utility_analysis_new/tests/dp_engine_test.py
```diff
@@ -106,29 +112,39 @@ def test_utility_analysis_params(self):
...
-                "public_partitions": [1]
-            },
-            {
-                "desc": "contribution bounds are already enforced",
-                "params": params_with_contribution_bounds_already_enforced,
-                "data_extractor": default_extractors,
-                "public_partitions": [1]
-            },
-        ]
+        test_cases = [{
```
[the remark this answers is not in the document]
monsieurmuffin: I see, thanks! I would vote for splitting this into several tests if some of the setup actions aren't necessary for all validations. But this seems unrelated to this PR.

utility_analysis_new/tests/dp_engine_test.py
monsieurmuffin: Thank you!

monsieurmuffin (approved):
Thank you!

PR #374, opened 2022-12-02
Rename compute_dataset_histograms

dvadym:
The function name is changed to represent more what function does. Along a way docstring was extended

### DIFF
--- a/utility_analysis_new/histograms.py
+++ b/utility_analysis_new/histograms.py
@@ -290,10 +290,18 @@ def _compute_partition_privacy_id_count_histogram(
         col, backend, HistogramType.COUNT_PRIVACY_ID_PER_PARTITION)
 
 
-def compute_contribution_histograms(
-        col, data_extractors: pipeline_dp.DataExtractors,
-        backend: pipeline_backend.PipelineBackend) -> DatasetHistograms:
-    """Computes privacy id contribution histograms."""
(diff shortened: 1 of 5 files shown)


utility_analysis_new/histograms.py
```diff
@@ -290,10 +290,18 @@ def _compute_partition_privacy_id_count_histogram(
         col, backend, HistogramType.COUNT_PRIVACY_ID_PER_PARTITION)
 
 
-def compute_contribution_histograms(
-        col, data_extractors: pipeline_dp.DataExtractors,
-        backend: pipeline_backend.PipelineBackend) -> DatasetHistograms:
-    """Computes privacy id contribution histograms."""
+def compute_dataset_histograms(col, data_extractors: pipeline_dp.DataExtractors,
+                               backend: pipeline_backend.PipelineBackend):
```
monsieurmuffin, line 294: nit: I find type annotations very helpful. We don't have one here because the method can return different types or is there any other reason?

monsieurmuffin (approved):

utility_analysis_new/histograms.py
dvadym: Yes, this method can return different types, namely it returns a collection which corresponds to pipeline framework used (PCollection for Beam, Iterable for Local run etc). Currenlty in PipelineDP for such arguments/output type annotations is not used.

PR #376, opened 2022-12-06
Refactor UtilityAnalysisEngine.analyze

dvadym:
This is a refactoring PR which contains the following changes.

1. Renaming `DPEngine.aggregate` -> `DPEngine.analyze` (this function performs analysis, not DP aggregation)
2. This function now takes `UtilityAnalysisOptions` instead of bunch of paramers

### DIFF
--- a/examples/restaurant_visits/run_without_frameworks_utility_analysis.py
+++ b/examples/restaurant_visits/run_without_frameworks_utility_analysis.py
@@ -112,10 +112,14 @@ def per_partition_utility_analysis():
     data_extractors = get_data_extractors()
     public_partitions = list(range(1, 8)) if FLAGS.public_partitions else None
 
-    result = utility_analysis_engine.analyze(
-        restaurant_visits_rows, aggregate_params, data_extractors,
-        public_partitions, get_multi_params(),
-        FLAGS.partition_sampling_probability)
+    options = utility_analysis_new.UtilityAnalysisOptions(
+        1,
+        1e-6,
+        aggregate_params,
+        multi_param_configuration=get_multi_params(),
+        partitions_sampling_prob=FLAGS.partition_sampling_probability)
+    result = utility_analysis_engine.analyze(restaurant_visits_rows, options,
+                                             data_extractors, public_partitions)
 
     budget_accountant.compute_budgets()
 
--- a/examples/restaurant_visits/run_without_frameworks_utility_analysis.py
+++ b/examples/restaurant_visits/run_without_frameworks_utility_analysis.py
@@ -113,9 +113,9 @@ def per_partition_utility_analysis():
     public_partitions = list(range(1, 8)) if FLAGS.public_partitions else None
 
     options = utility_analysis_new.UtilityAnalysisOptions(
-        1,
-        1e-6,
-        aggregate_params,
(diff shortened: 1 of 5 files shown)


examples/restaurant_visits/run_without_frameworks_utility_analysis.py
```diff
@@ -112,10 +112,14 @@ def per_partition_utility_analysis():
     data_extractors = get_data_extractors()
     public_partitions = list(range(1, 8)) if FLAGS.public_partitions else None
 
-    result = utility_analysis_engine.aggregate(
-        restaurant_visits_rows, aggregate_params, data_extractors,
-        public_partitions, get_multi_params(),
-        FLAGS.partition_sampling_probability)
+    options = utility_analysis_new.UtilityAnalysisOptions(
+        1,
```
monsieurmuffin, line 116: Is there a guidance to add comments for non-obvious parameter names in Python? 
In Java this would look like /* epsilon= */ 1. This helps readability.

utility_analysis_new/dp_engine.py
```diff
@@ -33,53 +33,42 @@ def __init__(self, budget_accountant: budget_accounting.BudgetAccountant,
...
             partition selection will be performed.
-          multi_param_configuration: if provided the utility analysis for
-            multiple parameters will be performed: 'params' is used as
-            blue-print and non-None attributes from 'multi_param_configuration'
-            are used for creating multiple AggregateParams. See docstring for
-            MultiParameterConfiguration for more details.
-          partitions_sampling_prob: todo
 
         Returns:
             A collection with elements (pk, utility analysis metrics).
```
monsieurmuffin, line 54: nit: I would write "partition_key" so that it's not confused with "privacy_key".

utility_analysis_new/dp_engine.py
```diff
@@ -33,53 +33,42 @@ def __init__(self, budget_accountant: budget_accounting.BudgetAccountant,
...
-        _check_utility_analysis_params(params, public_partitions)
+        _check_utility_analysis_params(options.aggregate_params,
+                                       public_partitions)
+        self._options = options
         self._is_public_partitions = public_partitions is not None
-        self._multi_run_configuration = multi_param_configuration
-        self._sampling_probability = partitions_sampling_prob
-        result = super().aggregate(col, params, data_extractors,
-                                   public_partitions)
+        result = super().aggregate(col, options.aggregate_params,
```
monsieurmuffin, line 60: I am slightly confused, does super().aggregate() run utility analysis or aggregation?

monsieurmuffin (approved):

utility_analysis_new/dp_engine.py
dvadym: Thanks! For docstrings of course we should use the whole name

examples/restaurant_visits/run_without_frameworks_utility_analysis.py
dvadym: Python has better solution, named arguments, i.e. (epsilon=1, delta = ...)
I've updated the code.

dvadym (commented):
Thanks a lot for review!

PR #379, opened 2022-12-19
Renaming utility_analysis_new folder to analysis

dvadym:
This PR contains the following changes:
1. `utility_analysis_new` directory was renamed to `analysis`
2. `analysis/dp_engine.py` was renamed to  `analysis/utility_analysis_engine.py`  (which is more correct, since this file contains `UtilityAnalysisEngine` class not `DPEngine`)


analysis/__init__.py
```diff
@@ -0,0 +1,28 @@
+# Copyright 2022 OpenMined.
```
dvadym, line 1: This is move, for some reason it's considered as deleted/added

monsieurmuffin (approved):
LG, thanks! Please could you consider making the request title a bit more specific (e.g., mention what has been renamed). Otherwise all renaming requests will look the same and it won't be easy to search for the specific changes in the history.

dvadym:
Thanks, I've renamed PR

PR #380, opened 2022-12-28
Implement cross-partition metrics for Sum

miracvbasaran:


analysis/combiners.py
```diff
@@ -644,6 +645,169 @@ def explain_computation(self):
...
+        abs_error_l0_expected = probability_to_keep * metrics.expected_cross_partition_error
+        abs_error_linf_expected = probability_to_keep * (
+            metrics.per_partition_error_min + metrics.per_partition_error_max)
+        abs_error_l0_variance = probability_to_keep * metrics.std_cross_partition_error**2
+        abs_error_variance = probability_to_keep * (
+            metrics.std_cross_partition_error**2 + metrics.std_noise**2)
+        loc_cpe_ne = metrics.expected_cross_partition_error
+        std_cpe_ne = math.sqrt(metrics.std_cross_partition_error**2 +
+                               metrics.std_noise**2)
+        abs_error_quantiles = []
```
dvadym, line 680: Computing quantiles is complicated, could you please extract it in a separate method/function (and it would be great to unify that with count computations)

analysis/combiners.py
```diff
@@ -644,6 +645,169 @@ def explain_computation(self):
...
+            rel_error_l0_expected=rel_error_l0_expected,
+            rel_error_linf_expected=rel_error_linf_expected,
+            rel_error_l0_variance=rel_error_l0_variance,
+            rel_error_variance=rel_error_variance,
+            rel_error_quantiles=rel_error_quantiles,
+            abs_error_expected_w_dropped_partitions=
+            abs_error_expected_w_dropped_partitions,
+            rel_error_expected_w_dropped_partitions=
+            rel_error_expected_w_dropped_partitions,
+            noise_std=noise_variance,
```
dvadym, line 746: this is incorrect -> variance set as std. Please fix also for count, where this problem also present

analysis/combiners.py
```diff
@@ -644,6 +645,169 @@ def explain_computation(self):
...
+                error / abs(metrics.sum) for error in abs_error_quantiles
+            ]
+            rel_error_expected_w_dropped_partitions = abs_error_expected_w_dropped_partitions / abs(
+                metrics.sum)
+
+        # Noise metrics
+        noise_variance = metrics.std_noise**2
+
+        return AggregateErrorMetricsAccumulator(
+            num_partitions=1,
```
dvadym, line 726: please add todo: to compute `num_partitions`

analysis/metrics.py
```diff
@@ -55,6 +59,10 @@ class SumMetrics:
       std_cross_partition_error: the standard deviation of the error due to cross-partition contribution bounding.
       std_noise: the noise standard deviation.
       noise_kind: the type of noise used.
+
+  s.t. the following holds:
+  Ex(sum_after_contribution_bounding) = sum + Ex(error)
```
dvadym, line 64: Ex -> `Expectation` or `E` (and then to note that E is expectation)
Ex is not much used as a shortcut for expectation

analysis/combiners.py
```diff
@@ -644,6 +645,169 @@ def explain_computation(self):
...
+class SumAggregateErrorMetricsCombiner(pipeline_dp.Combiner):
+    """A combiner for aggregating errors across partitions for Sum"""
+    AccumulatorType = AggregateErrorMetricsAccumulator
+
+    def __init__(self, error_quantiles: List[float]):
+        # The contribution bounding error is negative, so quantiles <0.5 for the
+        # error distribution (which is the sum of the noise and the contribution
+        # bounding error) should be used to come up with the worst error
+        # quantiles.
+        self._error_quantiles = [(1 - q) for q in error_quantiles]
```
dvadym, line 657: Nit: this comment is present twice (here and in count combiner). WDYT about extracting this conversion to a helper function and having the comment there?
Another option is to have a common base class for both combiners

analysis/combiners.py
```diff
@@ -644,6 +645,169 @@ def explain_computation(self):
...
+        total_aggregate = metrics.sum
+        # TODO: Compute data_dropped metrics
+        data_dropped_l0 = 0
+        data_dropped_linf = 0
+        data_dropped_partition_selection = 0
+
+        # Absolute error metrics
+        abs_error_l0_expected = probability_to_keep * metrics.expected_cross_partition_error
+        abs_error_linf_expected = probability_to_keep * (
+            metrics.per_partition_error_min + metrics.per_partition_error_max)
```
dvadym, line 673: AFAIU we don't have yet aggregation of per_partition_error_max and per_partition_error_min? PR is already very complicated, please leave TODO to add them

dvadym (commented):
Thanks!

analysis/combiners.py
miracvbasaran: WDYM exactly?

dvadym: Ah, sry, I thought that this is computing metrics and we don't compute num_partitions. I see, that this is creating an accumulator and of course it corresponds to 1 partition

analysis/utility_analysis_engine.py
```diff
@@ -105,17 +105,17 @@ def _create_compound_combiner(
                 internal_combiners.append(
                     utility_analysis_combiners.PartitionSelectionCombiner(
                         combiners.CombinerParams(budget, params)))
-            if pipeline_dp.Metrics.COUNT in aggregate_params.metrics:
+            if pipeline_dp.Metrics.SUM in aggregate_params.metrics:
```
dvadym, line 108: Just wondering, why the order is changed?

analysis/combiners.py
```diff
@@ -849,3 +802,37 @@ def metrics_names(self) -> List[str]:
 
     def explain_computation(self):
         pass
+
+
+def invert_error_quantiles(quantiles: List[float]) -> List[float]:
```
dvadym, line 807: Nit:  `_invert...` (to emphasize that this is private function for this file.

analysis/combiners.py
```diff
@@ -849,3 +802,37 @@ def metrics_names(self) -> List[str]:
...
+    # The contribution bounding error is negative, so quantiles <0.5 for the
+    # error distribution (which is the sum of the noise and the contribution
+    # bounding error) should be used to come up with the worst error
+    # quantiles.
+    return [(1 - q) for q in quantiles]
+
+
+def compute_error_quantiles(quantiles: List[float], prob_to_keep: float,
```
dvadym, line 815: Nit:  `_compute..` (to emphasize that this is private function for this file.

analysis/combiners.py
```diff
@@ -849,3 +802,37 @@ def metrics_names(self) -> List[str]:
...
+                            metric) -> List[float]:
```
dvadym, line 816: Can we add type annotations for `metric` ?

analysis/combiners.py
dvadym, line 815: Nit: add a docstring, like
"""Computes quantiles of per partition errors for the sum of DP noise and contribution bounding error.""""

analysis/combiners.py
```diff
@@ -849,3 +802,37 @@ def metrics_names(self) -> List[str]:
...
+    loc_cpe_ne = metric.expected_cross_partition_error
```
dvadym, line 817: Nit: loc_cpe_ne -> error_expectation

analysis/combiners.py
```diff
@@ -849,3 +802,37 @@ def metrics_names(self) -> List[str]:
...
+    std_cpe_ne = math.sqrt(metric.std_cross_partition_error**2 +
```
dvadym, line 818: -> error_std

dvadym (commented):
Thanks! Some minor comments

analysis/utility_analysis_engine.py
miracvbasaran: The order needs to be the same in both places (utility_analysis_engine.py and utility_analysis.py), so the order *is* important.

Given the order is important, I thought this order would make a bit of sense (going from the most complicated aggregation to least complicated). Similarly, the inverse would also make a bit of sense.

analysis/combiners.py
miracvbasaran: This function is overloaded, metric could be SumMetrics or CountMetrics.

If/when we merge Sum & Count, we could add type annotations here as well.

analysis/utility_analysis_engine.py
dvadym: Ok, SG, thanks, for explanations!

dvadym (approved):
Thanks for implementing Sum cross-partition metrics!

PR #381, opened 2023-01-03
Refactoring of Combiners, PrivacyIdCount and Count now use Sum

miracvbasaran:
both for per-partition and cross-partition aggregation.

### DIFF
--- a/analysis/combiners.py
+++ b/analysis/combiners.py
@@ -465,37 +465,42 @@ class SumAggregateErrorMetricsCombiner(pipeline_dp.Combiner):
         self._error_quantiles = self._invert_error_quantiles(error_quantiles)
 
     def create_accumulator(self,
-                           m: metrics.SumMetrics,
+                           partition_metrics: metrics.SumMetrics,
                            prob_to_keep: float = 1) -> AccumulatorType:
         """Creates an accumulator for metrics."""
         # Data drop ratio metrics
-        total_aggregate = m.sum
+        total_aggregate = partition_metrics.sum
         # TODO: Compute data_dropped metrics
         data_dropped_l0 = 0
         data_dropped_linf = 0
         data_dropped_partition_selection = 0
         if self._metric_type != metrics.AggregateMetricType.SUM:
-            data_dropped_l0 = -m.expected_cross_partition_error
-            data_dropped_linf = -m.per_partition_error_max
-            data_dropped_partition_selection = (
-                1 - prob_to_keep) * (m.sum + m.expected_cross_partition_error +
-                                     m.per_partition_error_max)
+            data_dropped_l0 = -partition_metrics.expected_cross_partition_error
+            data_dropped_linf = -partition_metrics.per_partition_error_max
+            data_dropped_partition_selection = (1 - prob_to_keep) * (
+                partition_metrics.sum +
+                partition_metrics.expected_cross_partition_error +
+                partition_metrics.per_partition_error_max)
 
         # Absolute error metrics
-        error_l0_expected = prob_to_keep * m.expected_cross_partition_error
-        error_linf_min_expected = prob_to_keep * m.per_partition_error_min
-        error_linf_max_expected = prob_to_keep * m.per_partition_error_max
+        error_l0_expected = prob_to_keep * partition_metrics.expected_cross_partition_error
+        error_linf_min_expected = prob_to_keep * partition_metrics.per_partition_error_min
+        error_linf_max_expected = prob_to_keep * partition_metrics.per_partition_error_max
         error_linf_expected = error_linf_min_expected + error_linf_max_expected
-        error_l0_variance = prob_to_keep * m.std_cross_partition_error**2
-        error_variance = prob_to_keep * (m.std_cross_partition_error**2 +
-                                         m.std_noise**2)
-        error_quantiles = self._compute_error_quantiles(prob_to_keep, m)
+        error_l0_variance = prob_to_keep * partition_metrics.std_cross_partition_error**2
+        error_variance = prob_to_keep * (
+            partition_metrics.std_cross_partition_error**2 +
+            partition_metrics.std_noise**2)
+        error_quantiles = self._compute_error_quantiles(prob_to_keep,
+                                                        partition_metrics)
         error_expected_w_dropped_partitions = prob_to_keep * (
-            m.expected_cross_partition_error + m.per_partition_error_min +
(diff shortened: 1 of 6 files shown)


analysis/tests/combiners_test.py
```diff
@@ -1117,13 +1216,167 @@ def test_create_accumulator(self, probability_to_keep, metric, expected):
...
 
         # Test that no type is np.float64
         self.assertTrue(_check_none_are_np_float64(acc))
 
 
+def assert_aggregate_error_metrics_accumulator_equals(
+        test: parameterized.TestCase,
+        expected: combiners.AggregateErrorMetricsAccumulator,
+        actual: combiners.AggregateErrorMetricsAccumulator):
+    test.assertEqual(expected.num_partitions, actual.num_partitions)
```
dvadym, line 1232: It feels that using reflection would make things more manageble :), basically we need to compare all fields and there are 3 options:
1.  enum attributes
2.  double attributes
3.  list attributes

What do you think about writing a helper function, which take 2 dataclasses of the same type and do comparison? 
For getting fields you can use [dataclasses.fields](https://docs.python.org/3/library/dataclasses.html#dataclasses.fields) or [dataclasses.asdict](https://docs.python.org/3/library/dataclasses.html#dataclasses.fields) or other methods from `dataclasses` module.

This helper function can be useful in other tests as well.

analysis/tests/combiners_test.py
```diff
@@ -1117,13 +1216,167 @@ def test_create_accumulator(self, probability_to_keep, metric, expected):
...
+                            delta=1e-5)
+    test.assertAlmostEquals(expected.error_variance,
+                            actual.error_variance,
+                            delta=1e-5)
+    test.assertEqual(len(expected.error_quantiles), len(actual.error_quantiles))
+    [
+        test.assertAlmostEqual(expected.error_quantiles[i],
+                               actual.error_quantiles[i],
+                               delta=1e-5)
+        for i in range(len(expected.error_quantiles))
```
dvadym, line 1265: Nit: Pythonic style is using zip instead of indices:

`for exp, act in zip(expected.error_quantiles, actual.error_quantiles):`

analysis/combiners.py
```diff
@@ -492,190 +465,63 @@ def __init__(self, metric_type: metrics.AggregateMetricType,
...
-        self._error_quantiles = _invert_error_quantiles(error_quantiles)
-
-    def create_accumulator(self,
-                           metrics: metrics.SumMetrics,
+                           m: metrics.SumMetrics,
                            prob_to_keep: float = 1) -> AccumulatorType:
         """Creates an accumulator for metrics."""
         # Data drop ratio metrics
-        total_aggregate = metrics.sum
+        total_aggregate = m.sum
```
dvadym, line 472: Nit: `m` is too short in that long context of the whole function, for 1-2 line context it might be fine. I see that the problem is that metrics is the name of the module. What do you think about m -> `partition_metrics` ?

analysis/metrics.py
```diff
@@ -20,37 +20,12 @@
...
-    std_cross_partition_error: float
-    std_noise: float
-    noise_kind: pipeline_dp.NoiseKind
-
-
 @dataclass
 class SumMetrics:
     """Stores metrics for the sum utility analysis.
 
+  Also used for count & privacy_id_count.
```
dvadym, line 27: Nit: It is also used for COUNT and PRIVACY_ID_COUNT per partition metrics.

analysis/metrics.py
```diff
@@ -20,37 +20,12 @@
...
-    per_partition_error: int
-    expected_cross_partition_error: float
-    std_cross_partition_error: float
-    std_noise: float
-    noise_kind: pipeline_dp.NoiseKind
-
-
 @dataclass
 class SumMetrics:
     """Stores metrics for the sum utility analysis.
```
dvadym, line 25: Nit: ... analysis **per partition.**

analysis/metrics.py
```diff
@@ -20,37 +20,12 @@
...
 
+  Also used for count & privacy_id_count.
+
```
dvadym, line 28: Nit: in  the docstring in this and other dataclasses in this file the spacing is weird. Could you please fix it?
```
class ...
        """" # (4 spaces before """")
        text
        """" 
```

analysis/metrics.py
```diff
@@ -98,12 +73,16 @@ class AggregateErrorMetrics:
```
dvadym, line 73: Please extend the docstring with ephasizing that this cross-partition aggregate metrics

dvadym (commented):
Thanks a lot! I left some comments

analysis/tests/combiners_test.py
```diff
@@ -822,308 +970,259 @@ def test_create_accumulator(self, metric_type, probability_to_keep, metric,
...
         # Test that no type is np.float64
         self.assertTrue(_check_none_are_np_float64(acc))
 
 
+def assert_dataclasses_are_equal(test, expected, actual):
+    test.assertEquals(type(expected), type(actual))
+    expected = dataclasses.asdict(expected)
+    actual = dataclasses.asdict(actual)
+
+    for (f1, exp), (f2, act) in zip(expected.items(), actual.items()):
```
dvadym, line 1197: Are there guarantees that `expected.items()` are `actual.items()` are in the order? It feels that it is not (both are hash dictionaries, and hash function might be in principle chosen randomly). I'd suggest something like

```
for field_name in actual.keys():
  expected_value = expected[...]
  actual = actual[...]
```

analysis/tests/combiners_test.py
```diff
@@ -822,308 +970,259 @@ def test_create_accumulator(self, metric_type, probability_to_keep, metric,
...
+            metric_type, [0.5])
         metric = combiner.compute_metrics(acc)
-        self.assertEqual(expected, metric)
+        assert_dataclasses_are_equal(self, expected, metric)
 
         # Test that no type is np.float64
         self.assertTrue(_check_none_are_np_float64(acc))
 
 
+def assert_dataclasses_are_equal(test, expected, actual):
```
dvadym, line 1192: Nice, using of "reflection"!

dvadym, line 1192: Nit: **_**assert_data...

dvadym, line 1192: Nit: add 1 line docstring for this function

dvadym (approved):
Thanks!

PR #383, opened 2023-01-05
Refactoring: tests comparing dataclasses now rely on the common function

miracvbasaran:


analysis/tests/utility_analysis_test.py
```diff
@@ -82,98 +84,103 @@ def test_wo_public_partitions(self, pre_aggregated: bool):
...
-            2.06410,
-            delta=1e-5)
+        expected_partition_selection_metrics = metrics.PartitionSelectionMetrics(
+            num_partitions=10,
+            dropped_partitions_expected=7.08783,
+            dropped_partitions_variance=2.06410)
+        common._assert_dataclasses_are_equal(
+            self, expected_partition_selection_metrics,
+            output.partition_selection_metrics)
         # Assert count metrics are reasonable.
```
dvadym, line 94: Nice improvement of readability! Thanks!

analysis/tests/common.py
```diff
@@ -0,0 +1,73 @@
...
+"""Common functions for utility analysis tests"""
+
+import dataclasses
+import enum
+import typing
+
+from absl.testing import parameterized
+
+
+def _assert_dataclasses_are_equal(test: parameterized.TestCase,
```
dvadym, line 23: Nit: remove _ in the name (it's not private function in this file anymore)

dvadym (commented):
Thanks!

analysis/tests/common.py
miracvbasaran: Good point, thanks!

dvadym (approved):

PR #385, opened 2023-01-23
Fix split budget for multi configuration in Utility Analysis

dvadym:
##What was broken?

For simplicity assume that `metrics=[count, sum]` and there are 2 configurations to compute - `max_partition_contributed = [1,2]`.

For computing each metric for each input configuration `UtilityAnalysisCombiner` is created, before this PR the code was the following

```
for metric in metrics:
  for configuration in configurations:
      budget = request_budget()
      // create  combiner
```

The problem is that it requests budget 2*2 = 4 times. Which is incorrect, since different configurations have different budget. This PR fixes that by ensuring that for each metric independently of the number of configurations `request_budget` is called once.

### DIFF
--- a/analysis/tests/utility_analysis_engine_test.py
+++ b/analysis/tests/utility_analysis_engine_test.py
@@ -330,14 +330,14 @@ class UtilityAnalysisEngineTest(parameterized.TestCase):
                                per_partition_error_max=0.0,
                                expected_cross_partition_error=-0.5,
                                std_cross_partition_error=0.5,
-                               std_noise=11.6640625,
+                               std_noise=5.87109375,
                                noise_kind=pipeline_dp.NoiseKind.GAUSSIAN),
             metrics.SumMetrics(sum=1.0,
                                per_partition_error_min=0.0,
                                per_partition_error_max=0.0,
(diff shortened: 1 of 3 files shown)


analysis/tests/utility_analysis_engine_test.py
```diff
@@ -346,19 +346,21 @@ def test_multi_parameters(self):
                                per_partition_error_max=-1.0,
                                expected_cross_partition_error=-0.5,
                                std_cross_partition_error=0.5,
-                               std_noise=11.6640625,
+                               std_noise=5.87109375,
```
monsieurmuffin, line 349: Is this change here because we are now using twice more budget?
Please could you add a comment mentioning the epsilon and delta parameters for which the STD was computed?

monsieurmuffin (approved):

analysis/tests/utility_analysis_engine_test.py
dvadym: Yes, that's because of having 2 times more budget. I checked that in Colab and it aligned (C++ implementation has relative precision 1e-3)

dvadym:
Thanks for review!

PR #388, opened 2023-01-24
Speedup of utility combiners

dvadym:
This PR speeds up utility analysis per partition for multi set of input configurations by vectorizing some operations with Numpy, 

namely before this PR
Each combiner gets `List[(count:int, sum:float, n_partitions:int)]` and all processing is in Python

Now each combiner gets
2. `(counts:np.ndarray, sums:np.ndarray, n_partitions:np.ndarray)` and all processing is in Numpy, which is much faster.


analysis/combiners.py
```diff
@@ -160,14 +161,27 @@ def _compute_pmf(self) -> poisson_binomial.PMF:
                                       Optional[SumOfRandomVariablesMoments]]
 
 
+def _combine_list(a: List, b: List) -> List:
+    """Combiners 2 lists in 1 and returns it.
```
miracvbasaran, line 165: nit: `Combines`

analysis/combiners.py
```diff
@@ -160,14 +161,27 @@ def _compute_pmf(self) -> poisson_binomial.PMF:
...
+
+    Warning: it changes arguments.
```
miracvbasaran, line 167: supernit: `modifies` instead of `changes`

analysis/combiners.py
```diff
@@ -160,14 +161,27 @@ def _compute_pmf(self) -> poisson_binomial.PMF:
                                       Optional[SumOfRandomVariablesMoments]]
 
 
+def _combine_list(a: List, b: List) -> List:
```
miracvbasaran, line 164: Optional: WDYT of `merge` instead of `combine`? That is what we use to describe the same operation for DP Building Blocks primitives.

analysis/combiners.py
```diff
@@ -160,14 +161,27 @@ def _compute_pmf(self) -> poisson_binomial.PMF:
...
+    """Combiners 2 lists in 1 and returns it.
+
+    Warning: it changes arguments.
+    """
+    # Extend the larger list (for performance reasons).
```
miracvbasaran, line 169: supernit: I would remove the paragraphs since the performance reasons seem to be the key to this whole function

analysis/combiners.py
```diff
@@ -183,15 +197,15 @@ class PartitionSelectionCombiner(UtilityAnalysisCombiner):
...
         max_partitions = self._params.aggregate_params.max_partitions_contributed
-        prob_keep_contribution = min(1, max_partitions /
-                                     n_partitions) if n_partitions > 0 else 0
-
-        return ((prob_keep_contribution,), None)
+        prob_keep_partition = np.where(
+            n_partitions > 0, np.minimum(1, max_partitions / n_partitions), 0)
+        acc1 = (list(prob_keep_partition), None)
+        acc2 = ([], None)
+        return _merge_partition_selection_accumulators(acc1, acc2)
```
miracvbasaran, line 208: Why is the call to _merge_partition_selection_accumulators necessary?

analysis/combiners.py
```diff
@@ -216,31 +230,35 @@ class SumCombiner(UtilityAnalysisCombiner):
...
     def __init__(self, params: pipeline_dp.combiners.CombinerParams):
-        self._params = params
+        self._params = copy.copy(params)
 
-    def create_accumulator(self, data: PreaggregatedData) -> AccumulatorType:
+    def create_accumulator(
+            self, data: Tuple[np.ndarray, np.ndarray,
+                              np.ndarray]) -> AccumulatorType:
         count, partition_sum, n_partitions = data
+        del count  # not used for SumCombiner
```
miracvbasaran, line 239: Can't we do `_, partition_sum, n_partitions = data` instead in Python? (also in other places)

analysis/combiners.py
```diff
@@ -263,12 +281,32 @@ class CountCombiner(SumCombiner):
     # expected_cross_partition_error, var_cross_partition_error)
     AccumulatorType = Tuple[float, float, float, float, float]
 
-    def create_accumulator(self, data: PreaggregatedData) -> AccumulatorType:
-        count, _sum, n_partitions = data
-        data = count, count, n_partitions
-        self._params.aggregate_params.min_sum_per_partition = 0.0
-        self._params.aggregate_params.max_sum_per_partition = self._params.aggregate_params.max_contributions_per_partition
-        return super().create_accumulator(data)
+    def create_accumulator(
```
miracvbasaran, line 284: It seems we are duplicating logic again. Why don't we rely on super().create_accumulator()?

analysis/combiners.py
```diff
@@ -300,45 +340,46 @@ class CompoundCombiner(pipeline_dp.combiners.CompoundCombiner):
...
                             Optional[DenseAccumulatorType]]
 
     def create_accumulator(self, data: PreaggregatedData) -> AccumulatorType:
         if not data:
-            # This is an empty partition, what is None assigned to. It was added
-            # because of public partitions.
-            return ([(0, 0, 0)], None)
-        return ([data], None)
+            # This is an empty partition. It was added because of public
+            # partitions.
```
miracvbasaran, line 351: nit: Maybe something like  `# Handle empty partitions. Only encountered when public partitions are used.` could be better here

analysis/combiners.py
```diff
@@ -216,31 +230,35 @@ class SumCombiner(UtilityAnalysisCombiner):
     AccumulatorType = Tuple[float, float, float, float, float]
 
     def __init__(self, params: pipeline_dp.combiners.CombinerParams):
-        self._params = params
+        self._params = copy.copy(params)
 
-    def create_accumulator(self, data: PreaggregatedData) -> AccumulatorType:
+    def create_accumulator(
```
miracvbasaran, line 235: Maybe I am missing something but I don't see how using np.ndarrays help speed up here. We only have a single value inside the np.ndarrays, no? So how is it faster than simply doing the arithmetic with the int/float primitives?

analysis/combiners.py
dvadym: Thanks a lot! This change in this PR was made before your refactoring )), of course we should use super().create_accumulator(), done

dvadym (commented):
Thanks for review! I've addressed comments. PTAL

analysis/combiners.py
miracvbasaran: That makes sense!

Could you please add some documentation about this, for example to the CompoundCombiner's explanation of the dense-sparse optimization?

analysis/combiners.py
miracvbasaran: I see. IMHO when you do `_, partition_sum, n_partitions = data` you don't even think about `count`, so there is a smaller cognitive load while reading the code; so I would prefer that. But I definitely see the benefit of being explicit. So, no strong opinions.

miracvbasaran (approved):

analysis/combiners.py
dvadym: Good point, done!

dvadym (commented):
Thanks for the review!

PR #389, opened 2023-01-24
Fix docstring of the output argument for tuning

dvadym:


analysis/parameter_tuning.py
```diff
@@ -176,7 +176,7 @@ def tune(col,
          data_extractors: Union[pipeline_dp.DataExtractors,
                                 analysis.PreAggregateExtractors],
          public_partitions=None,
-         return_utility_analysis_per_partition: bool = False) -> TuneResult:
```
miracvbasaran, line 179: Could we do a `Union` instead? Or something else in Python that tells this is "either/or"?

miracvbasaran (approved):

analysis/parameter_tuning.py
dvadym: No, it was mistaken type annotation. The result is not TuneResult but a collection with 1 element of TuneResult

dvadym (commented):
Thanks for review!

PR #395, opened 2023-02-08
Introducing public_partitions_already_filtered

dvadym:


monsieurmuffin (commented):
Thanks! Please could you add a description of the PR adding a bit more details on what the flag is doing and why we need it?

pipeline_dp/aggregate_params.py
```diff
@@ -131,6 +131,9 @@ class AggregateParams:
          max_contributions_per_partition. This option can be used if the dataset
          does not contain any identifiers that can be used to enforce
          contribution bounds automatically.
+        partitions_already_filtered: if true, it is assumed that partitions are
```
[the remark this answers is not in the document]
dvadym: Thanks for comments! I've addressed them. PTAL

examples/movie_view_ratings/run_all_frameworks.py
```diff
@@ -83,6 +83,7 @@ def calc_dp_rating_metrics(movie_views, backend, public_partitions):
         max_contributions_per_partition=1,
         min_value=1,
         max_value=5,
+        partitions_already_filtered=True,
```
monsieurmuffin, line 86: nit: should this be public_partitions_already_filtered=True ?

pipeline_dp/aggregate_params.py
```diff
@@ -131,6 +131,9 @@ class AggregateParams:
          max_contributions_per_partition. This option can be used if the dataset
          does not contain any identifiers that can be used to enforce
          contribution bounds automatically.
+        public_partitions_already_filtered: if true, it is assumed that records
+         with partition_key which are not in public_partitions are already
+         removed from the dataset. It applies only for public partitions.
```
monsieurmuffin, line 136: "It applies only for public partitions." -> "The flag can only be used with public partitions"

monsieurmuffin (approved):

pipeline_dp/aggregate_params.py
dvadym: Thanks

dvadym (commented):
Thanks for review!

PR #400, opened 2023-02-15
Refactor analysis output metrics: part 1 Adding new metrics

dvadym:
Filling this dataclasses will be done in the following PRs.

### DIFF
--- a/analysis/metrics.py
+++ b/analysis/metrics.py
@@ -48,6 +48,100 @@ class SumMetrics:
     noise_kind: pipeline_dp.NoiseKind
 
 
+class AggregateMetricType(Enum):
+    PRIVACY_ID_COUNT = 'privacy_id_count'
+    COUNT = 'count'
+    SUM = 'sum'
+
+
+@dataclass
+class AggregateErrorMetrics:
+    """Stores aggregate cross-partition metrics for utility analysis.
+    All attributes in this dataclass are averages across partitions; except for
+    ratio_* attributes, which are simply the ratios of total data dropped
+    aggregated across partitions.
+    """
+    metric_type: AggregateMetricType
+
+    ratio_data_dropped_l0: float
+    ratio_data_dropped_linf: float
+    # This cannot be computed at PartitionSelectionMetrics and needs to be
+    # computed for each aggregation separately, since it takes into account data
+    # drop from contribution bounding and that is aggregation-specific.
+    ratio_data_dropped_partition_selection: float
+
+    error_l0_expected: float
+    error_linf_expected: float
+    error_linf_min_expected: float
+    error_linf_max_expected: float
+    error_expected: float
+    error_l0_variance: float
+    error_variance: float
+    error_quantiles: List[float]
+    rel_error_l0_expected: float
+    rel_error_linf_expected: float
+    rel_error_linf_min_expected: float
+    rel_error_linf_max_expected: float
+    rel_error_expected: float
+    rel_error_l0_variance: float
+    rel_error_variance: float
+    rel_error_quantiles: List[float]
+
+    # The following error metrics include error from dropped partitions.
+    #
+    # Consider the following example with a single partition to see how they are
+    # different from abs/rel_error_expected metrics:
+    #
+    # Given 1 partition with probability_to_keep=0.4, actual_count=100,
+    # abs_error_expected=-50;
+    # -> abs_error_expected = (0.4*-50)/0.4=-50
+    # -> abs_error_expected_w_dropped_partitions = 0.4*-50+0.6*-100=-80
+    #
+    # When public partitions are used, these will be exactly equal to
+    # abs/rel_error_expected.
+    error_expected_w_dropped_partitions: float
+    rel_error_expected_w_dropped_partitions: float
+
+    noise_std: float
+
+    # RMSE = sqrt(bias**2 + variance), more details in
+    # https://en.wikipedia.org/wiki/Bias-variance_tradeoff.
+    def absolute_rmse(self) -> float:
+        return math.sqrt(self.error_expected**2 + self.error_variance)
+
+    def relative_rmse(self) -> float:
+        return math.sqrt(self.rel_error_expected**2 + self.rel_error_variance)
+
+
+@dataclass
+class PartitionSelectionMetrics:
+    """Stores aggregate metrics about partition selection."""
+
+    num_partitions: float
+    dropped_partitions_expected: float
+    dropped_partitions_variance: float
+
+
+@dataclass
+class AggregateMetrics:
+    """Stores result of the utility analysis for specific input parameters.
+    Attributes:
+        input_aggregate_params: input parameters for which this utility analysis
+          was computed.
+        count_metrics: utility analysis of count. It is non None, if Count
+          metric is in input_aggregate_params.metrics.
+        privacy_id_count_metrics: utility analysis of sum. It is non None, if
+          Sum  metric is in input_aggregate_params.metrics.
+        partition_selection_metrics: utility analysis of selected partition. It
+          is not None if the utility analysis is for private partition selection.
+    """
+    input_aggregate_params: pipeline_dp.AggregateParams
+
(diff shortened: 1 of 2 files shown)


analysis/metrics.py
```diff
@@ -103,24 +124,15 @@ class AggregateErrorMetrics:
     error_expected_w_dropped_partitions: float
     rel_error_expected_w_dropped_partitions: float
```
miracvbasaran, line 125: I think we should move this to MeanValueErrors dataclass

dvadym (commented):

analysis/metrics.py
```diff
@@ -143,3 +144,157 @@ class AggregateMetrics:
...
+
+
+@dataclass
+class UtilityAnalysisResult:
+    """Stores result of the utility analysis for specific input parameters.
+
+    Attributes:
+        input_aggregate_params: input parameters for which this utility analysis
+          was computed.
+        metric_errors: utility analysis of metrics (e.g. COUNT, SUM, P
```
miracvbasaran, line 293: The line break here seems off, can't `P` and `RIVACY_ID_COUNT` be on the same line?

analysis/metrics.py
```diff
@@ -143,3 +144,157 @@ class AggregateMetrics:
...
+    # -> abs_error_expected = (0.4*-50)/0.4=-50
+    # -> abs_error_expected_w_dropped_partitions = 0.4*-50+0.6*-100=-80
+    #
+    # When public partitions are used, these will be exactly equal to
+    # abs/rel_error_expected.
+    with_dropped_partitions: float
+
+
+@dataclass
+class DroppedDataInfo:
```
miracvbasaran, line 223: Optional: `DataDropInfo`

analysis/metrics.py
```diff
@@ -143,3 +144,157 @@ class AggregateMetrics:
...
+    """Information about the data dropped during different procedures.
```
miracvbasaran, line 224: Optional: `different DP stages`

analysis/metrics.py
```diff
@@ -143,3 +144,157 @@ class AggregateMetrics:
...
+        metric: DP metric for which this analysis was performed.
+        num_dataset_partitions: the number of partitions in dataset.
+        num_non_public_partitions: the number of partitions dropped because
+          of public partitions.
+        num_empty_partitions: the number of empty partitions added because of
+          public partitions.
+        noise_std: the standard deviation of added noise.
+        noise_kind: the noise kind (Laplace or Gaussian)
+        ratio_data_dropped: the information about dropped data.
+        absolute_error: error of (dp_value - actual_value).
```
miracvbasaran, line 256: Optional: `error in terms of (dp_value - actual_value)`

(also below)

analysis/metrics.py
```diff
@@ -143,3 +144,157 @@ class AggregateMetrics:
...
+    linf: float
+    linf_min: float
+    linf_max: float
+
+
+@dataclass
+class ValueErrors:
+    """Errors between true and dp metric.
+
+    This class describes breakdown of errors for (dp_value - actual_value),
```
miracvbasaran, line 182: nit: Use only `true` or `actual` for consistency (see the line above). I don't have a preference on one over the other, but I think we should have consistency to avoid confusion and allow easier readability.

analysis/metrics.py
```diff
@@ -143,3 +144,157 @@ class AggregateMetrics:
...
+      rmse_per_partition = sqrt(E(dp_value - actual_value)^2)
+      self.rmse = mean(rmse_per_partition)
+
+    Attributes:
+        bounding_errors: contribution bounding error.
+        bias: averaged across partitions E(dp_value - actual_value).
+        variance: averaged across partitions Var(dp_value - actual_value).
+        rmse: averaged across partitions sqrt(E(dp_value - actual_value)^2).
+        l1:  averaged across partitions E|dp_value - actual_value|.
+        with_dropped_partitions: error which takes into consideration dropped
```
miracvbasaran, line 198: This sentence seems a bit off, maybe "error which takes into consideration partitions dropped due to partition selection"?

analysis/metrics.py
```diff
@@ -143,3 +144,157 @@ class AggregateMetrics:
...
+    var: float
+
+
+@dataclass
+class ContributionBoundingErrors:
+    """Contains an error breakdown by types of contribution bounding.
+
+    Attributes:
+        l0: max_partition_contributed (aka l0) bounding error. The output of l0
+          bounding is a random variable for each partition.
```
miracvbasaran, line 165: This one we know to be a Normally distributed when number of partitions is large, is it worth mentioning that?

analysis/metrics.py
```diff
@@ -143,3 +144,157 @@ class AggregateMetrics:
...
+        linf: per partition (aka linf) bounding errors. The output of linf
```
miracvbasaran, line 166: nit: `error` instead of `errors` since you talk about min & max separately below.

analysis/metrics.py
```diff
@@ -143,3 +144,157 @@ class AggregateMetrics:
...
+    variable and it comes from different sources - contribution bounding error
+    and DP noise. This class contains different error metrics.
+
+    All attributes correspond to the errors computed per partition and then
+    averaged across partitions, e.g.
+      rmse_per_partition = sqrt(E(dp_value - actual_value)^2)
+      self.rmse = mean(rmse_per_partition)
+
+    Attributes:
+        bounding_errors: contribution bounding error.
```
miracvbasaran, line 193: nit: `contribution bounding errors`

analysis/metrics.py
```diff
@@ -143,3 +144,157 @@ class AggregateMetrics:
...
+@dataclass
+class ValueErrors:
+    """Errors between true and dp metric.
+
+    This class describes breakdown of errors for (dp_value - actual_value),
+    where value can be a metric like count, sum etc. The value error is a random
+    variable and it comes from different sources - contribution bounding error
+    and DP noise. This class contains different error metrics.
+
+    All attributes correspond to the errors computed per partition and then
```
miracvbasaran, line 187: Does this apply to bounding_errors as well?

analysis/metrics.py
```diff
@@ -143,3 +144,157 @@ class AggregateMetrics:
...
+    """Stores aggregate metrics about partition selection."""
+
+    strategy: pipeline_dp.PartitionSelectionStrategy
+    num_partitions: float
+    dropped_partitions: MeanVariance
+    ratio_dropped_data: float
+
+
+@dataclass
+class UtilityAnalysisResult:
```
miracvbasaran, line 287: Not sure if I mentioned this before but I suggest renaming this to `UtilityReport`:
1) It conveys the same meaning but significantly shorter
2) "Utility Analysis" is an overloaded term
3) It gives a nice "this is the ultimate end result" type of feeling

analysis/metrics.py
dvadym: Yes

dvadym (commented):
Thanks for the comments! I've addressed them. PTAL

analysis/metrics.py
```diff
@@ -143,3 +144,157 @@ class AggregateMetrics:
...
+    # Dropped data breakdown.
+    ratio_data_dropped: DroppedDataInfo
+
+    # Value errors
+    absolute_error: ValueErrors
+    relative_error: ValueErrors
+
+
+@dataclass
+class PartitionSelectionUtility:
```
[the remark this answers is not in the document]
miracvbasaran: I'd prefer if we can make `PartitionSelectionUtility` about partition sselection in general, including both private and public one. But if it's not feasible, then renaming to `PrivatePartitionSelectionUtility` SG

miracvbasaran (approved):

analysis/metrics.py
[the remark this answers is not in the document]
dvadym: Thanks, replaced with `PrivatePartitionSelectionUtility`

dvadym (commented):
Thanks for review!

PR #401, opened 2023-02-17
Fix handling stage names in private_beam

dvadym:
Apache Beam requires that all stage_names were different. `pipeline_dp.BeamBackend` object ensures the uniqueness of the Beam stage names. But it is required to have the same object `BeamBackend` for all operations in private_beam.py.

This PR implements does by introducing a module level variable `_beam_backend` for this.


monsieurmuffin (approved):

dvadym:
Thanks for review!

Issue #259, opened 2022-04-06
Advance composition in PipelineDP


dvadym:
# Short DP references
**Definition:** Basic (or naive) composition of differential privacy mechanisms with parameters `(eps1, delta1)` and `(eps2, delta2)` is a mechanism with parameter `(eps1+eps2, delta1+delta2)`. 

There are many advance composition methods ([RDP](https://arxiv.org/abs/1702.07476), [CDP](https://arxiv.org/pdf/1605.02065.pdf), [Privacy Loss Distributions (PLD)](https://eprint.iacr.org/2018/820.pdf) etc), which allow to get total budget smaller than with basic composition. 

There is [dp_accounting](https://github.com/google/differential-privacy/tree/main/python/dp_accounting) library developed by Google, which implements PLD and RDP methods.

# PipelineDP budget accounting design
In the PipelineDP initial release only Basic DP composition was supported, but the budget accounting was designed with keeping in mind introduction of other types of composition later.

The basic budget accounting is performed by class [NaiveBudgetAccountant](https://github.com/OpenMined/PipelineDP/blob/229a24033265182b8e8ee247a3596012d552aeb8/pipeline_dp/budget_accounting.py#L200).

There is already class  [PLDBudgetAccountant]( https://github.com/OpenMined/PipelineDP/blob/229a24033265182b8e8ee247a3596012d552aeb8/pipeline_dp/budget_accounting.py#L298), but it's not fully supported by PipelinedP.

## How BudgetAccountant is used
BudgetAccountant is used in the following steps:
1. Specify total budget
2. Specify DP mechanisms by requesting budget for them
3. Compute budget per mechanism
4. Use budget per mechanism.

To be specific the following is an example of `NaiveBudgetAccountant` usage:
```
    #  Creating budget accountant with specified total budet.
    budget_accountant = NaiveBudgetAccountant(total_epsilon=1, total_delta=1e-6) 
    # Request budget for 2 mechanisms - laplace and gaussian.
    budget1 = budget_accountant.request_budget(mechanism_type=MechanismType.LAPLACE)
    budget2 = budget_accountant.request_budget(
        mechanism_type=MechanismType.GAUSSIAN, weight=3)
    # Having all mechanisms it computes the budget per mechanism.
    budget_accountant.compute_budgets()
  
    # As a result the budget (1, 1e-6) splits over 2 mechanism, with weight 1 and 3:
    # budget1.eps = 0.25, budget1.delta = 0  
[...]

PR #405, opened 2023-02-24
(WIP) Support to PLD composition #259

ricardocarvalhods:
## Description
Solving "Advance composition in PipelineDP" #259 

The approach I'm using is: 
- Propagates noise_standard_deviation (in the case of NaiveAccountant, it is None throughout)
- When it reaches apply_laplace_mechanism/apply_gaussian_mechanism, if noise_standard_deviation is NOT None, we use it for noise (ignoring eps/delta)
- Otherwise (if noise_standard_deviation is None), eps/delta are used to add noise.

Tasks:
- [x] Add noise_standard_deviation to **scalar** computations and parameters
- [x] Ajust tests to pass after changes
- [ ] Create tests for **scalar** computations with noise_standard_deviation
- [ ] Same as above, just to **vector** sum computation

Additionally:
- I've also did some minor changes, like two TODOs in the `dp_computations.py` and formatting.
- I see that `request_budget` receives `noise_standard_deviation`, do you know why? not sure if it is needed, as `noise_standard_deviation` is only calculated in compute_budgets.

### DIFF
--- a/pipeline_dp/dp_computations.py
+++ b/pipeline_dp/dp_computations.py
@@ -33,6 +33,7 @@ class ScalarNoiseParams:
     max_partitions_contributed: int
     max_contributions_per_partition: Optional[int]
     noise_kind: pipeline_dp.NoiseKind  # Laplace or Gaussian
+    noise_standard_deviation: Optional[float] = None
 
     def __post_init__(self):
         assert (self.min_value is None) == (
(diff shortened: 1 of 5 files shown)


pipeline_dp/dp_computations.py
```diff
@@ -103,50 +104,70 @@ def compute_sigma(eps: float, delta: float, l2_sensitivity: float):
         delta: The delta value.
         l2_sensitivity: The L2 sensitivity.
     """
-    # TODO: use named arguments, when argument names are added in PyDP on PR
```
dvadym, line 106: Thanks for fixing this TODO

pipeline_dp/dp_computations.py
```diff
@@ -103,50 +104,70 @@ def compute_sigma(eps: float, delta: float, l2_sensitivity: float):
...
         l1_sensitivity: The L1 sensitivity.
+        noise_standard_deviation: The standard deviation for the noise.
 
     Returns:
         The value resulted after adding the noise.
     """
-    mechanism = dp_mechanisms.LaplaceMechanism(epsilon=eps,
-                                               sensitivity=l1_sensitivity)
+    if noise_standard_deviation is not None:
+        mechanism = dp_mechanisms.LaplaceMechanism(epsilon=1 /
```
dvadym, line 129: Please add comment that here workaround is used, since we can't set the Laplace parameter directly

tests/budget_accounting_test.py
```diff
@@ -198,9 +198,8 @@ def test_not_enough_aggregations(self, use_num_aggregations):
 class PLDBudgetAccountantTest(unittest.TestCase):
 
     def test_noise_not_calculated(self):
-        with self.assertRaises(AssertionError):
-            mechanism = MechanismSpec(MechanismType.LAPLACE)
-            print(mechanism.noise_standard_deviation())
+        mechanism = MechanismSpec(MechanismType.LAPLACE)
+        self.assertEqual(None, mechanism.noise_standard_deviation)
```
dvadym, line 202: Nit: self.assertNone(

pipeline_dp/dp_computations.py
```diff
@@ -370,10 +403,18 @@ def compute_dp_mean(count: int, normalized_sum: float,
         dp_params.eps, dp_params.delta, 2)
     l0_sensitivity = dp_params.l0_sensitivity()
 
+    # Increases noise std.dev. equally due to multiple computations
+    count_noise_standard_deviation = \
+        sum_noise_standard_deviation = \
+            2*dp_params.noise_standard_deviation \
+                if dp_params.noise_standard_deviation is not None \
```
dvadym, line 410: Nit: maybe something like that
```
noise_std = None
if dp_params.noise_standard_deviation is not None:
  noise_std = 2*dp_params.noise_standard_deviation
count_noise_standard_deviation = sum_noise_standard_deviation = noise_std
```

pipeline_dp/dp_computations.py
```diff
@@ -103,50 +104,70 @@ def compute_sigma(eps: float, delta: float, l2_sensitivity: float):
...
+                                                    delta=delta,
+                                                    sensitivity=l2_sensitivity)
     return mechanism.add_noise(1.0 * value)
 
 
 def _add_random_noise(
     value: float,
     eps: float,
     delta: float,
+    noise_standard_deviation: float,
```
dvadym, line 170: Optional[float]?

dvadym (commented):
Thanks! It looks good. I left some comments.

Please add tests for apply_laplace/gaussian with std_deviation (no need to add tests for all function count/sum etc), because after the next refactoring they will be removed

PR #411, opened 2023-03-02
Refactor analysis output metrics: part 3 Per-partition partition selection metrics to cross-partition

dvadym:


analysis/cross_partition_combiners.py
```diff
@@ -94,3 +91,15 @@ def _sum_metrics_to_metric_utility(
         ratio_data_dropped=data_dropped,
         absolute_error=absolute_error,
         relative_error=relative_error)
+
+
+def _partition_selection_per_to_cross_partition(
+        prob_keep: float) -> metrics.PrivatePartitionSelectionUtility:
+    """Creates cross-partition partition selection utility from 1 partition utility."""
```
miracvbasaran, line 98: The name `PrivatePartitionSelectionUtility` sounds odd, should we rename them to `PrivatePartitionSelectionMetrics` instead?

Also, what does `from 1 partition utility` mean in the docstring? It seems to create a utility from 1 partition keep probability?

analysis/metrics.py
```diff
@@ -308,8 +314,8 @@ class MetricUtility:
 class PrivatePartitionSelectionUtility:
     """Stores aggregate metrics about partition selection."""
 
-    strategy: pipeline_dp.PartitionSelectionStrategy
-    num_partitions: float
+    strategy: Optional[pipeline_dp.PartitionSelectionStrategy]
```
miracvbasaran, line 317: Why is this optional?

miracvbasaran (approved):

analysis/metrics.py
dvadym: technical reason, when it's created strategy isn't available. Having something optional is not a big problem, e.g. optional is recommended for all fields in protobuffers

analysis/cross_partition_combiners.py
dvadym: Agree, `PrivatePartitionSelectionMetrics` is better

Fixed the docstring

dvadym (commented):
Thanks for review!

PR #413, opened 2023-03-02
Refactor analysis output metrics: part 5 Converting per partition to cross partition metrics

dvadym:


analysis/metrics.py
```diff
@@ -49,7 +49,7 @@ class SumMetrics:
 
 
 @dataclass
-class PerPartitionUtility:
+class PerPartitionUtilityMetrics:
```
miracvbasaran, line 52: Optional: Do we need the `Utility` in the name?

It seems it would be more consistent to remove it

analysis/cross_partition_combiners.py
```diff
@@ -152,3 +152,31 @@ def _multiply_float_dataclasses_field(dataclass, factor: float):
...
+    prob_to_keep = per_partition_utility.partition_selection_probability_to_keep
+    partition_selection_utility = None
+    if not public_partitions:
+        partition_selection_utility = _partition_selection_per_to_cross_partition(
+            prob_to_keep)
+    # Fill metric errors.
+    metric_errors = None
+    if dp_metrics:
+        metric_errors = []
+        assert len(per_partition_utility.metric_errors) == len(dp_metrics)
```
miracvbasaran, line 172: We never expect this to happen, right? So maybe add an error message or a code comment that says `This should not happen`?

miracvbasaran (approved):

analysis/metrics.py
dvadym: Agree, removed

analysis/cross_partition_combiners.py
dvadym: According the [style guide](https://google.github.io/styleguide/pyguide.html#244-decision) "assert is used to ensure internal correctness", i.e. assert False shouldn't happen, no need to write comment about that

dvadym (commented):
Thanks for review!

PR #415, opened 2023-03-03
Utility analysis performance optimization

dvadym:
This PR improves the speed of per-partition analysis metric computations. 

**Before this PR:** accumulator for per-partititions metrics could be in either sparse or dense mode and sparse -> dense transfer happens when sparse is big enough. And if sparse & dense merged, sparse is converted to dense. Converting is expensive and the sparse might contain only 1 data point (so many conversion).

**After this PR:** sparse and dense part is kept both, `merge((sparse1, dense1), (sparse2, dense2))` is `(merge(sparse1, sparse2), merge(dense1, dense2))`. That excludes sparse->dense conversion with small amount of elements


analysis/combiners.py
```diff
@@ -358,9 +358,33 @@ def _to_dense(self,
             ]),
         )
 
+    def _merge_sparse(self, acc1, acc2):
+        if acc1 is None:
+            return acc2
+        if acc2 is None:
+            return acc1
+        return tuple(_merge_list(s, t) for s, t in zip(acc1, acc2))
```
miracvbasaran, line 366: I don't remember this in detail. Was the sparse mode a tuple of lists?

miracvbasaran (approved):
Looks great, thanks!

analysis/combiners.py
dvadym: Yes, sparse mode is keeping raw values to aggregate (in lists).

dvadym:
Thanks for review!

PR #416, opened 2023-03-05
Refactor analysis output metrics: part 7 Cross-partition metrics combiner

dvadym:
This PR
1. Introduces CrossPartitionCombiner for computing utility global (i.e. cross-partition) analysis metrics 
2. All information about partitions (incl. partition selection) is now in `PartitionMetrics` (previously `PrivatePartitionSelectionMetrics` ), which addresses one of the inconsitencies previosly that Count Metrics contain information about partitions

### DIFF
--- a/analysis/cross_partition_combiners.py
+++ b/analysis/cross_partition_combiners.py
@@ -83,7 +83,7 @@ def _sum_metrics_to_metric_utility(
         sum_metrics, keep_prob=partition_keep_probability)
     relative_error = absolute_error.to_relative(sum_metrics.sum)
 
-    return metrics.UtilityReport(metric=dp_metric,
+    return metrics.MetricUtility(metric=dp_metric,
                                  noise_std=sum_metrics.std_noise,
                                  noise_kind=sum_metrics.noise_kind,
                                  ratio_data_dropped=data_dropped,
@@ -91,7 +91,7 @@ def _sum_metrics_to_metric_utility(
                                  relative_error=relative_error)
 
 
-def _create_partition_metrics_for_public_partitions(
+def _partition_metrics_public_partitions(
         is_empty_partition: bool) -> metrics.PartitionMetrics:
     result = metrics.PartitionMetrics(public_partitions=True,
                                       num_dataset_partitions=0,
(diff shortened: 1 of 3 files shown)


analysis/cross_partition_combiners.py
```diff
@@ -83,28 +83,34 @@ def _sum_metrics_to_metric_utility(
...
-        ratio_dropped_data=0)  # todo(dvadym): implement ratio_dropped_data
+    return metrics.MetricUtility(metric=dp_metric,
+                                 noise_std=sum_metrics.std_noise,
+                                 noise_kind=sum_metrics.noise_kind,
+                                 ratio_data_dropped=data_dropped,
+                                 absolute_error=absolute_error,
+                                 relative_error=relative_error)
+
+
+def _partition_metrics_public_partitions(
```
miracvbasaran, line 94: Optional: Maybe rename to `_to_partition_metrics()` instead and handle both cases (private and public) inside the same function?

analysis/cross_partition_combiners.py
```diff
@@ -156,17 +162,20 @@ def _multiply_float_dataclasses_field(dataclass, factor: float) -> None:
             _multiply_float_dataclasses_field(value, factor)
 
 
-def _per_partition_to_cross_partition_utility(
+def _per_partition_to_cross_partition_metrics(
```
miracvbasaran, line 165: nit: This returns the UtilityReport, AFAIU the last step in the process. If so, the name and the doctring doesn't really match that.

analysis/cross_partition_combiners.py
dvadym: I thought about that, but that function is going to be pretty complicated, that's why I split in 2 functions.

dvadym (commented):
Thanks for comments! PTAL

analysis/metrics.py
```diff
@@ -311,13 +303,32 @@ class MetricUtility:
 
 
 @dataclass
-class PrivatePartitionSelectionMetrics:
-    """Stores aggregate metrics about partition selection."""
+class PartitionMetrics:
```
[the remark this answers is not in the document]
miracvbasaran: I think `Partition(s)Info(rmation)` sounds better than `PartitionMetrics` since it can be distinguished from the other ones. And maybe we can think of something better later on.

miracvbasaran (approved):

analysis/metrics.py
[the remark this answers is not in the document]
dvadym: Agree, I've renamed.

dvadym:
Thanks for review!

PR #417, opened 2023-03-07
Refactor analysis output metrics, part 8: perform_utility_analysis with new combiners

dvadym:
This PR implements  `perform_utility_analysis_new` which uses new cross-partition combiners and returns new the result in new dataclasses.

Note: after migrating old code, the old `perform_utility_analysis` will be dropped and `perform_utility_analysis_new` will be renamed to `perform_utility_analysis`

Also this PR contains small changes in dataclasses and in cross partition combiners.


analysis/utility_analysis.py
```diff
@@ -160,3 +163,118 @@ def _create_aggregate_error_compound_combiner(
...
+      options: options for utility analysis.
+      data_extractors: functions that extract needed pieces of information
+        from elements of 'col'. In case if the analysis performed on
+        pre-aggregated data, it should have type PreAggregateExtractors
+        otherwise DataExtractors.
+      public_partitions: A collection of partition keys that will be present
+        in the result. If not provided, the utility analysis with private
+        partition selection will be performed.
+      return_per_partition: if true, it returns tuple, with the 2nd element
+        utility analysis per partitions.
```
miracvbasaran, line 192: IMO it feels incomplete to only mention the 2nd element of the tuple here

analysis/utility_analysis.py
```diff
@@ -160,3 +163,118 @@ def _create_aggregate_error_compound_combiner(
...
+    Returns:
+         if return_per_partition == False:
```
miracvbasaran, line 194: I don't see return_per_partition used anywhere in the code. Is this up-to-date?

analysis/utility_analysis.py
```diff
@@ -160,3 +163,118 @@ def _create_aggregate_error_compound_combiner(
...
+            returns collections which contains metrics.UtilityReport with
+            one report per each input configuration
+        else returns tuple, with the 1st element as in first 'if' clause and the
+        2nd a collection which contains utility analysis results per partition).
```
miracvbasaran, line 198: What exactly does utility analysis results per partition mean exactly?

miracvbasaran (approved):

analysis/utility_analysis.py
dvadym: Thx, added types of the output of per partition utility results

dvadym (commented):
Thanks a lot for review!

PR #418, opened 2023-03-09
Private contribution bounds computation

### LINKED ISSUE #261: Auto-tunning contribution bounding parameters
# Context

[DPEngine.aggregate](https://github.com/OpenMined/PipelineDP/blob/41b70a3c7e19b82024e2d0f44842aaab570440bd/pipeline_dp/dp_engine.py#L52) is API function that performs DP aggregation. [Parameters](https://github.com/OpenMined/PipelineDP/blob/41b70a3c7e19b82024e2d0f44842aaab570440bd/pipeline_dp/aggregate_params.py#L55) of this function include parameters related to contribution bounding (`max_partitions_contributed`,  `max_contributions_per_partition`). 

For `max_partitions_contributed`,  `max_contributions_per_partition` if a privacy unit (definition of privacy unit and more on contribution bounding is [here](https://pipelinedp.io/key-definitions/) ) contributes more, contributions are sub-sampled. So smaller these parameters, bigger the ratio of input data which is dropped, but smaller DP noise is added. So there is a trade-off of choosing those parameters. Choosing optimal (in some sense) parameters is data dependent.

# Goals
To introduce estimating "good" `max_partitions_contributed` and `max_contributions_per_partition` with differential privacy, that allow to achieve reasonable accuracy.

**Note:** this is researchy problem, it will require experimenting and knowledge of Differential Privacy. There is no clear best solution and there are many possible approaches.

**Some ideas:**
1. Choosing `max_partitions_contributed` and  `max_contributions_per_partition` might be performed independently (at least probably that's a good start)
2. Choosing `max_partitions_contributed` and  `max_contributions_per_partition` might be considered as a selection problem and for example [exponential mechanism](https://en.wikipedia.org/wiki/Exponential_mechanism_(differential_privacy)) might be used.
3. There might be many different score functions to be used - for example some tradeoff between dropped data and DP noise or that dropped data because of sampling are around some fixed ratio (say 1%).

I'm ready to help in discussions and experimenting.

RamSaw:
## Description
Implements differentially private algorithm to choose max_partitions_contributed bound. Also, introduces a common API that will support calculation of max_contributions_per_partition too.
Closes #261

## Affected Dependencies
List any dependencies that are required for this change.

## How has this been tested?
- Describe the tests that you ran to verify your changes.
- Provide instructions so we can reproduce.
- List any relevant details for your test configuration.

## Checklist


pipeline_dp/dp_engine.py
```diff
@@ -57,6 +57,46 @@ def _add_report_stages(self, stages_description):
         for stage_description in stages_description:
             self._add_report_stage(stage_description)
 
+    def calculate_private_contribution_bounds(
+        self,
+        col,
+        params: pipeline_dp.CalculatePrivateContributionBoundsParams,
+        data_extractors: pipeline_dp.DataExtractors,
+        partitions: Any,
```
RamSaw, line 65: So, actually, we will need to get the number of partitions. But if it is PCollection, it will be not trivial, but doable. I guess we can do something similar to what histograms computation does. We calculate the number and get PCollection<Integer> and then we calculate the histogram and get PCollection<Histogram>. And then I guess in python it will be possible to flatten it even though they have different types (if not we can create type HistogramOrInt :) ?): backend.flatten(histogram, number). And then we will have PCollection<Object> and we will call backend.to_list(pCollection) and then we will be able to apply map function, something like lambda el: el[0] is histogram, el[1] is number of partitions.
Does it make sense to you?

dvadym: Yeah, I agree that backend.flatten(histogram, number) approach would work and it seems to be a good approach. And no worries about the type, it's Python :).

Since otherwise we would need to ask user to compute the number of partitions, So we have a tradeoff between user convenience and complexity of the implementation. Since complexity is not high, here I'm for computing the count by ourselves.

RamSaw: Nice, then I will send you implementation next week!

RamSaw:
Hi!

I implemented the algo, but didn't test it thoroughly. Let's discuss the structure, agree on that and then I will add tests and comprehensive comments to the code.
To implement algo in DPEngine I had to break circular dependencies and therefore I had to move histograms.py and PreAggregateExtractors class to pipeline_dp module. Also, I had to extract DataExtractors into its own file. Does it make sense? Or we should structure it differently. This refactoring broke some tests in analysis package and maybe in utility_analysis package. But I think it should be fixable and I will fix it if the current structure LGTM.

pipeline_dp/private_contribution_bounds.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def _calculate_number_of_partitions(self):
+        distinct_partitions = self._backend.distinct(
+            self._partitions, "Keep only distinct partitions")
+        return self._backend.size(distinct_partitions,
+                                  "Calculate number of partitions")
+
+    def _lower_bounds_of_bins(self, histogram_col):
+
+        def histogram_to_bins(hist: Histogram):
```
dvadym, line 68: Nit: -> List[int]
Nit: histogram_to_bin_lowers

pipeline_dp/data_extractors.py
```diff
@@ -0,0 +1,37 @@
+import dataclasses
+from typing import Callable
+
+
+@dataclasses.dataclass
+class DataExtractors:
```
dvadym, line 6: Thanks for extracting it to a separate file!

pipeline_dp/dp_computations.py
```diff
@@ -641,3 +643,42 @@ def create_additive_mechanism(
...
+
+    def _calculate_probabilities(self, eps: float,
+                                 inputs_to_score_col: typing.List[Any]):
+        scores = np.array(
+            list(
+                map(lambda k: self._scoring_function.score(k),
+                    inputs_to_score_col)))
+        denominator = self._scoring_function.global_sensitivity
+        if not self._scoring_function.is_monotonic:
+            denominator *= 2
```
dvadym, line 682: Nice, to take into consideration monotonic sensitivity!

pipeline_dp/dp_computations.py
```diff
@@ -641,3 +643,42 @@ def create_additive_mechanism(
...
+class ExponentialMechanism:
+
+    class ScoringFunction(abc.ABC):
+
+        def __init__(self) -> None:
+            super().__init__()
+
+        @abc.abstractmethod
+        def score(self, k) -> float:
+            """"""
```
dvadym, line 657: Nit: pass or (even better) extend docstring

pipeline_dp/private_contribution_bounds.py
```diff
@@ -0,0 +1,127 @@
...
+        self._partitions = partitions
+        self._histograms = histograms
+
+    @dataclass
+    class Inputs:
+        l0_histogram: Histogram
+        number_of_partitions: int
+        possible_contribution_bounds: List[int]
+
+    @lru_cache(maxsize=None)
```
dvadym, line 28: This would cache calculation per calculator, right? Do we need to call it more than once?

pipeline_dp/private_contribution_bounds.py
```diff
@@ -0,0 +1,127 @@
...
+            noise_kind=self._params.aggregation_noise_kind,
+            min_value=None,
+            max_value=None,
+            min_sum_per_partition=None,
+            max_sum_per_partition=None)
+        return (self._number_of_partitions *
+                compute_dp_count_noise_std(noise_params))
+
+    def _l0_impact_dropped(self, k):
+        capped_contributions = map(
```
dvadym, line 117: This is basically the number of dropped records, right? AFAIU it provides quadratic complexity.
I'd suggest implementing that in histograms class with 1 pass in different PR. It would be super useful for non-dp parameter tuning as well.

dvadym (commented):

pipeline_dp/pipeline_backend.py
```diff
@@ -172,6 +172,46 @@ def flatten(self, cols: Iterable, stage_name: str):
...
+    def size(self, col, stage_name: str):
+        """Returns a one element collection that contains the size of the input collection."""
+
+        col = self.map(col, lambda x: "fake_common_key",
+                       f"{stage_name}: mapping to the same key")
+        col = self.count_per_element(
+            col, f"{stage_name}: counting the number of elements")
+        return self.values(col, f"{stage_name}: dropping the fake_common_key")
+
+    def collect(self, cols: typing.List, container_class: typing.Type,
```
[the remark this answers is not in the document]
RamSaw: what do you think about collectToContainer?

pipeline_dp/private_contribution_bounds.py
```diff
@@ -0,0 +1,127 @@
...
+        l0_histogram = self._backend.map(
+            self._histograms, lambda h: h.l0_contributions_histogram,
+            "Extract l0_contributions_histogram from DatasetHistograms")
+        l0_histogram = self._backend.to_multi_transformable_collection(
+            l0_histogram)
+        number_of_partitions = self._calculate_number_of_partitions()
+        number_of_partitions = self._backend.to_multi_transformable_collection(
+            number_of_partitions)
+        possible_contribution_bounds = self._lower_bounds_of_bins(l0_histogram)
+        possible_contribution_bounds = self._backend.to_multi_transformable_collection(
```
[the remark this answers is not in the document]
RamSaw: ah, I think I used it for debugging, will remove it :)

pipeline_dp/dp_computations.py
RamSaw: what do you mean?

pipeline_dp/private_contribution_bounds.py
RamSaw: yes, it is. yeah, we can precalculate the values ones (like partial sums) and then make _l0_impact_dropped with O(1) time instead of O(n).

pipeline_dp/pipeline_backend.py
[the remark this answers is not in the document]
dvadym: SGTM!

pipeline_dp/dp_computations.py
dvadym: it's better 
```
def score():
  pass
```
or
```
def score():
  """docstring"""
```

pipeline_dp/pipeline_backend.py
```diff
@@ -172,6 +172,46 @@ def flatten(self, cols: Iterable, stage_name: str):
     def distinct(self, col, stage_name: str):
         """Returns a collection containing distinct elements of the input collection."""
 
+    def size(self, col, stage_name: str):
```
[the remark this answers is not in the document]
RamSaw: Done. But I named the file more specific: "pipeline_composite_functions", there is a best practice to avoid names utils, because they are not specific enough and it will become a dump of different methods not related to each other. WDYT?

pipeline_dp/private_beam.py
```diff
@@ -414,7 +414,7 @@ def expand(self, pcol: pvalue.PCollection) -> pvalue.PCollection:
             params,
             data_extractors,
             self._public_partitions,
-            out_explain_computaton_report=self._explain_computaton_report)
+            out_explain_computation_report=self._explain_computaton_report)
```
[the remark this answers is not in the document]
RamSaw: np :)

pipeline_dp/dp_engine.py
```diff
@@ -57,6 +46,57 @@ def _add_report_stages(self, stages_description):
         for stage_description in stages_description:
             self._add_report_stage(stage_description)
 
+    def calculate_private_contribution_bounds(
```
dvadym, line 49: Could you please move it as a last method of DPEngine?

pipeline_dp/private_contribution_bounds.py
```diff
@@ -0,0 +1,190 @@
...
+        return composite_funcs.size(self._backend, distinct_partitions,
+                                    "Calculate number of partitions")
+
+    def _lower_bounds_of_bins(self, histogram_col):
+        # TODO: is correct to do so? does it preserve dp? The problem is that
+        # we infer it from the histogram that used raw data and if for example
+        # there are 7 different public partitions but each user contributed to 6
+        # partitions max, then we will choose only from 1..6 and not 1..7, but
+        # it seems incorrect because we eliminate one of the options by
+        # observing the data. (Example taken from restaurant visits).
```
RamSaw, line 98: What do you think about this concern in TODO?

pipeline_dp/private_contribution_bounds.py
dvadym: sg, it makes sense.

pipeline_dp/private_contribution_bounds.py
```diff
@@ -0,0 +1,191 @@
...
+        Returns a one element pCollection containing the value for the l0
+        bound."""
+        l0_histogram = self._backend.to_multi_transformable_collection(
+            self._backend.map(
+                self._histograms, lambda h: h.l0_contributions_histogram,
+                "Extract l0_contributions_histogram from DatasetHistograms"))
+        number_of_partitions = self._calculate_number_of_partitions()
+        possible_contribution_bounds = self._lower_bounds_of_bins(l0_histogram)
+
+        l0_calculation_input_col = self._backend.to_multi_transformable_collection(
```
dvadym, line 67: Do we need `to_multi_transformable_collection` here?

pipeline_dp/dp_computations.py
```diff
@@ -641,3 +643,57 @@ def create_additive_mechanism(
         return GaussianMechanism(spec.epsilon, spec.delta, sensitivities.l2)
 
     assert False, f"{spec.noise_kind} not supported."
+
+
+class ExponentialMechanism:
+    """Exponential mechanism that can be used to choose a parameter
+    from a set of possible parameters in a differentially private way.
+
+    https://en.wikipedia.org/wiki/Exponential_mechanism"""
```
dvadym, line 652: Please indicate limitations of this class in docstring: all computations are in memory, so the set from which we choose should fit in memory

pipeline_dp/dp_computations.py
```diff
@@ -641,3 +643,57 @@ def create_additive_mechanism(
...
+        def is_monotonic(self) -> bool:
+            """Whether score(k) is monotonic.
+
+            If true then results will be more precise."""
+
+    def __init__(self, scoring_function: ScoringFunction) -> None:
+        self._scoring_function = scoring_function
+
+    def apply(self, eps: float, inputs_to_score_col: typing.List[Any]) -> Any:
+        """Applies exponential mechanism and chooses a parameter from the list
```
dvadym, line 683: Nit: by style guide, docstring

"""Short description.

Long description
""""
Please fix

pipeline_dp/dp_computations.py
```diff
@@ -641,3 +643,57 @@ def create_additive_mechanism(
...
+        of possible parameters in a differentially private way."""
+
+        probs = self._calculate_probabilities(eps, inputs_to_score_col)
+        return np.random.default_rng().choice(inputs_to_score_col, p=probs)
+
+    def _calculate_probabilities(self, eps: float,
+                                 inputs_to_score_col: typing.List[Any]):
+        scores = np.array(
+            list(
+                map(lambda k: self._scoring_function.score(k),
```
dvadym, line 693: Nit: map( self._scoring_function.score,  inputs_to_score_col))Nit: map( self._scoring_function.score,  inputs_to_score_col))

tests/private_contribution_bounds_test.py
```diff
@@ -0,0 +1,278 @@
...
+
+    def test_calculate_one_bound_has_much_higher_score_returns_it(self):
+        # Arrange
+        params = construct_params(
+            aggregation_eps=0.9,
+            aggregation_delta=0,
+            calculation_eps=0.1,
+            aggregation_noise_kind=pipeline_dp.NoiseKind.LAPLACE,
+            max_partitions_contributed_upper_bound=2)
+        partitions = [i + 1 for i in range(200)]
```
dvadym, line 241: Nit: list(range(1, 201))

tests/pipeline_composite_functions_test.py
```diff
@@ -0,0 +1,171 @@
...
+
+# TODO: check whether possible to move it inside.
+@dataclass
+class TestContainer:
+    x: int
+    y: str
+    z: List[str]
+
+
+class PipelineCompositeFunctionsTest(parameterized.TestCase):
```
RamSaw, line 65: @dvadym , do you have idea why tests in this file in python 3.8 and 3.9 on Ubuntu might fail for RDD with errors like this: https://paste.googleplex.com/5441345786216448?

tests/private_contribution_bounds_test.py
```diff
@@ -0,0 +1,278 @@
...
+        l0_scoring_function = private_contribution_bounds.L0ScoringFunction(
+            params, number_of_partitions, l0_histogram)
+
+        with self.assertRaises(RuntimeError):
+            l0_scoring_function.score(-1)
+
+        with self.assertRaises(RuntimeError):
+            l0_scoring_function.score(0)
+
+    def test_score_gaussian_noise_valid_value_returns_value_of_a_correct_order_of_magnitude(
```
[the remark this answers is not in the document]
RamSaw: hm, the value you gave me in the meeting didn't work, maybe I composed it with l2_sensitivity incorrectly. Could you share the colab? I did this: std * l2_sensitivity, is it correct?

pipeline_dp/private_contribution_bounds.py
```diff
@@ -0,0 +1,191 @@
...
+        return composite_funcs.size(self._backend, distinct_partitions,
+                                    "Calculate number of partitions")
+
+    def _lower_bounds_of_bins(self, histogram_col):
+        # TODO: is correct to do so? does it preserve dp? The problem is that
+        # we infer it from the histogram that used raw data and if for example
+        # there are 7 different public partitions but each user contributed to 6
+        # partitions max, then we will choose only from 1..6 and not 1..7, but
+        # it seems incorrect because we eliminate one of the options by
+        # observing the data. (Example taken from restaurant visits).
```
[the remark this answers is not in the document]
RamSaw: implemented.

tests/pipeline_backend_test.py
```diff
@@ -330,6 +328,11 @@ def test_sum_per_key(self):
         result = self.backend.sum_per_key(data).collect()
         self.assertEqual(set(result), set([(1, 6), (2, -2), (3, 8), (10, 5)]))
 
+    @unittest.skipIf(
+        sys.platform == "darwin" and sys.version_info.major == 3 and
+        (sys.version_info.minor == 8 or sys.version_info.minor == 9),
+        "There are some problems with PySpark setup on macOS in python 3.8 and 3.9."
+    )
```
RamSaw, line 335: submit it in a separate PR without reverting this

tests/pipeline_composite_functions_test.py
dvadym: I checked Spark tests and recalled that there was some problem with creating Spark context multiple times. Could you please try to create it only once? [Here](https://github.com/OpenMined/PipelineDP/blob/be9087db6d7a7ab7aaa13aed7aacaa052e28686e/tests/private_spark_test.py#L35) an example

pipeline_dp/pipeline_backend.py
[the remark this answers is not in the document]
RamSaw: Done

tests/pipeline_composite_functions_test.py
RamSaw: Rewrote the test file, now we have separate test classes for each backend, the same way pipeline_backend_test.py is written and these tests are not parametrized by backend anymore. I think this way is better because it is consistent with pipeline_backend_test.py and also more maintainable.

pipeline_dp/dp_engine.py
RamSaw: Done.

tests/pipeline_backend_test.py
RamSaw: Done

pipeline_dp/histograms.py
```diff
@@ -111,7 +108,9 @@ class DatasetHistograms:
 
 
 def _to_bin_lower(n: int) -> int:
-    """Finds the lower bound of the histogram bin which contains the given integer."""
+    """Finds the lower bound of the histogram bin which contains the given integer.
+    Keep in sync with private_contribution_bounds._generate_possible_contribution_bounds.
```
dvadym, line 112: Nit: add blank line

pipeline_dp/private_contribution_bounds.py
```diff
@@ -0,0 +1,193 @@
...
+    def __init__(self,
+                 params: pipeline_dp.CalculatePrivateContributionBoundsParams,
+                 number_of_partitions: int, l0_histogram: Histogram):
+        super().__init__()
+        self._params = params
+        self._number_of_partitions = number_of_partitions
+        self._l0_histogram = l0_histogram
+
+    def score(self, k: int) -> float:
+        """
```
dvadym, line 104: Nit: add 1 line docstring description, something like "Computes... "

pipeline_dp/private_contribution_bounds.py
```diff
@@ -0,0 +1,193 @@
...
+        P := number_of_partitions
+        std := count_noise_std
+        B := _max_partitions_contributed_best_upper_bound (= global_sensitivity)
+        D := dataset
+        uid := user identifier
+        #contributions(uid, D) = number of partitions in D where uid contributed
+        at least once.
+
+        score(k) = -0.5 * impact_noise(k) - 0.5 * impact_dropped(k) =(1)
```
dvadym, line 113: Nit: add introductory word, something like
Then 
 score(k) =

dvadym (commented):
Thanks it looks great. A few minor comments

pipeline_dp/dp_computations.py
```diff
@@ -643,6 +645,63 @@ def create_additive_mechanism(
...
+        @abc.abstractmethod
+        def global_sensitivity(self) -> float:
+            """Global sensitivity of the scoring function."""
+
+        @property
+        @abc.abstractmethod
+        def is_monotonic(self) -> bool:
+            """Whether score(k) is monotonic.
+
+            Definition of a monotonic function:
```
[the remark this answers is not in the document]
RamSaw: Oh, really. I thought it is the standard definition. What is the definition then?

tests/private_contribution_bounds_test.py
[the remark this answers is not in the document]
RamSaw: Recalculated and now it works, looks like I made a mistake somewhere when I tried to calculate it in the past.

tests/pipeline_composite_functions_test.py
RamSaw: Now it works, probably we had to call SparkContext.parallelize(input_collection).

dvadym (approved):
Thanks!

PR #419, opened 2023-03-13
Introducing AdditiveMechanism class

dvadym:
This is the 1st PR in refactoring on making noise addition more OOP. In later PRs AdditiveMechanism classes will be used for computing DP aggregations.


tests/dp_computations_test.py
```diff
@@ -657,5 +657,102 @@ def test_compute_dp_sum_noise_std_gaussian(self, eps: float, delta: float,
...
+
+    @parameterized.parameters(
+        dict(epsilon=2, l1_sensitivity=4.5, expected_noise=2.25),
+        dict(epsilon=0.1, l1_sensitivity=0.55, expected_noise=5.5),
+    )
+    def test_laplace_mechanism_creation(self, epsilon, l1_sensitivity,
+                                        expected_noise):
+        mechanism = dp_computations.LaplaceMechanism(
+            epsilon=epsilon, l1_sensitivity=l1_sensitivity)
+        self.assertEqual(mechanism.noise_kind, pipeline_dp.NoiseKind.LAPLACE)
```
monsieurmuffin, line 670: Very nit: please could you add a blank line between "arrange" "act" and  "assert" sections here and in the other tests? I agree that in most of the tests it's hard to mix them up since the code is nice and simple, but just to follow the standard code style conventions.

tests/dp_computations_test.py
```diff
@@ -657,5 +657,102 @@ def test_compute_dp_sum_noise_std_gaussian(self, eps: float, delta: float,
...
+                                            value, expected_noise_scale):
+        # Use Kolmogorov-Smirnov test to verify the output noise distrition.
+        # https://en.wikipedia.org/wiki/Kolmogorov-Smirnov_test
+        mechanism = dp_computations.LaplaceMechanism(
+            epsilon=epsilon, l1_sensitivity=l1_sensitivity)
+        expected_cdf = stats.laplace(loc=value, scale=expected_noise_scale).cdf
+        noised_values = [mechanism.add_noise(value) for _ in range(30000)]
+
+        res = stats.ks_1samp(noised_values, expected_cdf)
+        self.assertGreater(res.pvalue, 1e-4)
```
monsieurmuffin, line 701: Do you know what flakiness we expect here approximately?

monsieurmuffin (approved):

tests/dp_computations_test.py
dvadym: It shouldn't be more than 1e-4 per test (that's why it tested against 1e-4). So in total something like 4*4e-4 (since there are 4 tests).

tests/dp_computations_test.py
dvadym: Good point. Done

dvadym (commented):
Thanks for review!

PR #420, opened 2023-03-14
Factory method for creating AdditiveMechanism

dvadym:


pipeline_dp/dp_computations.py
```diff
@@ -571,3 +571,68 @@ def std(self) -> float:
...
         return self._mechanism.l2_sensitivity
+
+
+@dataclass
+class Sensitivities:
+    """Contains sensitivities of the additive DP mechanism."""
+    l0: Optional[int] = None
+    linf: Optional[float] = None
+    l1: Optional[float] = None
+    l2: Optional[float] = None
```
monsieurmuffin, line 582: Why do we want to have one class that contains all sensitivities at the same time? Wouldn't it be cleaner to have different classes for l1 and l2 sensitivities and include all necessary elements in each of them?

tests/dp_computations_test.py
```diff
@@ -757,6 +757,105 @@ def test_gaussian_mechanism_distribution(self, epsilon, delta,
...
+        dict(l0_sensitivity=None,
+             linf_sensitivity=None,
+             l1_sensitivity=0,
+             l2_sensitivity=None,
+             expected_error="L1 has to be positive"),
+        dict(l0_sensitivity=None,
+             linf_sensitivity=None,
+             l1_sensitivity=None,
+             l2_sensitivity=-5,
+             expected_error="L2 has to be positive"),
```
monsieurmuffin, line 780: nit/opt: I would say "must be positive"

tests/dp_computations_test.py
```diff
@@ -757,6 +757,105 @@ def test_gaussian_mechanism_distribution(self, epsilon, delta,
...
+             l1_sensitivity=7,
+             l2_sensitivity=None,
+             expected_error="L1=7 != .*=8"),
+        dict(l0_sensitivity=4,
+             linf_sensitivity=5,
+             l1_sensitivity=None,
+             l2_sensitivity=9,
+             expected_error="L2=9 != .*=10"),
+    )
+    def test_sensitivities_post_init_validation(self, l0_sensitivity,
```
monsieurmuffin, line 792: Shall we also test that either l1 or l2 or (both l0 and linf) is set?

monsieurmuffin (approved):

dvadym (commented):
Thanks for the review!

PR #422, opened 2023-03-15
Using AdditiveMechanism in Combiners

dvadym:
This is the part of the refactoring for making adding noise more OOP. This refactoring is for supporting PLD and l1 sensitivities in future.

This PR implements using recently impelment AdditiveMechanism in Count/PrivacyIDCount/Sum combiners. Namely, it introduces [mixin class](https://en.wikipedia.org/wiki/Mixin#In_Python)  `MechanismContainerMixin` for creating and keeping AdditiveMechanism in Combiners.

### DIFF
--- a/pipeline_dp/combiners.py
+++ b/pipeline_dp/combiners.py
@@ -177,9 +177,6 @@ class CombinerParams:
 
 class MechanismContainerMixin(abc.ABC):
 
-    def __init__(self):
-        self._mechanism = None
-
     @abc.abstractmethod
     def sensitivities(self) -> dp_computations.Sensitivities:
         pass
--- a/pipeline_dp/combiners.py
+++ b/pipeline_dp/combiners.py
@@ -176,6 +176,7 @@ class CombinerParams:
 
 
 class MechanismContainerMixin(abc.ABC):
+    """Abstract class with implementation of handling DP mechanism."""
 
     @abc.abstractmethod
     def sensitivities(self) -> dp_computations.Sensitivities:
@@ -187,7 +188,8 @@ class MechanismContainerMixin(abc.ABC):
 
     def __getstate__(self):
         state = self.__dict__.copy()
-        # _mechanism is not serializable. And it is created on demand.
+        # Do not serialize _mechanism, because it is not serializable and it is
+        # created on demand.
         if "_mechanism" in state:
             del state["_mechanism"]
         return state
--- a/pipeline_dp/combiners.py
+++ b/pipeline_dp/combiners.py
@@ -185,6 +185,7 @@ class CountCombiner(Combiner):
 
     def __init__(self, params: CombinerParams):
         self._params = params
+        self._mechanism = None
 
     def create_accumulator(self, values: Sized) -> AccumulatorType:
         return len(values)
@@ -194,11 +195,7 @@ class CountCombiner(Combiner):
         return count1 + count2
 
     def compute_metrics(self, count: AccumulatorType) -> dict:
-        return {
-            'count':
-                dp_computations.compute_dp_count(
-                    count, self._params.scalar_noise_params)
-        }
+        return {'count': self._get_mechanism().add_noise(count)}
 
     def metrics_names(self) -> List[str]:
         return ['count']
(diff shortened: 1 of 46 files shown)


pipeline_dp/combiners.py
```diff
@@ -175,16 +175,49 @@ def additive_vector_noise_params(
...
+
+    @abc.abstractmethod
+    def sensitivities(self) -> dp_computations.Sensitivities:
+        pass
+
+    @abc.abstractmethod
+    def mechanism_spec(self) -> budget_accounting.MechanismSpec:
+        pass
+
+    def __getstate__(self):
```
miracvbasaran, line 189: What is state? I don't see references to it in the rest of the code.

pipeline_dp/combiners.py
```diff
@@ -225,56 +263,67 @@ def merge_accumulators(self, accumulator1: AccumulatorType,
...
 
-    def __init__(self, params: CombinerParams):
-        self._params = params
-        self._bouding_per_partition = params.aggregate_params.bounds_per_partition_are_set
+    def __init__(self, mechanism_spec: budget_accounting.MechanismSpec,
+                 aggregate_params: pipeline_dp.AggregateParams):
+        self._mechanism_spec = mechanism_spec
+        self._sensitivities = dp_computations.compute_sensitivities_for_sum(
+            aggregate_params)
+        self._bouding_per_partition = aggregate_params.bounds_per_partition_are_set
```
miracvbasaran, line 295: typo in "bouding"

pipeline_dp/combiners.py
```diff
@@ -225,56 +263,67 @@ def merge_accumulators(self, accumulator1: AccumulatorType,
...
         return accumulator1 + accumulator2
 
-    def compute_metrics(self, accumulator: AccumulatorType) -> dict:
-        return {
-            'privacy_id_count':
-                dp_computations.compute_dp_count(
-                    accumulator, self._params.scalar_noise_params)
-        }
+    def compute_metrics(self, count: AccumulatorType) -> dict:
+        return {"privacy_id_count": self.get_mechanism().add_noise(count)}
```
miracvbasaran, line 267: This seems to be a shift in architecturefrom one  where dp_computations were mostly responsible for dp logic to one where combiners actually do the dp logic but only use dp_computations to add noise.

Is supporting PLD inherently incompatible with using primitives for Sum/Count/etc. from the building block libraries?

pipeline_dp/combiners.py
dvadym: Python methods \_\_some_name\._\_ are service methods (like \_\init\_\_ \_\eq\._\_ etc).
\_\getstate\._\_ is called when serialization happens, [some examples](https://stackoverflow.com/questions/1939058/simple-example-of-use-of-setstate-and-getstate).

I've aded comment in this function to explain in more details.

pipeline_dp/combiners.py
dvadym: This is not shift from dp_computations, this refactoring for making dp_computations more OOP.
Namely it introduces nice OOP concept:
1. [AdditiveMechanism](https://github.com/OpenMined/PipelineDP/blob/28d30a020adf3e171c871ab7e86f0b0fcb08026a/pipeline_dp/dp_computations.py#L496), which encapsulate everythng related to dealing with mechanism.
2. [Sensitivities](https://github.com/OpenMined/PipelineDP/blob/28d30a020adf3e171c871ab7e86f0b0fcb08026a/pipeline_dp/dp_computations.py#L579) which keeps sensitivities and makes conversion from l0/linf -> l1/l2 sensitivities

And there are factory functions for creating `Sensitivities`/`AdditiveMechanism` for different situations. 

There are many benefits of having classes instead of plain functions, even w/o PLD, e.g. it simplifies testing of noise added by combiner, we can take mechanism from combiner and check that the mechanism is correct (noise_kind, noise_parameter), in the previous setup there were no easy method.

Supporting PLD will be simple - just needed to create another factory method for creating mechanism. 

Also this design plays nice with L1 contribution bounding - `Sensitivities` already supports l1 :).

dvadym (commented):
Thanks for comments! PTAL

miracvbasaran (approved):

dvadym:
Thanks for review!

PR #428, opened 2023-04-20
Compute Sensitivities for COUNT, SUM, PRIVACY_ID_COUNT

dvadym:
This is the part of the refactoring for making adding noise more OOP. This refactoring is for supporting PLD and l1 sensitivities in future.

### DIFF
--- a/pipeline_dp/aggregate_params.py
+++ b/pipeline_dp/aggregate_params.py
@@ -262,7 +262,7 @@ class AggregateParams:
                     "both max_partitions_contributed and "
                     "max_contributions_per_partition must be set.")
             elif n_not_none == 1:
-                raise ValueError("AggregateParams: either none or both from "
+                raise ValueError("AggregateParams: either none or both "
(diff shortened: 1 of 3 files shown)


pipeline_dp/aggregate_params.py
```diff
@@ -262,10 +262,9 @@ def __post_init__(self):
                     "both max_partitions_contributed and "
                     "max_contributions_per_partition must be set.")
             elif n_not_none == 1:
-                raise ValueError(
-                    "AggregateParams: either none or both from "
-                    "max_partitions_contributed and "
-                    " max_contributions_per_partition must be set.")
+                raise ValueError("AggregateParams: either none or both from "
```
monsieurmuffin, line 265: nit: I would delete "from" ("either none or both max_partitions_contributed and ...")

monsieurmuffin (approved):

dvadym (commented):
Thanks for review!

PR #431, opened 2023-04-20
Precomputation of histogram tails

RamSaw:
## Description
A histogram has a sequence of bins. In many situations it is useful to be able to have some aggregated statistics given a bin.lower (aka bin id). In this cl we will introduce aggregations over the tail. For example, given bin.lower = k, what is the total sum of counts from bins with bin.lower >= k.

## Checklist

### DIFF
--- a/pipeline_dp/histograms.py
+++ b/pipeline_dp/histograms.py
@@ -98,6 +98,18 @@ class Histogram:
         return result[::-1]
 
 
+def histogram_precomputed_tails(hist: Histogram):
+    result = [None] * len(hist.bins)
+    accumulated_weight = 0
+    accumulated_count = 0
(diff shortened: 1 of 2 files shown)


pipeline_dp/histograms.py
```diff
@@ -98,6 +98,18 @@ def quantiles(self, q: List[float]) -> List[int]:
         return result[::-1]
 
 
+def histogram_precomputed_tails(hist: Histogram):
```
dvadym, line 101: compute_percentage_dropped_for_threshold? (and then to update computations)

RamSaw: ok, but what do you want to return then? Only one value? or two values: accumulated_weight / total_weight and accumulated_count / total_count?

PR #432, opened 2023-04-28
Add noise in explain computation report

dvadym:
This PR extends explain computation report by adding noise kind and noise parameters for count/privacy_id_count/sum. For other aggregation it will be added later

Now it looks like

5. Computed DP count with
     Gaussian mechanism:  parameter=17.64452389929557  eps=0.3333333333333333  delta=3.333333333333333e-07  l2_sensitivity=1.4142135623730951
 6. Computed DP sum with
     Gaussian mechanism:  parameter=88.22261949647785  eps=0.3333333333333333  delta=3.333333333333333e-07  l2_sensitivity=7.0710678118654755
 7. Computed DP privacy_id_count with
     Gaussian mechanism:  parameter=17.64452389929557  eps=0.3333333333333333  delta=3.333333333333333e-07  l2_sensitivity=1.4142135623730951

### DIFF
--- a/tests/combiners_test.py
+++ b/tests/combiners_test.py
@@ -224,6 +224,12 @@ class CountCombinerTest(parameterized.TestCase):
                                expected_noise_param,
                                delta=1e-3)
 
+    def test_explain_computation(self):
+        combiner = self._create_combiner(no_noise=False)
+        expected = ("Computed DP count.*\n.*Gaussian mechanism:  "
+                    "parameter=15.8.*eps=1.0  delta=1e-05.*l2_sensitivity=4.2")
+        self.assertRegex(combiner.explain_computation()(), expected)
+
 
 class PrivacyIdCountCombinerTest(parameterized.TestCase):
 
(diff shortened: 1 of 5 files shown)


tests/combiners_test.py
```diff
@@ -224,6 +224,12 @@ def test_mechanism(self, mechanism_type: pipeline_dp.MechanismType,
                                expected_noise_param,
                                delta=1e-3)
 
+    def test_explain_computation(self):
+        combiner = self._create_combiner(no_noise=False)
+        expected = ("Computed DP count.*\n.*Gaussian mechanism:  "
+                    "parameter=15.8.*eps=1.0  delta=1e-05.*l2_sensitivity=4.2")
+        self.assertRegex(combiner.explain_computation()(), expected)
```
miracvbasaran, line 231: How does the double function call work exactly?

miracvbasaran (approved):
LG, thanks!

tests/combiners_test.py
dvadym: `explain_compuation()` returns a function which returns string with explanation.

dvadym:
Thanks for review!

PR #436, opened 2023-05-05
L1 contribution histogram

dvadym:


analysis/contribution_bounders.py
```diff
@@ -61,11 +61,14 @@ def bound_contributions(self, col, params, backend, report_generator,
...
             num_partitions_contributed = len(partition_values)
+            num_contributions = sum(
+                (len(values) for _, values in partition_values))
             for partition_key, values in partition_values:
                 if sampler is not None and not sampler.keep(partition_key):
                     continue
                 yield (privacy_id, partition_key), (len(values), sum(values),
-                                                    num_partitions_contributed)
+                                                    num_partitions_contributed,
+                                                    num_contributions)
```
miracvbasaran, line 71: Why is L0LinfBounder returning num_contributions? Shouldn't we have a separate bounder in this file for this?

pipeline_dp/histograms.py
```diff
@@ -125,37 +127,66 @@ def _to_bin_lower(n: int) -> int:
...
-                                 deduplicate: bool = False):
+def _compute_frequency_histogram(col, backend: pipeline_backend.PipelineBackend,
+                                 name: HistogramType):
     """Computes histogram of element frequencies in collection.
+
     Args:
       col: collection with positive integers.
       backend: PipelineBackend to run operations on the collection.
       name: name which is assigned to the computed histogram.
-      deduplicate: performs deduplication of elements in the histogram. Namely,
```
miracvbasaran, line 137: Why remove this?

pipeline_dp/histograms.py
```diff
@@ -125,37 +127,66 @@ def _to_bin_lower(n: int) -> int:
...
+
+    col = backend.map_values(col, lambda x: int(round(x)), "Round")
+    # (int, sum_weights:int)
+
+    # Combiner elements to histogram buckets of increasing sizes. Having buckets
+    # of width = 1 is not scalable.
+    return _compute_frequency_histogram_helper(backend, col, name)
+
+
+def _compute_frequency_histogram_helper(backend, col, name: HistogramType):
```
miracvbasaran, line 172: Please add type hint for backend and have a consistent parameter order here and in the function above

pipeline_dp/histograms.py
```diff
@@ -230,6 +262,30 @@ def _compute_l0_contributions_histogram(
                                         HistogramType.L0_CONTRIBUTIONS)
 
 
+def _compute_l1_contributions_histogram(
+        col, backend: pipeline_backend.PipelineBackend):
+    """Computes histogram of the number of distinct records contributed by a privacy id.
+    This histogram contains: number of privacy ids which contributes to 1
```
miracvbasaran, line 268: Optional: "which contributes 1 record, 2 records, etc."`

Also, what do we do in the case of sum?

tests/histograms_test.py
```diff
@@ -65,17 +65,19 @@ def test_compute_frequency_histogram(self, input, expected):
...
+        histogram2 = hist.Histogram(hist.HistogramType.L1_CONTRIBUTIONS, None)
+        histogram3 = hist.Histogram(hist.HistogramType.LINF_CONTRIBUTIONS, None)
+        histogram4 = hist.Histogram(hist.HistogramType.COUNT_PER_PARTITION,
                                     None)
-        histogram4 = hist.Histogram(
+        histogram5 = hist.Histogram(
             hist.HistogramType.COUNT_PRIVACY_ID_PER_PARTITION, None)
         histograms = hist._list_to_contribution_histograms(
-            [histogram2, histogram1, histogram3, histogram4])
+            [histogram2, histogram1, histogram3, histogram5, histogram4])
```
miracvbasaran, line 75: nit: It is a bit odd that the order is different from the numeric suffix of the histograms

pipeline_dp/histograms.py
dvadym: This l1 contribution histogram is about number of total contributions, which can be used for estimating l1 sensitivity bounds for sum as well.

dvadym (commented):
Thanks for comments! PTAL

pipeline_dp/histograms.py
miracvbasaran: What is num_contributions when a user contributes -1,+1 across different partitions in the case of sum? Is it 0?

miracvbasaran (approved):

pipeline_dp/histograms.py
dvadym: There are multiple ways how to make l1 contribution bounding. The easiest is that
`l1_bound = num_contributions*max(abs(min_value), abs(max_value))`
And for such contribution bounding `num_contributions` is needed

PR #437, opened 2023-05-10
Computing percentage data dropped from histograms.

dvadym:


pipeline_dp/histograms.py
```diff
@@ -99,6 +100,46 @@ def quantiles(self, q: List[float]) -> List[int]:
...
+    threshold (e.g. in case of L0 histogram bounding_threshold is
+    max_partition_contribution).
+
+    Args:
+        contribution_histogram: histogram of contributions. It can be L0, L1,
+          Linf contribution bounding histogram.
+
+    Returns:
+        A sequence of sorted pairs
+        (bounding_threshold:int, ratio_data_draped:float). Where
```
miracvbasaran, line 118: nit: `float), where`

pipeline_dp/histograms.py
```diff
@@ -99,6 +100,46 @@ def quantiles(self, q: List[float]) -> List[int]:
...
+        bounding_thresholds are all lower of histogram.bins and
```
miracvbasaran, line 119: nit: `contribution_histogram` instead of histogram (also below)

pipeline_dp/histograms.py
```diff
@@ -99,6 +100,46 @@ def quantiles(self, q: List[float]) -> List[int]:
...
+    """
+    if not contribution_histogram.bins:
+        return []
+    dropped = elements_larger = 0
+    total_sum = contribution_histogram.total_sum()
+    ratio_dropped = []
+    bins = contribution_histogram.bins
+    previous_value = bins[-1].lower  # lower of the largest bin.
+    if contribution_histogram.max_value != bins[-1].lower:
+        # Add ratio for max_value when max_value is not lower in bins.
```
miracvbasaran, line 130: I am not sure I understand this. Why do we need this step?

miracvbasaran (approved):

pipeline_dp/histograms.py
dvadym: It's mostly convenience thing. Those ratios will be used for finding candidates for max_partition_contributed (and other bounds), and it's convenient to have max contribution as one of the candidates

PR #444, opened 2023-05-25
Search parameters grid with constant relative step

RamSaw:
Add functionality to search parameters in a grid with "constant" (from 0.1% to 1%) relative step.


analysis/parameter_tuning.py
```diff
@@ -153,6 +162,28 @@ def _find_candidates(histogram: histograms.Histogram) -> List:
...
+    candidates = histogram.quantiles(QUANTILES_TO_USE)
+    candidates.append(histogram.max_value)
+    candidates = list(set(candidates))  # remove duplicates
+    candidates.sort()
+    return candidates
+
+
+def _find_candidates_constant_relative_step(
+        histogram: histograms.Histogram) -> List:
+    n_max = 1000
```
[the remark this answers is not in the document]
RamSaw: added it as a parameter

analysis/parameter_tuning.py
```diff
@@ -153,6 +162,28 @@ def _find_candidates(histogram: histograms.Histogram) -> List:
...
+def _find_candidates_quantiles(histogram: histograms.Histogram) -> List:
+    QUANTILES_TO_USE = [0.9, 0.95, 0.98, 0.99, 0.995]
+    candidates = histogram.quantiles(QUANTILES_TO_USE)
+    candidates.append(histogram.max_value)
+    candidates = list(set(candidates))  # remove duplicates
+    candidates.sort()
+    return candidates
+
+
+def _find_candidates_constant_relative_step(
```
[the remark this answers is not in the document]
RamSaw: discussed over GVC, the current implementation is ok.

analysis/parameter_tuning.py
```diff
@@ -26,6 +28,9 @@
 from enum import Enum
 import numpy as np
 
+from pipeline_dp.private_contribution_bounds import \
+    _generate_possible_contribution_bounds
```
dvadym, line 32: Nit1: please remove leading '_' in `_generate_possible_contribution_bounds` (it's not private anymore)
Nit2: please keep import in 1 line (it's ok for import to be longer than 80 chars)

analysis/parameter_tuning.py
```diff
@@ -153,6 +163,31 @@ def _find_candidates(histogram: histograms.Histogram) -> List:
         max_contributions_per_partition=linf_bounds)
 
 
+def _find_candidates_quantiles(histogram: histograms.Histogram) -> List:
```
dvadym, line 166: Nit: ->list[int]

dvadym, line 166: Please add 1 line docstring to this function.

analysis/parameter_tuning.py
```diff
@@ -153,6 +163,31 @@ def _find_candidates(histogram: histograms.Histogram) -> List:
...
+    QUANTILES_TO_USE = [0.9, 0.95, 0.98, 0.99, 0.995]
+    candidates = histogram.quantiles(QUANTILES_TO_USE)
+    candidates.append(histogram.max_value)
+    candidates = list(set(candidates))  # remove duplicates
+    candidates.sort()
+    return candidates
+
+
+def _find_candidates_constant_relative_step_func(n_max: int):
```
dvadym, line 175: Please add doctring for this function.

analysis/parameter_tuning.py
```diff
@@ -111,27 +116,32 @@ class TuneResult:
...
+    CONSTANT_RELATIVE_STEP = 2
+
+
 def _find_candidate_parameters(
         hist: histograms.DatasetHistograms,
         parameters_to_tune: ParametersToTune,
-        metric: pipeline_dp.Metrics) -> analysis.MultiParameterConfiguration:
+        metric: pipeline_dp.Metrics,
+        strategy: ParametersSearchStrategy = ParametersSearchStrategy.QUANTILES,
+        n_max: int = None) -> analysis.MultiParameterConfiguration:
```
dvadym, line 129: It's better to specify `max_candidates` from the caller code and that this function would detect how many to generate candidates by each coordinate. 
Example
`max_candidates = 25`
`max_candidates = 25`
  if`
  if` parameters_to_tune.max_partitions_contributed == parameters_to_tune.max_contributions_per_partition == True`
then it takes by 5 candidate by  l0 and linf
but if "max_partitions_contributed == True and max_contributions_per_partition == False` then it takes
25 options for l0.

WDYT?

analysis/parameter_tuning.py
```diff
@@ -153,6 +163,31 @@ def _find_candidates(histogram: histograms.Histogram) -> List:
...
+    candidates = histogram.quantiles(QUANTILES_TO_USE)
+    candidates.append(histogram.max_value)
+    candidates = list(set(candidates))  # remove duplicates
+    candidates.sort()
+    return candidates
+
+
+def _find_candidates_constant_relative_step_func(n_max: int):
+
+    def _find_candidates_constant_relative_step(
```
dvadym, line 177: Nit: it would be nice to simplify code by removing returning function. It looks it can be done by introducing in `_find_candidates_quantiles` function the argument `n_max` . WDYT?

analysis/parameter_tuning.py
```diff
@@ -153,6 +163,31 @@ def _find_candidates(histogram: histograms.Histogram) -> List:
...
+            histogram: histograms.Histogram) -> List:
+        max_value = histogram.max_value
+        # relative step varies from 1% to 0.1%
+        candidates = _generate_possible_contribution_bounds(max_value)
+        if len(candidates) > n_max:
+            candidates = candidates[::math.ceil(len(candidates) / n_max)]
```
dvadym, line 183: Maybe to do what we discussed on VC to return [candidates[int(i*delta)] for i in range(n_max)]

analysis/parameter_tuning.py
```diff
@@ -153,6 +163,31 @@ def _find_candidates(histogram: histograms.Histogram) -> List:
...
+        if candidates[-1] != max_value:
+            candidates.append(max_value)
```
dvadym, line 185: Does it mean that it can be more than n_max elements? maybe candidates[-1] = max_value ?

dvadym (commented):

analysis/parameter_tuning.py
RamSaw: yes, nice idea, implemented and made small refactoring which I think makes code more readable.

analysis/parameter_tuning.py
```diff
@@ -153,6 +179,35 @@ def _find_candidates(histogram: histograms.Histogram) -> List:
...
 
+def _find_candidates_quantiles(histogram: histograms.Histogram,
+                               n_max: int) -> List[int]:
+    """Implementation of ParametersSearchStrategy.QUANTILES strategy."""
+    quantiles_to_use = [0.9, 0.95, 0.98, 0.99, 0.995]
+    candidates = histogram.quantiles(quantiles_to_use)
+    candidates.append(histogram.max_value)
+    candidates = list(set(candidates))  # remove duplicates
+    candidates.sort()
+    return candidates[:n_max]
```
RamSaw, line 190: maybe log a warning because such behaviour can signal about a mistake? becuase n_max > 5 is weird value.

analysis/parameter_tuning.py
```diff
@@ -153,6 +179,35 @@ def _find_candidates(histogram: histograms.Histogram) -> List:
         max_contributions_per_partition=linf_bounds)
```
RamSaw, line 179: btw, which value does it take if we provide None? I mean for tuning we need to specify some value, we can't not specify l0 or linf. Is it specified by a user in this case?

analysis/parameter_tuning.py
```diff
@@ -153,6 +179,35 @@ def _find_candidates(histogram: histograms.Histogram) -> List:
...
 
 
+def _find_candidates_quantiles(histogram: histograms.Histogram,
+                               n_max: int) -> List[int]:
```
dvadym, line 183: n_max -> max_candidates

analysis/tests/parameter_tuning_test.py
```diff
@@ -75,12 +76,114 @@ def test_find_candidate_parameters(
...
+            max_candidates=1000)
+
+        expected_superset = set(
+            list(range(1, 1000, 1)) + list(range(1000, 10000, 10)) +
+            list(range(10000, 100000, 100)) +
+            list(range(100000, 1000000, 1000))).union({999999})
+        self.assertTrue(
+            set(candidates.max_partitions_contributed).issubset(
+                expected_superset))
+        self.assertEqual(len(set(candidates.max_partitions_contributed)),
```
dvadym, line 136: Nit: assertLen

analysis/tests/parameter_tuning_test.py
```diff
@@ -75,12 +76,114 @@ def test_find_candidate_parameters(
...
+            max_contributions_per_partition=True)
+
+        candidates = parameter_tuning._find_candidate_parameters(
+            mock_histograms,
+            parameters_to_tune,
+            pipeline_dp.Metrics.COUNT,
+            ParametersSearchStrategy.CONSTANT_RELATIVE_STEP,
+            max_candidates=10)
+
+        self.assertEqual(10, len(candidates.max_contributions_per_partition))
```
dvadym, line 161: Nit: Please hard code the expected output, here there are only 10 elements, and it will be seen from the output how the algorithm generates them.

analysis/tests/parameter_tuning_test.py
```diff
@@ -75,12 +76,114 @@ def test_find_candidate_parameters(
...
+            max_contributions_per_partition=True)
+
+        candidates = parameter_tuning._find_candidate_parameters(
+            mock_histograms,
+            parameters_to_tune,
+            pipeline_dp.Metrics.COUNT,
+            ParametersSearchStrategy.CONSTANT_RELATIVE_STEP,
+            max_candidates=100)
+
+        self.assertEqual([i for i in range(1, 51)],
```
dvadym, line 184: Nit: list(range(1, 51))

analysis/parameter_tuning.py
```diff
@@ -153,6 +179,35 @@ def _find_candidates(histogram: histograms.Histogram) -> List:
...
+    candidates.sort()
+    return candidates[:n_max]
+
+
+def _find_candidates_constant_relative_step(histogram: histograms.Histogram,
+                                            n_max: int) -> List[int]:
+    """Implementation of ParametersSearchStrategy.CONSTANT_RELATIVE_STEP
+       strategy."""
+    max_value = histogram.max_value
+    # relative step varies from 1% to 0.1%
```
dvadym, line 198: This comment is not understandable w/o context. Could you please write down in more details why candidates in that way are chosen? Something short, a few lines should be fine

analysis/tests/parameter_tuning_test.py
RamSaw: I updated but actually I don't think it is easy to understand why elements are exactly these and not the others.

PR #448, opened 2023-06-07
Utility report per partition size

dvadym:
This PR implements the computation of `UtilityReport` hisogram per partition size.

**Histogram description:**

Hisogram bucket bounds are [0, 1, 10, 20, 50, 100, 200, 500, 1000, ...].
Hisotgram contains  `UtilityReport` (which is aggregated Utility analysis information) for partitions of corresponding size, e.g. it contains  `UtilityReport` for partitions of size [10, 20), `UtilityReport` for partitions of size [20, 50) etc.

**Implementation details:**
1.`UtilityReportHistogram` dataclass with histogram `[partition_size_from, partition_size_to]` -> `UtilityReport`
2. Computing `UtilityReportHistogram` in `perform_utility_analysis`. That's done by extending the existing computation by computation histogram as well.


analysis/metrics.py
```diff
@@ -242,3 +242,19 @@ class UtilityReport:
...
+
+    Attributes:
+        partition_size_from: lower bound of partitions size.
+        partition_size_to: upper bound of partitions size.
+        reports: i-th element stores the result of utility analysis for
+          partitions of size [partition_size_from[i], partition_size_to[i]).
+    """
+    partition_size_from: List[int]
+    partition_size_to: List[int]
+    reports: List[UtilityReport]
```
miracvbasaran, line 260: Optional: Maybe we can have a bin object instead?

```
class UtilityReportBin:
    from: int
    to: int
    report: UtilityReport

class UtilityReportHistogram:
    bins: List[UtilityReportBin]
```

analysis/utility_analysis.py
```diff
@@ -146,3 +161,74 @@ def _pack_per_partition_metrics(
...
+        if lower_bucket_bound is None:
+            # only the report which corresponds to all partitions does not have
+            # the bucket.
+            global_report = report
+        else:
+            histogram_reports.append((lower_bucket_bound, report))
+    if global_report is None:
+        # it should not happen, but it better to process gracefully in case
+        # if it happens and return None.
+        return None
```
miracvbasaran, line 221: Maybe we should throw an exception then?

miracvbasaran (approved):

dvadym (commented):
Thanks for review!

PR #449, opened 2023-06-09
Decrease test flakiness

dvadym:
Expected flakiness is decreased from 2\*10^-3 to 2\*10^-4

### DIFF
--- a/tests/dp_computations_test.py
+++ b/tests/dp_computations_test.py
@@ -69,16 +69,16 @@ class DPComputationsTest(parameterized.TestCase):
                                          expected_beta: float):
         laplace_sample = np.random.laplace(expected_mean, expected_beta,
                                            num_trials)
-        (statistic, pvalue) = stats.ks_2samp(results, laplace_sample)
-        self.assertGreaterEqual(pvalue, 0.001)
+        statistic, pvalue = stats.ks_2samp(results, laplace_sample)
+        self.assertGreaterEqual(pvalue, 10e-4)
 
     def _test_gaussian_kolmogorov_smirnov(self, num_trials: int, results,
                                           expected_mean: float,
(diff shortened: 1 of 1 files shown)


tests/dp_computations_test.py
```diff
@@ -69,16 +69,16 @@ def _test_laplace_kolmogorov_smirnov(self, num_trials: int, results,
                                          expected_beta: float):
         laplace_sample = np.random.laplace(expected_mean, expected_beta,
                                            num_trials)
-        (statistic, pvalue) = stats.ks_2samp(results, laplace_sample)
-        self.assertGreaterEqual(pvalue, 0.001)
+        statistic, pvalue = stats.ks_2samp(results, laplace_sample)
+        self.assertGreaterEqual(pvalue, 10e-4)
```
monsieurmuffin, line 73: Please could you add a comment mentioning what flakiness to expect? 
I.e., if I see in the future that the test fails 20% of time , this information will help me to understand if we have a real issue or such a flakiness is expected.

monsieurmuffin (approved):

tests/dp_computations_test.py
monsieurmuffin: Also, for debugging it might be helpful to know how you calculated the expected pValue and how it maps to the expected flakiness. I.e., if the tests start failing more often than expected, we will need to know this in order to make sure that the test setup is correct.

dvadym: It's equal to a value which we check, that's just a definition of p-value
"The P value is defined as the probability under the assumption of no effect or no difference (null hypothesis), of obtaining a result equal to or more extreme than what was actually observed."

2*10**-4, just becase we do it 2 times :)

dvadym (commented):
Thanks for review!

PR #450, opened 2023-06-09
Small updates of typing and documentation for CustomCombiner

dvadym:
`CustomCombiner` is the way how client can implement custom DP aggregation, with PipelineDP manages all infrastructure stuff (Contribution bounding, Partition selection, budget management etc).

In the initial version of this API, it was assumed that `CustomCombiner` requests the budget only once. But 
`CustomCombiner` might need to request multiple budget objects, for example for computing mean, it's needed to add noise 2 times - for count and sum.  This PR changes that. `CustomCombiner` can use any structure for keeping budgets (e.g. dictionary) rather than 1 budget object.

### DIFF
--- a/pipeline_dp/private_beam.py
+++ b/pipeline_dp/private_beam.py
@@ -16,7 +16,7 @@ import dataclasses
 import typing
 from apache_beam.transforms import ptransform
 from abc import abstractmethod
-from typing import Callable, Optional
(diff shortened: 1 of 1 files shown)


monsieurmuffin (approved):
Thank you so much for the detailed description, it was super-helpful!

dvadym:
Thanks for review!

PR #452, opened 2023-06-14
Renaming in UtilityReportBin

dvadym:
There are also 2 changes in reformatting, which is caused that yapf formatter was updated recently in Github continuous testing system.


monsieurmuffin (approved):
Thank you!

PR #455, opened 2023-06-15
Remove test with runtime warning

RamSaw:
Delete test that produces RuntimeWarning.

Warning:
```
tests/private_contribution_bounds_test.py::L0ScoringFunctionTest::test_score_gaussian_noise_invalid_values_throws_exception
  /home/runner/work/PipelineDP/PipelineDP/pipeline_dp/dp_computations.py:103: RuntimeWarning: invalid value encountered in sqrt
    return np.sqrt(l0_sensitivity) * linf_sensitivity
```

We delete because in some systems this warning is treated as failure.

### DIFF
--- a/tests/private_contribution_bounds_test.py
+++ b/tests/private_contribution_bounds_test.py
@@ -200,7 +200,7 @@ class L0ScoringFunctionTest(unittest.TestCase):
             params, number_of_partitions, l0_histogram)
 
         with self.assertRaises(RuntimeError):
-            l0_scoring_function.score(-1)
+            l0_scoring_function.score(0)
 
 
 class PrivateL0CalculatorTest(unittest.TestCase):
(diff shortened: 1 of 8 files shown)


tests/private_contribution_bounds_test.py
```diff
@@ -202,9 +202,6 @@ def test_score_gaussian_noise_invalid_values_throws_exception(self):
         with self.assertRaises(RuntimeError):
             l0_scoring_function.score(-1)
```
dvadym, line 204: There is still warning

"  /home/runner/work/PipelineDP/PipelineDP/pipeline_dp/dp_computations.py:103: RuntimeWarning: invalid value encountered in sqrt
    return np.sqrt(l0_sensitivity) * linf_sensitivity"

we should remove l0_scoring_function.score(-1)

dvadym (approved):
Thanks!

PR #457, opened 2023-06-16
(Polishing) Dropping return_per_partition argument

dvadym:
This argumen just adds complexity, let's return always per_partition results

### DIFF
--- a/analysis/utility_analysis.py
+++ b/analysis/utility_analysis.py
@@ -59,10 +59,10 @@ def perform_utility_analysis(
         in the result. If not provided, the utility analysis with private
         partition selection will be performed.
     Returns:
-        returns a tuple, with the 1st element is a collection which contains
-          metrics.UtilityReport with one report per each input configuration and
-          the 2nd a collection with elements
-          (partition_key, [metrics.PerPartitionMetrics]).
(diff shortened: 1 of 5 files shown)


analysis/utility_analysis.py
```diff
@@ -59,16 +58,11 @@ def perform_utility_analysis(
...
-        partitions as a 2nd element.
     Returns:
-         if return_per_partition == False:
-            returns collections which contains metrics.UtilityReport with
-            one report per each input configuration
-         else:
-            returns a tuple, with the 1st element as in first 'if' clause and
-            the 2nd a collection with elements
-            (partition_key, [metrics.PerPartitionMetrics]).
+        returns a tuple, with the 1st element is a collection which contains
```
monsieurmuffin, line 62: returns a tuple. Its 1st element is a collection which contains
          metrics.UtilityReport with one report per each input configuration. The 2nd element of the tuple is a collection with elements
          (partition_key, [metrics.PerPartitionMetrics]).

monsieurmuffin (approved):
Thank you for removing the flag! It should make the logic simpler.

analysis/utility_analysis.py
dvadym: Thanks!

PR #460, opened 2023-06-23
Add more candidates if l0/linf candidate sets have fewer candidates than requested

RamSaw:
If l0 candidate set is smaller than requested then we can add more candidates to linf candidate set and vice verca.

### DIFF
--- a/analysis/parameter_tuning.py
+++ b/analysis/parameter_tuning.py
@@ -157,6 +157,20 @@ def _find_candidate_parameters(
         linf_candidates = find_candidates_func(
             hist.linf_contributions_histogram, max_candidates_per_parameter)
         l0_bounds, linf_bounds = [], []
+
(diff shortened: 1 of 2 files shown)


analysis/parameter_tuning.py
```diff
@@ -157,6 +157,20 @@ def _find_candidate_parameters(
         linf_candidates = find_candidates_func(
             hist.linf_contributions_histogram, max_candidates_per_parameter)
         l0_bounds, linf_bounds = [], []
+
+        # if linf or l0 has fewer candidates than requested then we can add more
+        # candidates for the other parameter.
+        if (len(linf_candidates) < max_candidates_per_parameter and
```
dvadym, line 163: It's a cool approach!

dvadym (approved):
Thanks!

PR #466, opened 2023-07-06
Fix computing weighted average for UtilityReports 

dvadym:
`UtilityReport` (which are cross-partition Utility Analysis metrics) are computed as weighted average of per-partition utility analysis metrics, with weights equal to probability that the partition kept. There is a bug that per-partition utility metrics are not multipled on the prob_keep. This PR fixes that.


analysis/cross_partition_combiners.py
```diff
@@ -83,6 +83,12 @@ def _sum_metrics_to_value_error(sum_metrics: metrics.SumMetrics,
         l1=l1,
         rmse_with_dropped_partitions=rmse_with_dropped_partitions,
         l1_with_dropped_partitions=l1_with_dropped_partitions)
+    if keep_prob != 1:
```
miracvbasaran, line 86: Is the multiplication really important for performance here? If not I think we can simply skip (especially given that keep_prob=1 only when public partitions are used (?)

miracvbasaran (approved):

analysis/cross_partition_combiners.py
dvadym: yes, multiplication for performance reasons. And I'd prefer to keep it. `keep_prob = 1` for any large enough partition. I'

PR #467, opened 2023-07-13
Fix scaling partition info

dvadym:
`PartitionInfo` contains information about the expected number of kept partition during the partition selection. But ithere was a bug, `PartitionInfo` was averaged across partitions. This PRs fixes that

### DIFF
--- a/analysis/cross_partition_combiners.py
+++ b/analysis/cross_partition_combiners.py
@@ -258,22 +258,22 @@ def _average_utility_report(report: metrics.UtilityReport,
                             public_partitions: bool,
                             sums_actual: Tuple) -> None:
     """Averages fields of the 'report' across partitions."""
+    if not report.metric_errors:
+        return
     partitions = report.partitions_info
     if public_partitions:
         num_output_partitions = partitions.num_dataset_partitions + partitions.num_empty_partitions
     else:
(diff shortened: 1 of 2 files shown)


analysis/cross_partition_combiners.py
```diff
@@ -258,22 +258,22 @@ def _average_utility_report(report: metrics.UtilityReport,
...
+    if not report.metric_errors:
+        return
     partitions = report.partitions_info
     if public_partitions:
         num_output_partitions = partitions.num_dataset_partitions + partitions.num_empty_partitions
     else:
         num_output_partitions = partitions.kept_partitions.mean
-    _multiply_float_dataclasses_field(report.partitions_info,
-                                      1.0 / num_output_partitions)
-    if report.metric_errors:
```
miracvbasaran, line 268: Why did this line change?

miracvbasaran (approved):

analysis/cross_partition_combiners.py
dvadym: I've moved this check up for simplifying the code. Now, if `metric_errors` are empty then this function is no-op, having that in the beginning makes it more clear

dvadym:
Thanks for review!

PR #469, opened 2023-07-14
(WIP) Parameter tuning

dvadym:


analysis/parameter_tuning.py
```diff
@@ -96,7 +96,7 @@ class TuneOptions:
     parameters_to_tune: ParametersToTune
     partitions_sampling_prob: float = 1
     pre_aggregated_data: bool = False
-    parameters_search_strategy: ParametersSearchStrategy = ParametersSearchStrategy.QUANTILES
+    parameters_search_strategy: ParametersSearchStrategy = ParametersSearchStrategy.CONSTANT_RELATIVE_STEP
```
RamSaw, line 99: so it will make it default to all users and we will change behaviour for them?

analysis/parameter_tuning.py
```diff
@@ -328,18 +342,25 @@ def _convert_utility_analysis_to_tune_result(
...
-            f"Tuning is supported only for Count and Privacy id count, but {metrics[0]} given."
-        )
+    if not metrics:
+        # Empty metrics means tuning for select_partitions.
+        if is_public_partitions:
+            # Empty metrics means that partition selection tuning is performed.
+            raise ValueError("Empty metrics means tuning of partition selection"
+                             " but public partitions were provided.")
+    elif len(metrics) > 1:
+        raise ValueError(
```
RamSaw, line 355: btw, do we plan to support multiple metrics?

analysis/tests/parameter_tuning_test.py
```diff
@@ -263,6 +319,42 @@ def test_tune_count(self):
         self.assertEqual(utility_reports[0].metric_errors[0].metric,
                          pipeline_dp.Metrics.COUNT)
 
+    def test_select_partitions(self):
+        # Arrange.
+        input = [(i % 10, f"pk{i/10}") for i in range(10)]
```
RamSaw, line 324: maybe describe in comment how it looks like to not ask people reading this use their cognitive forces to understand partitions and pids dist?

RamSaw: or just hardcode these 10 values

analysis/tests/parameter_tuning_test.py
```diff
@@ -263,6 +319,42 @@ def test_tune_count(self):
...
+
+        tune_result = list(tune_result)[0]
+
+        self.assertEqual(tune_options, tune_result.options)
+        self.assertEqual(contribution_histograms,
+                         tune_result.contribution_histograms)
+        utility_reports = tune_result.utility_reports
+        self.assertLen(utility_reports, 1)
+        self.assertIsInstance(utility_reports[0], metrics.UtilityReport)
+        self.assertIsNone(utility_reports[0].metric_errors)
```
RamSaw, line 356: should we assert that something is present? I think current assertions do not verify that for `select_partitions` we calculated something.

RamSaw (approved):

analysis/tests/parameter_tuning_test.py
dvadym: Thanks, good point. Done

dvadym (commented):
Thanks for review!

PR #471, opened 2023-07-18
Make histogram.max_value to be method instead of property

dvadym:
This is done mainly to simplify mocking, while there is no particular reason, namely mocking of properties are complicated, it requires mocking of classes, like

`setattr(mock_l0_histogram.__class__, 'max_value', 6)`

But such mocking is bad, because it changes value of `max_value` in **other tests** :(. For example the `test_count` in parameter_tuning.py was incorrect because of incorrect `max_value`.

The simplest solution is to make `max_value` to be method, which seems to be not worse than property.

### DIFF
--- a/analysis/tests/parameter_tuning_test.py
+++ b/analysis/tests/parameter_tuning_test.py
@@ -63,9 +63,10 @@ class ParameterTuning(parameterized.TestCase):
     ):
         mock_l0_histogram = histograms.Histogram(None, None)
         mock_l0_histogram.quantiles = mock.Mock(return_value=[1, 1, 2])
-        setattr(mock_l0_histogram.__class__, 'max_value', 6)
+        mock_l0_histogram.max_value = mock.Mock(return_value=6)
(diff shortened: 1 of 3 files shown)


analysis/tests/parameter_tuning_test.py
```diff
@@ -245,15 +249,15 @@ def test_tune_count(self):
         # Assert.
         tune_result, per_partition_utility_analysis = result
         per_partition_utility_analysis = list(per_partition_utility_analysis)
-        self.assertLen(per_partition_utility_analysis, 40)
```
RamSaw, line 248: so test was incorrect in the past and 40 and 4 below are incorrect expectations?

RamSaw (approved):

analysis/tests/parameter_tuning_test.py
dvadym: `tune()` incorrectly worked, because `max_value` was incorrect.

PR #474, opened 2023-07-24
Computing privacy_id_count and count in PerPartitionUtilityResults

dvadym:
This PR contains:
1. Introducing `StatisticsCombiner` per-partition utility analysis combiner, for computing count, privacy_id_count (in future it can be extended to more statistics)
1. Adding  `metrics.Statistics(privacy_id_count, count)` to `PerPartitionMetrics`.
1. Plumbing work to fill` PerPartitionMetrics.statistincs`


analysis/combiners.py
```diff
@@ -320,6 +317,22 @@ def create_accumulator(
         return super().create_accumulator(data)
 
 
+class StatisticsCombiner(UtilityAnalysisCombiner):
+    """A combiner for computing per-partition statistics (count etc)."""
```
miracvbasaran, line 321: We already have other combiners for computing per-partition metrics, no? How is this different from the others? I believe having more descriptive documentation here would help.

analysis/metrics.py
```diff
@@ -59,9 +59,16 @@ class SumMetrics:
     noise_kind: pipeline_dp.NoiseKind
 
 
+@dataclass
+class Statistics:
```
miracvbasaran, line 63: I don't think this is very descriptive. Maybe something like `RawStatistics` could be more appropriate?

analysis/utility_analysis.py
```diff
@@ -146,12 +146,14 @@ def _pack_per_partition_metrics(
     """
     n_metrics = len(utility_result) // n_configurations
 
+    statistics = utility_result[0]
     # Create 'result' with empty elements.
-    empty_partition_metric = lambda: metrics.PerPartitionMetrics(1, [])
+    empty_partition_metric = lambda: metrics.PerPartitionMetrics(
+        1, copy.copy(statistics), [])
```
miracvbasaran, line 152: If we don't explicitly copy, does it pass by reference?

miracvbasaran (changes requested):

analysis/utility_analysis.py
dvadym: yes. I've thought again it's actually fine. I've removed copying

analysis/metrics.py
dvadym: Thanks, it's a good point. I've renamed to RawStatistics

analysis/combiners.py
dvadym: We don't always compute both count/privacy_id_count, e.g. when metrics = [COUNT], privacy_id_count is not computed, that's why `StatisticsCombiner` was introduced.

But, it seems to be much details for this class docstring.

dvadym (commented):

dvadym:
Thanks for review!

PR #476, opened 2023-07-26
Adding input params validation for PLD accounting

dvadym:


pipeline_dp/dp_engine.py
```diff
@@ -495,6 +498,29 @@ def _check_calculate_private_contribution_bounds_params(
...
+                                                   pipeline_dp.Metric],
+                                               custom_combiner: bool) -> None:
+        if isinstance(self._budget_accountant,
+                      pipeline_dp.NaiveBudgetAccountant):
+            # All aggregations support NaiveBudgetAccountant.
+            return
+        if not is_public_partition:
+            raise NotImplementedError("PLD budget accounting does not support "
+                                      "private partition selection")
+        supported_metrics = [
```
miracvbasaran, line 512: I wonder if it makes sense to make this a constant in the PLD file.

miracvbasaran (approved):

pipeline_dp/dp_engine.py
dvadym: It's actually not PLD limitations, but Var and Quantile metrics. And there is no PLD file. So I think, this is the best place :)

dvadym (commented):
Thanks for review!

PR #477, opened 2023-07-31
Small refactoring in DPEngine

dvadym:
This PRs implements extracting `(privacy_id, partition_key, value)` from raw records as a first computation in `DPEngine.aggregate`. Previously it was done on later stages.

 This has performance improvements, for cases when the input record are large.

### DIFF
--- a/pipeline_dp/dp_engine.py
+++ b/pipeline_dp/dp_engine.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 """DP aggregations."""
 import functools
-from typing import Any, Optional, Tuple
+from typing import Any, Callable, Optional, Tuple
 
 import pipeline_dp
 from pipeline_dp import budget_accounting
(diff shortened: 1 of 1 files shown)


pipeline_dp/dp_engine.py
```diff
@@ -273,15 +273,13 @@ def sample_unique_elements_fn(pid_and_pks):
 
         return col
 
-    def _drop_partitions(self, col, partitions,
-                         data_extractors: pipeline_dp.DataExtractors):
+    def _drop_partitions(self, col, partitions, partition_extractor: Callable):
```
sushkoy, line 276: nit: fix the pydoc, public_partitions > partitions.

dvadym: Done

sushkoy (approved):

PR #479, opened 2023-08-07
Fix Utility analysis on preagregation

dvadym:
On [PR](https://github.com/OpenMined/PipelineDP/pull/477) it was changed the order in which extracting columns from the data was performed. As a result it broke running utility analysis on pre-aggregated data. This PR fixed that


miracvbasaran (approved):

dvadym:
Thanks for review!

PR #480, opened 2023-08-08
Add support to compute linf sum contributions histogram

RamSaw:
Linf sum contributions histogram is a histogram where bins (lowers) are X_1, X_2, ..., X_n, where X_i is sum of values that have same (pid, pk)-pair (rounded to the bucket). Count of such histogram is number of (pid, pk) that have this sum. Sum of such histogram is sum of all sums in this bucket.

### DIFF
--- a/pipeline_dp/dataset_histograms/computing_histograms.py
+++ b/pipeline_dp/dataset_histograms/computing_histograms.py
@@ -322,7 +322,7 @@ def _compute_linf_sum_contributions_histogram(
     lowers = backend.flatten(lowers, "Flatten lowers")
 
     return _compute_frequency_histogram_helper_with_lowers(
-        col, backend, hist.HistogramType.LINF_CONTRIBUTIONS, lowers)
+        backend, col, hist.HistogramType.LINF_CONTRIBUTIONS, lowers)
 
 
 def _compute_partition_count_histogram(
--- a/pipeline_dp/dataset_histograms/computing_histograms.py
+++ b/pipeline_dp/dataset_histograms/computing_histograms.py
@@ -322,7 +322,7 @@ def _compute_linf_sum_contributions_histogram(
     lowers = backend.flatten(lowers, "Flatten lowers")
 
     return _compute_frequency_histogram_helper_with_lowers(
-        backend, col, hist.HistogramType.LINF_CONTRIBUTIONS, lowers)
+        backend, col, hist.HistogramType.LINF_SUM_CONTRIBUTIONS, lowers)
 
 
 def _compute_partition_count_histogram(
--- a/pipeline_dp/dataset_histograms/computing_histograms.py
+++ b/pipeline_dp/dataset_histograms/computing_histograms.py
@@ -66,7 +66,7 @@ def _compute_frequency_histogram(col, backend: pipeline_backend.PipelineBackend,
 
     # Combiner elements to histogram buckets of increasing sizes. Having buckets
     # of width = 1 is not scalable.
-    return _compute_frequency_histogram_helper(backend, col, name)
+    return _compute_frequency_histogram_helper(col, backend, name)
 
 
 def _compute_weighted_frequency_histogram(
@@ -90,11 +90,11 @@ def _compute_weighted_frequency_histogram(
 
     # Combiner elements to histogram buckets of increasing sizes. Having buckets
     # of width = 1 is not scalable.
-    return _compute_frequency_histogram_helper(backend, col, name)
+    return _compute_frequency_histogram_helper(col, backend, name)
 
 
 def _compute_frequency_histogram_helper(
-        backend: pipeline_backend.PipelineBackend, col,
+        col, backend: pipeline_backend.PipelineBackend,
         name: hist.HistogramType):
     """Computes histogram of element frequencies in collection.
 
@@ -119,11 +119,11 @@ def _compute_frequency_histogram_helper(
 
     col = backend.map_tuple(col, _map_to_frequency_bin, "To FrequencyBin")
     # (lower_bin_value, hist.FrequencyBin)
-    return _convert_frequency_bins_into_histogram(backend, col, name)
+    return _convert_frequency_bins_into_histogram(col, backend, name)
 
 
 def _compute_frequency_histogram_helper_with_lowers(
-        backend: pipeline_backend.PipelineBackend, col,
+        col, backend: pipeline_backend.PipelineBackend,
         name: hist.HistogramType, lowers_col):
     """Computes histogram of element frequencies in collection.
 
@@ -154,11 +154,11 @@ def _compute_frequency_histogram_helper_with_lowers(
                                        (lowers_col,), "To FrequencyBin")
     # (lower_bin_value, hist.FrequencyBin)
 
-    return _convert_frequency_bins_into_histogram(backend, col, name)
+    return _convert_frequency_bins_into_histogram(col, backend, name)
 
 
 def _convert_frequency_bins_into_histogram(
-        backend: pipeline_backend.PipelineBackend, col, name):
+        col, backend: pipeline_backend.PipelineBackend, name):
     """Converts (lower_bin_value, hist.FrequencyBin) into histogram.
 
     The input collection is not expected to have frequency bins reduced per
@@ -320,7 +320,7 @@ def _compute_linf_sum_contributions_histogram(
         col, NUMBER_OF_BUCKETS_IN_LINF_SUM_CONTRIBUTIONS_HISTOGRAM, backend)
 
     return _compute_frequency_histogram_helper_with_lowers(
-        backend, col, hist.HistogramType.LINF_SUM_CONTRIBUTIONS, lowers)
+        col, backend, hist.HistogramType.LINF_SUM_CONTRIBUTIONS, lowers)
 
 
 def _min_max_lowers(col, number_of_buckets,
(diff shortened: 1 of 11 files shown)


pipeline_dp/dataset_histograms/computing_histograms.py
```diff
@@ -99,8 +110,9 @@ def _compute_frequency_histogram_helper(
     """
     col = backend.map_tuple(
         col, lambda n, f:
-        (_to_bin_lower(n),
-         hist.FrequencyBin(lower=_to_bin_lower(n), count=f, sum=f * n, max=n)),
+        (_to_bin_lower_logarithmic(n),
```
RamSaw, line 113: should we calculate it only once?

pipeline_dp/pipeline_backend.py
```diff
@@ -470,6 +475,13 @@ def flat_map(self, col, fn, stage_name: str = None):
     def map_tuple(self, col, fn, stage_name: str = None):
         return map(lambda x: fn(*x), col)
 
+    def map_tuple_with_side_inputs(self,
+                                   col,
+                                   fn,
+                                   side_input_cols,
+                                   stage_name: str = None):
+        return map(lambda x: fn(*x, *side_input_cols), col)
```
RamSaw, line 483: so fn will take all side input arguments

pipeline_dp/pipeline_functions.py
```diff
@@ -97,3 +97,26 @@ def create_key_fn(key):
...
+                stage_name: str):
+    return choose_element(backend, col, min, stage_name)
+
+
+T = TypeVar('T')
+
+
+def choose_element(backend: pipeline_backend.PipelineBackend, col,
+                   choosing_function: Callable[[T, T], T], stage_name: str):
+    return backend.values(
```
RamSaw, line 117: here, I have question about efficiency and scalability. My first implementation was `backend.map(backend.to_list(col), lambda l: max(l))` but this is obviously bad and won't fit into the memory.

This implementation seems okaish and will work on Beam (https://beam.apache.org/documentation/transforms/java/aggregation/combine/) but there is also a dedicated way to get max and min (https://beam.apache.org/documentation/transforms/python/aggregation/max/) by using Combine.Globally(). I wonder whether we should implement it framework agnostic and rely on `reduce_per_key` implementation or define max/min functions in backend interface and implement them separately for each framework. WDYT?

pipeline_dp/dataset_histograms/computing_histograms.py
```diff
@@ -403,7 +502,7 @@ def _compute_linf_contributions_histogram_on_preaggregated_data(
     """
     linf = backend.map_tuple(
         col,
-        lambda _, x: x[0],  # x is (count, sum, n_partitions)
+        lambda _, x: x[0],  # x is (count, sum, n_partitions, n_contributions)
```
RamSaw, line 505: Here I am confused. In some places it is written `x is (count, sum, n_partitions, n_contributions)`, in some places `n_contributions` is omitted.

pipeline_dp/dataset_histograms/computing_histograms.py
```diff
@@ -117,6 +129,46 @@ def bins_to_histogram(bins):
     return backend.map(col, bins_to_histogram, "To histogram")
 
 
+def _compute_frequency_histogram_helper_with_lowers(
```
RamSaw, line 132: I will about better design to avoid this code duplication, for now it is just a draft

pipeline_dp/pipeline_backend.py
dvadym: Yeah, it looks reasonable for me

pipeline_dp/pipeline_functions.py
dvadym: That's a good question. I agree that it makes sense to compute with `reduce_per_key`.
What do you think about computing both min and max at one step (which should be better from performance point of view), namely

```
# let col has elements floats
col = backend.map(col, lambda x: (None,(x, x))) # None is dummy key
col = backend.reduce_per_key(col, lambda x, y: (min(x[0], y[0]), min(x[1], y[1])) 
```
then no need to flatten and combine min and max later.

WDYT?

dvadym (commented):
Thanks, it looks great. I left some comments

pipeline_dp/pipeline_functions.py
RamSaw: hm, this approach is definitely the most efficient one we can do, however I wonder whether Beam will be smart and fuse these 2 operation, so each element of the collection will be processed only once and min/max functions will be applied with one read. The advantage is that we have min and max separately and it makes interface is cleaner.

However, since this is the first case when we need min and max and also most probably in our cases we will need always both, I guess we should save time checking whether beam fuses these 2 operations or not and just implement it as you proposed. Will do it.

pipeline_dp/dataset_histograms/computing_histograms.py
RamSaw: I meant that in the next line we call again _to_bin_lower(n) with the same argument. We can cache it or just store in variable. The second option is better because we know that n will be always different.

pipeline_dp/dataset_histograms/computing_histograms.py
RamSaw: There are more places, I will update them too.

pipeline_dp/dataset_histograms/computing_histograms.py
dvadym: ah, good point! Yeah, let's compute it only once. Probably it would more readable to extract lambda in an internal function.

pipeline_dp/dataset_histograms/computing_histograms.py
RamSaw: updated.

pipeline_dp/dataset_histograms/computing_histograms.py
```diff
@@ -346,7 +470,7 @@ def _compute_l0_contributions_histogram_on_preaggregated_data(
 
     Args:
         col: collection with a pre-aggregated dataset, each element is
-        (partition_key, (count, sum, n_partitions)).
+        (partition_key, (count, sum, n_partitions, n_contributions)).
```
RamSaw, line 473: Why `n_partitions` isn't always 1? if it is statistic per one partition

and why `count` is not the same as `n_contributions`?

pipeline_dp/pipeline_backend.py
```diff
@@ -377,6 +393,10 @@ def map(self, rdd, fn, stage_name: str = None):
             return self._sc.parallelize(rdd).map(fn)
         return rdd.map(fn)
 
+    def map_with_side_inputs(self, rdd, fn, side_input_cols, stage_name: str):
+        raise NotImplementedError("map_with_side_inputs "
```
RamSaw, line 397: I tried to find but didn't find any information how side inputs work in apache spark and whether they are supported at all. Do you know how it works in Spark?

pipeline_dp/pipeline_backend.py
```diff
@@ -700,6 +728,14 @@ def map(self, col, fn, stage_name: typing.Optional[str] = None):
                                       chunksize=self.chunksize,
                                       **self.pool_kwargs)
 
+    def map_with_side_inputs(self,
+                             col,
+                             fn,
+                             side_input_cols,
+                             stage_name: typing.Optional[str] = None):
+        side_inputs = [list(side_input) for side_input in side_input_cols]
```
RamSaw, line 736: not sure about it, whether we should materalize collections in multiproclocal this way.

pipeline_dp/dataset_histograms/computing_histograms.py
```diff
@@ -229,6 +293,57 @@ def _compute_linf_contributions_histogram(
...
+    returns bin lowers that equally divide the range in `number_of_buckets`.
+
+    Args:
+        col: collection of numbers.
+        backend: PipelineBackend to run operations on the collection.
+    Returns:
+        collection of lowers (float,).
+    """
+    min_max_values = pipeline_functions.min_max_elements(
+        backend, col, "Min and max value in dataset")
```
dvadym, line 339: Nit: add comment about `min_max_values` contents (like 1 element collection etc)

pipeline_dp/dataset_histograms/computing_histograms.py
```diff
@@ -229,6 +293,57 @@ def _compute_linf_contributions_histogram(
...
+    return backend.flat_map(
+        min_max_values,
+        lambda min_max: numpy.linspace(min_max[0], min_max[1],
+                                       (number_of_buckets + 1))[:-1],
+        "flatMap to lowers")
```
dvadym, line 344: Isn't it better to return a collection with 1 element  - np.array with lower? In that case we have guaranetees on the order of the elements (we need guarantees to run binary search)

pipeline_dp/dataset_histograms/computing_histograms.py
```diff
@@ -403,14 +527,44 @@ def _compute_linf_contributions_histogram_on_preaggregated_data(
...
+        1 element collection, which contains the computed histograms.Histogram.
+    """
+    col = backend.map_tuple(
+        col,
+        lambda _, x: x[1],  # x is (count, sum, n_partitions, n_contributions)
+        "Extract sum per partition contribution")
+    # col: (float,) where each element is the sum of values the
+    # corresponding privacy_id contributes to the partition.
+    col = backend.to_multi_transformable_collection(col)
+    lowers = _min_max_lowers(
```
dvadym, line 561: Nit: please add comment on the contents of `lowers`

dvadym (commented):
Thanks, it looks good! I left some comments

pipeline_dp/dataset_histograms/computing_histograms.py
RamSaw: ah, makes sense, thank you!

pipeline_dp/pipeline_backend.py
RamSaw: ack

dvadym (approved):
Thanks!

PR #483, opened 2023-08-24
Remove quantiles strategy

RamSaw:
Quantiles strategy which means picking up top 5 quantiles showed poor results, worse then constant relative step strategy. Therefore we remove it. From now on we will use only constant relative step strategy.


dvadym (approved):
Thanks!

PR #486, opened 2023-08-31
Add upper bound into frequency bin of a histogram.

RamSaw:
Add upper bound into the frequency bin data class that is stored in histograms. It is necessary because histograms omit empty bins and therefore taking next bin won't provide information about the size of the current bin.


pipeline_dp/dataset_histograms/histograms.py
```diff
@@ -27,24 +27,38 @@ class FrequencyBin:
...
+        [lower; upper). It can be included only for the last bin.
       count: the number of elements in the bin.
       sum: the sum of elements in the bin.
       max: the maximum element in the bin, which is smaller or equal to the
-       upper-1.
+        upper.
     """
     lower: Union[int, float]
+    upper: Union[int, float]
+    upper_included: bool
```
dvadym, line 42: `upper_included` feels not very neccessary, it's set to true only on the last bin in sum histogram.
WDYT about the following:
1. Put `lower`/`upper` fields in `Histogram` (which seems to be useful independently of this)
2. When bin.upper == histogram.upper it would automatically mean that upper_included

Of course another option is to assume that `upper_included = False` always, and to create additional bin which would include upper bound of `Histogram` (with `bin.lower = Histogram.upper`,` bin.upper = bin.lower + small_number`)

WDYT?

dvadym (commented):
Thanks!

pipeline_dp/dataset_histograms/histograms.py
RamSaw: I took a look and I think putting into histogram lower/upper will require quite some refactoring as we will need to propagate these values in PCollections. However, I think we can even avoid it at all because we will know the bounds always:
1. for int histogram we just take the last bin for upper and 1 as lower, as we don't have clear range in the histogram, it is kind of [1; +inf)
2. for float (sum) histogram we take first and last bins, they will be always present and then their lower and upper will define the histogram range (and only in this case for the last bin of the float hist the upper will be included)

And in the end, looks like we need upper inside bins more and for the histogram as a whole we don't really need it as for now and in general we don't care much whether the value is included or not, because when I removed now upper_included no tests failed.

WDYT?
Take a look, I left some comments in the code that explain that.

pipeline_dp/dataset_histograms/histograms.py
```diff
@@ -66,7 +77,15 @@ class HistogramType(enum.Enum):
 
 @dataclass
 class Histogram:
-    """Represents a histogram over integers."""
+    """Represents a histogram over integers.
+
+    To calculate range of the histogram (i.e. min and max values on axis X)
```
dvadym, line 82: Let's introduce definitions of histograms:

There are 2 types of histograms: 
1. Integer histogram, which is used for counts and bins are spaced logarithmically (see `computing_histograms. _to_bin_lower_logarithmic` for details how bins are generated )
2. Floating histograms, which is used for sums per partition. Which has the same size bins.

Please introduce property `IsInteger`  for detecting histogram type (this property can be evalualted based on Histogram.name)

pipeline_dp/dataset_histograms/histograms.py
```diff
@@ -66,7 +77,15 @@ class HistogramType(enum.Enum):
...
+    you can do the following:
+      * if it is an int histogram, min is 1 and max is bins[-1].upper (or
```
dvadym, line 84: Is it hard to introduce lower/upper field to the histogram? (for count lower = 1, upper = None, for sum lower and upper based on computation)

dvadym (commented):
Thanks, it looks great. I left some comments.

pipeline_dp/dataset_histograms/histograms.py
RamSaw: what do you think about making them properties of histogram class that are calculated? I think it is a clean way and the easiest one. If we want to make them proper fields and initialize during creation then we need to change `_convert_frequency_bins_into_histogram` method and duplicate there is_integer method because we need to determine the type of histogram (int or float).

pipeline_dp/dataset_histograms/histograms.py
RamSaw: added, wdyt? Not sure, maybe we should make it the property of the `HistogramType` enum and not the property of Histogram class? 
and whether it should be bool. Better maybe to make it another enum (e.g. HistogramValueType) or return just `type(bins[0].lower)` or put as property of `HistogramType` enum (e.g. `HistogramType.L0_CONTRIBUTIONS.value_type` will return `int`) and then implement `lower` and `upper` accordingly.

I think "`HistogramType.L0_CONTRIBUTIONS.value_type` will return `int`" is the best interface.

dvadym: About enum, it might be a good a solution, but for me it's not that clear how excatly it will evolve. E.g. we might want to have histogram computed in different way - uniform (as sum) or logarithmic (as others) and maybe others type of logarithmic. But it's not that clear. So I'd say having bool property is good, because it's simple. 

The same with type of `HistogramType`, we don't know how it will evolve, in principle we might want to have different of histogram generation for the same type. Property of `Histogram` seems to be good and simple default.

`type(bins[0].lower)` is not reliable enough in Python, since Python doesn't have static type system, and in principle that it's double/int, bins[0].lower is int/float is not implementation dependent.

pipeline_dp/dataset_histograms/histograms.py
dvadym: Can we use the same logic and to create fields? It would be nice to have fields no properties, since we like to convert those dataclasses to protos.

dvadym (commented):
It looks that the last change are not pushed, could you please push them?

pipeline_dp/dataset_histograms/histograms.py
RamSaw: done, wdyt?

pipeline_dp/dataset_histograms/histograms.py
RamSaw: I see. implemented, wdyt?

pipeline_dp/dataset_histograms/histograms.py
RamSaw: if lgtm in general, I will polish the comments and add tests.

pipeline_dp/dataset_histograms/histograms.py
```diff
@@ -66,9 +77,39 @@ class HistogramType(enum.Enum):
...
+          histograms (i.e. +inf) and for floating histograms upper exists and is
+          always included.
+    """
     name: HistogramType
     bins: List[FrequencyBin]
+    lower: Union[None, int, float] = field(init=False)
+    upper: Union[None, float] = field(init=False)
+
+    def __post_init__(self):
+        if len(self.bins) == 0:
```
dvadym, line 95: Nice to do that in `__post_init__` and nice trick to use `field(init=False)` (I didn't know that)

pipeline_dp/dataset_histograms/histograms.py
```diff
@@ -66,9 +77,39 @@ class HistogramType(enum.Enum):
 
 @dataclass
 class Histogram:
-    """Represents a histogram over integers."""
+    """Represents a histogram over numbers.
+
+    Attributes:
+        lower: lower bound of the whole histogram, always included and always
```
dvadym, line 83: Nit: please add doscrings to name and bins

pipeline_dp/dataset_histograms/histograms.py
dvadym: LGTM, thanks!

pipeline_dp/dataset_histograms/histograms.py
RamSaw: thanks! :)

dvadym (approved):
Thanks!

PR #493, opened 2023-10-12
DataFrame API: QueryBuilder

dvadym:
This PR implements `QueryBuilder` class, which implements builder pattern for DataFrame queries.  `Query` will run DP aggregation (it will be implemented in the following PR). `QueryBuilder` and `Query` are agnostic to the type of DataFrames (Pandas, Spark, Beam).

### DIFF
--- a/pipeline_dp/dataframes.py
+++ b/pipeline_dp/dataframes.py
@@ -139,7 +139,7 @@ class QueryBuilder:
         """Adds count to the query.
 
         Args:
-            name: the name output column.
+            name: the name of the output column.
         """
         if self._groupby_column is None:
             raise ValueError(
@@ -160,7 +160,7 @@ class QueryBuilder:
         Args:
             column: column to perform summation
             min_value, max_value: capping limits to each value.
-            name: the name output column.
+            name: the name of the output column.
         """
         if self._groupby_column is None:
             raise ValueError(
--- a/pipeline_dp/dataframes.py
+++ b/pipeline_dp/dataframes.py
@@ -1,0 +1,130 @@
+# Copyright 2023 OpenMined.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Computing DP aggregations on (Pandas, Spark, Beam) Dataframes."""
+
+from dataclasses import dataclass
+from typing import Optional, List, Dict, Iterable
+
+import pipeline_dp
+
+
+@dataclass
+class Columns:
+    privacy_key: str
+    partition_key: str
+    value: Optional[str]
+
+
+@dataclass
+class ContributionBounds:
+    max_partitions_contributed: Optional[int] = None
+    max_contributions_per_partition: Optional[int] = None
+    min_value: Optional[float] = None
+    max_value: Optional[float] = None
+
+
+@dataclass
+class Budget:
+    epsilon: float
+    delta: float = 0
+
+    def __post_init__(self):
+        if self.epsilon <= 0:
+            raise ValueError(f"epsilon must be positive, not {self.epsilon}.")
+        if self.delta < 0:
+            raise ValueError(f"delta must be non-negative, not {self.delta}.")
+
+
+class Query:
+
+    def __init__(self, df, columns: Columns, metrics: Dict[pipeline_dp.Metric,
+                                                           List[str]],
+                 contribution_bonds: ContributionBounds, public_keys):
+        self._df = df
+        self._columns = columns
+        self._metrics = metrics
+        self._contribution_bonds = contribution_bonds
+        self._public_partitions = public_keys
+        self._expain_computation_report = None
+
+    def run_query(self,
+                  budget: Budget,
+                  noise_kind: Optional[pipeline_dp.NoiseKind] = None):
+        raise NotImplementedError("Running query is not yet implemented")
+
+
+class QueryBuilder:
+
+    def __init__(self, df, privacy_key_column: str):
+        if privacy_key_column not in df.scheme:
+            raise ValueError(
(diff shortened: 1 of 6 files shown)


pipeline_dp/dataframes.py
```diff
@@ -0,0 +1,189 @@
...
+                 contribution_bounds: ContributionBounds, public_keys):
+        self._df = df
+        self._columns = columns
+        self._metrics = metrics
+        self._contribution_bounds = contribution_bounds
+        self._public_partitions = public_keys
+
+    def run_query(self,
+                  budget: Budget,
+                  noise_kind: Optional[pipeline_dp.NoiseKind] = None):
```
RamSaw, line 66: should be Laplace by default? we can't run without noise.

pipeline_dp/dataframes.py
```diff
@@ -0,0 +1,189 @@
...
+            raise ValueError(f"delta must be non-negative, not {self.delta}.")
+
+
+class Query:
+    """DP Query on DataFrames.
+
+    Warning: Don't create it directly, use QueryBuilder().
+    """
+
+    def __init__(self, df, columns: Columns, metrics: Dict[pipeline_dp.Metric,
```
RamSaw, line 55: `self._metrics = {}  # map from pipeline_dp.Metric -> output column name` from below.
Then metrics here should be of type `Dict[Metric, str]`.

pipeline_dp/dataframes.py
```diff
@@ -0,0 +1,189 @@
...
+        self._df = df
+        self._privacy_unit_column = privacy_unit_column
+        self._groupby_column = None
+        self._value_column = None
+        self._metrics = {}  # map from pipeline_dp.Metric -> output column name
+        self._contribution_bounds = ContributionBounds()
+        self._public_keys = None
+
+    def groupby(self,
+                column: str,
```
RamSaw, line 107: how will the support for grouping by more columns be supported? Will we need to call it once again?

RamSaw: ah, I see, it can be called only once.

But maybe then make it a list? so that we have correct interface from the beginning. I guess the same then applies to privacy unit. But up to you if it is difficult to make right now.

pipeline_dp/dataframes.py
```diff
@@ -0,0 +1,189 @@
...
+                *,
+                max_groups_contributed: int,
+                max_contributions_per_group: int,
+                public_keys=Iterable[Any]) -> 'QueryBuilder':
```
RamSaw, line 111: interesting how it works, but we probably need to write `public_keys: Iterable[Any]` and make it default none, so the type is actually `Optional[Iterable[Any]]`.

pipeline_dp/dataframes.py
```diff
@@ -0,0 +1,189 @@
...
+            raise ValueError(
+                "Global aggregations are not supported. Use groupby.")
+        if pipeline_dp.Metrics.COUNT in self._metrics:
+            raise ValueError("count can be counted only once.")
+        self._metrics[pipeline_dp.Metrics.COUNT] = name
+        return self
+
+    def sum(self,
+            column: str,
+            *,
```
RamSaw, line 154: what does `*` here mean?

pipeline_dp/dataframes.py
```diff
@@ -0,0 +1,189 @@
...
+        self._contribution_bounds.min_value = min_value
+        self._contribution_bounds.max_value = max_value
+        return self
+
+    def build_query(self) -> Query:
+        """Builds the DP query."""
+        if self._groupby_column is None:
+            raise NotImplementedError(
+                "Global aggregations are not implemented yet. Call groupby.")
+        metrics = list(self._metrics.keys())
```
RamSaw, line 184: should we assert that metrics is not empty?

also, Query expects dictionary and we take here keys only

tests/dataframes_test.py
RamSaw, line 1: nit: maybe add empty lines betwee arrange/act/assert blocks?

tests/dataframes_test.py
```diff
@@ -0,0 +1,125 @@
...
+        builder.groupby("group_key",
+                        max_groups_contributed=1,
+                        max_contributions_per_group=1)
+        with self.assertRaisesRegex(ValueError,
+                                    "groupby can be called only once"):
+            builder.groupby("group_key",
+                            max_groups_contributed=1,
+                            max_contributions_per_group=1)
+
+    def test_count_query(self):
```
RamSaw, line 67: will we remove this test once we implemented run_query and will be able to test it through the public API? Then maybe add a comment or todo

tests/dataframes_test.py
RamSaw, line 1: I would add a test for public_keys

RamSaw (changes requested):
LGTM in general, left some comments that are probably tiny bugs and some nit proposals and questions/discussions.

pipeline_dp/dataframes.py
dvadym: Thx, good point. I've added check

tests/dataframes_test.py
dvadym: I'd say that we can keep it. This is test for `QueryBuilder` (i.e. that builder works correctly). `Query` test will be test for `run_query` (i.e. that the query is translated correctly to `DPEngine` API)

dvadym (commented):
Thanks for review! I've addressed comments. PTAL

pipeline_dp/dataframes.py
RamSaw: but we still get keys from metrics and pass to Query metrics as list. But the declartion of Query consturctor says that metrics is Dict[Metric, List[str]]. List is strange too. I think it should be Dict[Metric, str] and we don't need to do anything with self._metrics, just pass it as it is.

tests/dataframes_test.py
RamSaw: sg :)

RamSaw (approved):
One comment to address, but nothing else, so I give my approval to avoid one more loop of review :)

pipeline_dp/dataframes.py
dvadym: Yes, thanks

pipeline_dp/dataframes.py
dvadym: I though initially to make it automatic (if None is set). Let's go simple and set it to Laplace

pipeline_dp/dataframes.py
dvadym: Thanks, done

dvadym (commented):
Thanks for review!

PR #495, opened 2023-10-16
Spark Dataframe API

dvadym:
This PR implements running `DataFrame` queries for Spark `DataFrame`. It contains
1. `SparkConverter` for converting between `RDD` and `DataFrame`.
2. `Query.run_query()` implementation, which maps `Query` to `DPEngine.aggregate` call

### DIFF
--- a/pipeline_dp/dataframes.py
+++ b/pipeline_dp/dataframes.py
@@ -66,7 +66,8 @@ class SparkConverter(DataFrameConvertor):
     def __init__(self, spark: pyspark.sql.SparkSession):
         self._spark = spark
 
-    def dataframe_to_collection(self, df, columns: Columns) -> pyspark.RDD:
+    def dataframe_to_collection(self, df: SparkDataFrame,
+                                columns: Columns) -> pyspark.RDD:
         columns_to_keep = [columns.privacy_key, columns.partition_key]
         if columns.value is not None:
             columns_to_keep.append(columns.value)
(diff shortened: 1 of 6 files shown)


pipeline_dp/dataframes.py
```diff
@@ -34,16 +48,53 @@ class ContributionBounds:
...
+    if isinstance(df, SparkDataFrame):
+        return pipeline_dp.SparkRDDBackend(df.sparkSession.sparkContext)
+    raise NotImplementedError(
+        f"Dataframes of type {type(df)} not yet supported")
+
+
+def _create_dataframe_converter(df: SparkDataFrame) -> DataFrameConvertor:
+    """Creates a DataConvert based on type of DataFrame."""
+    if isinstance(df, SparkDataFrame):
+        return SparkConverter(df.sparkSession)
```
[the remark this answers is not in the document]
RamSaw: ah, this is internal API, I guess then it doesn't make much difference :)

tests/dataframes_test.py
```diff
@@ -135,15 +138,112 @@ def test_count_and_sum_query(self):
...
+            max_partitions_contributed=5,
+            max_contributions_per_partition=2,
+            min_value=min_value,
+            max_value=max_value)
+        query = dataframes.Query(df, columns, metrics_dict, bounds, public_keys)
+
+        # Act
+        query.run_query(dataframes.Budget(1, 1e-10))
+
+        # Assert
```
RamSaw, line 195: should we assert on the output of `run_query` as well?

tests/dataframes_test.py
```diff
@@ -135,15 +138,112 @@ def test_count_and_sum_query(self):
...
+        # Arrange
+        spark = self._get_spark_session()
+        df = spark.createDataFrame(get_pandas_df())
+        columns = dataframes.Columns("privacy_key", "group_key", "value")
+        metrics = {
+            pipeline_dp.Metrics.COUNT: "count_column",
+            pipeline_dp.Metrics.SUM: "sum_column"
+        }
+        min_value = max_value = None
+        if pipeline_dp.Metrics.SUM in metrics:
```
RamSaw, line 217: this is always true

tests/dataframes_test.py
```diff
@@ -135,15 +138,112 @@ def test_count_and_sum_query(self):
...
+        result_df = query.run_query(budget)
+
+        # Assert
+        pandas_df = result_df.toPandas()
+        pandas_df = pandas_df.sort_values(by=['group_key']).reset_index(
+            drop=True)
+        self.assertLen(pandas_df, 2)
+        # check row[0] = "key0", 0+noise, 0+noise
+        row0 = pandas_df.loc[0]
+        self.assertEqual(row0.group_key, "key0")
```
RamSaw, line 238: why here we access it as a field and below as a dic (`row0["count_column"]`)? if both are possible should we use the same API everywhere?

dvadym (commented):
Thanks for review!

PR #499, opened 2023-10-24
Remove not needed validation on privacy_id_extractor

dvadym:
When `contribution_bounds_already_enforced = True` then PipelineDP does no contribution bounding. As a privacy id is not needed. The current validation requires `privacy_id_extractor` to be None, but there is no problem if privacy id extractor is set up. But it's pretty confusing for users why `privacy_id_extractor` has to be `None`. Let's drop this validation for simplicity

### DIFF
--- a/pipeline_dp/dp_engine.py
+++ b/pipeline_dp/dp_engine.py
@@ -420,10 +420,6 @@ class DPEngine:
         if check_data_extractors:
             _check_data_extractors(data_extractors)
         if params.contribution_bounds_already_enforced:
-            if data_extractors.privacy_id_extractor:
-                raise ValueError(
-                    "privacy_id_extractor should be set iff "
-                    "contribution_bounds_already_enforced is False")
             if pipeline_dp.Metrics.PRIVACY_ID_COUNT in params.metrics:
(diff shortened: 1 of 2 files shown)


pipeline_dp/dp_engine.py
```diff
@@ -420,10 +420,6 @@ def _check_aggregate_params(self,
         if check_data_extractors:
             _check_data_extractors(data_extractors)
         if params.contribution_bounds_already_enforced:
-            if data_extractors.privacy_id_extractor:
```
miracvbasaran, line 423: Would it be possible to log a warning? Clients might have an incorrect configuration in this case.

miracvbasaran (approved):

pipeline_dp/dp_engine.py
dvadym: I'd say that having privacy_id_extractor is a correct configuration, even if no contribution bounding happens, e.g. in principle it would allow to compute PRIVACY_ID_COUNT per partition. So let's keep w/o warnings.

dvadym (commented):
Thanks for review!

PR #501, opened 2023-10-31
Implement groupby by multiple columns in Spark DataFrame API

dvadym:
DataFrame API allows to perform DP on DataFrames (now only Spark DataFrames are supported). In DataFrame API privacy_key, partition_key and values are specified by column names. Currently `partition_key` (aka group by key) can be specified only by 1 column. This PR implements possibility to have multiple columns as  `partition_key` .

### DIFF
--- a/pipeline_dp/dataframes.py
+++ b/pipeline_dp/dataframes.py
@@ -15,7 +15,7 @@
 import abc
 from collections import namedtuple
 from dataclasses import dataclass
-from typing import Any, Dict, Iterable, Optional
+from typing import Any, Callable, Dict, Iterable, Optional, Sequence, Union
 
 import pipeline_dp
 import pyspark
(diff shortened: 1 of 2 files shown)


pipeline_dp/dataframes.py
```diff
@@ -229,24 +250,27 @@ def __init__(self, df, privacy_unit_column: str):
...
-                max_contributions_per_group: int,
-                public_keys: Optional[Iterable[Any]] = None) -> 'QueryBuilder':
+    def groupby(
+            self,
+            by: Union[str, Sequence[str]],
+            *,
+            max_groups_contributed: int,
+            max_contributions_per_group: int,
+            public_keys: Optional[Iterable[Any]] = None,
+            column: Optional[str] = None  # deprecated
```
miracvbasaran, line 266: When will we remove this deprecated parameter?

miracvbasaran (approved):

pipeline_dp/dataframes.py
dvadym: I'd say in a ~month (to be sure that everybody who uses this parameter migrate)

dvadym (commented):
Thanks for review!

PR #504, opened 2023-11-14
Fix noise_std computation in utility analysis

dvadym:
Before this PR, the query noise std deviation was computed with assumption that the metric is COUNT/PRICACY_ID_COUNT. This PR changes that, along the way it switches to the usage of the new API of adding noise (AdditiveMechanism).


miracvbasaran (approved):

dvadym (commented):
Thanks for review!

PR #505, opened 2023-11-16
Implement mean in DataFrame API

dvadym:
This PR intorduces `QueryBuilder.mean`, to allow DP mean computation. Along the way small refactoring done: allmost all validations are performed in `QueryBuilderbuild_query` instead of individual aggregation function (`count`, `sum` etc). That's more convenient, since the conditions on the correct state can be complicated, provided that in future the will be more aggregations and it will be allowed to compute aggregates of multiple columns.

### DIFF
--- a/pipeline_dp/dataframes.py
+++ b/pipeline_dp/dataframes.py
@@ -15,7 +15,7 @@
 import abc
 from collections import namedtuple
 from dataclasses import dataclass
-from typing import Any, Callable, Dict, Iterable, Optional, Sequence, Union
+from typing import Any, Dict, Iterable, Optional, Sequence, Tuple, Union
 
 import pipeline_dp
 import pyspark
@@ -244,6 +244,8 @@ class QueryBuilder:
         .groupby("day", max_groups_contributed=3, max_contributions_per_group=1)
         .count()
         .sum("money_spent", min_value=0, max_value=100)
+        .mean("money_spent") # no need to specify min_value, max_value, they
+        # were already given in sum
         .build_query()
     """
 
@@ -260,11 +262,10 @@ class QueryBuilder:
         self._df = df
         self._privacy_unit_column = privacy_unit_column
         self._by = None
-        self._value_column = None
-        self._metrics = {}  # map from pipeline_dp.Metric -> output column name
-        self._contribution_bounds = ContributionBounds()
         self._public_keys = None
         self._aggregations_specs = []
+        self._max_partitions_contributed: int = None
+        self._max_contributions_per_partition: int = None
 
     def groupby(
             self,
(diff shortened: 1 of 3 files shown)


tests/dataframes_test.py
```diff
@@ -235,7 +321,8 @@ def test_run_query_e2e_run(self):
         columns = dataframes.Columns("privacy_key", "group_key", "value")
         metrics = {
             pipeline_dp.Metrics.COUNT: "count_column",
-            pipeline_dp.Metrics.SUM: None  # it returns default name "sum"
+            pipeline_dp.Metrics.SUM: None,  # it returns default name "sum"
+            pipeline_dp.Metrics.MEAN: None  # it returns default name "sum"
```
RamSaw, line 325: nit: mean

btw, I had exactly this question what happens if it is not provided (None), should we add it somewhere in documentation in the source class?

tests/dataframes_test.py
```diff
@@ -264,6 +351,7 @@ def test_run_query_e2e_run(self):
         self.assertEqual(row1["group_key"], "key1")
         self.assertAlmostEqual(row1["count_column"], 2, delta=1e-3)
         self.assertAlmostEqual(row1["sum"], 3, delta=1e-3)
+        self.assertAlmostEqual(row1["mean"], 1.5, delta=1e-3)
 
     def test_run_query_multiple_partition_keys_e2e_run(self):
```
RamSaw, line 356: should it be updated with mean as well?

tests/dataframes_test.py
```diff
@@ -199,7 +285,7 @@ def test_run_query(self, mock_aggregate, metrics, public_keys):
         columns = dataframes.Columns("privacy_key", "group_key", value_column)
         metrics_dict = dict([(m, m.name) for m in metrics])
         min_value = max_value = None
-        if pipeline_dp.Metrics.SUM in metrics:
+        if pipeline_dp.Metrics.SUM in metrics or pipeline_dp.Metrics.MEAN in metrics:
```
RamSaw, line 288: mhh, I personally don't very like when we start writing a lot of "if" conditions specific to parameters of a test, it makes the test complicated, I think in many cases copy pasting is better, improves test readability, but up to you. We can say that it is more or less "integration" ("big") test and it is ok for it to have complicated arrange and assert sections.

RamSaw (approved):
Thanks! I don't have any big substantial comments therefore approve :)

pipeline_dp/dataframes.py
```diff
@@ -309,50 +320,128 @@ def count(self, name: str = None) -> 'QueryBuilder':
...
+            raise ValueError(
+                "Global aggregations are not supported. Use groupby")
+        self._aggregations_specs.append(aggregation_spec)
+        return self
+
+    def _get_value_caps(self) -> Tuple[Optional[float], Optional[float]]:
+        metrics = set([spec.metric for spec in self._aggregations_specs])
+        # COUNT, PPRIVACY_ID_COUNT do not require caps.
+        metrics_which_need_caps = metrics.difference(
+            [pipeline_dp.Metrics.COUNT, pipeline_dp.Metrics.PRIVACY_ID_COUNT])
```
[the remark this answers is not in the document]
dvadym: It's a good point that we can put the logic checking if needed caps in some other place. Having them as a part of `Metric` object is not nice, because it's actually implementation dependendant (for example in future we can implement algorithm which finds caps automatically).
Another option is to put it in some helper function closer to `Metric`. Thoug, I'm not sure. So let's leave as is for now.

dvadym (commented):
Thanks for review!

PR #507, opened 2023-11-20
Implement 1 BudgetAccount per one configuration in utility analysis

dvadym:
Having 1 budget allows having different `noise_kind` in different parameters in one run of utility analysis. Which is especially important since DPAlgorithmSelector can choose different `noise_kind`


miracvbasaran (approved):

dvadym:
Thanks for review!

PR #508, opened 2023-11-21
Deriving Spark DataFrame schema on converting from RDD to DataFrame

dvadym:
In case if the output is not empty the schema can be deduced automatically, but if the output is empty (e.g. because all partitions are dropped by partition selection) there was an example "Empty RDD". Setting the schema fixes that.

### DIFF
--- a/tests/dataframes_test.py
+++ b/tests/dataframes_test.py
@@ -420,7 +420,7 @@ class QueryTest(parameterized.TestCase):
                                  public_partitions=None)
 
         # Act
-        budget = dataframes.Budget(1, 1e-10)  # large budget to get small noise
+        budget = dataframes.Budget(1, 1e-10)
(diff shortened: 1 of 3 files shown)


tests/dataframes_test.py
```diff
@@ -402,6 +402,32 @@ def test_run_query_multiple_partition_keys_e2e_run(self):
...
+            min_value=-5,
+            max_value=5)
+        query = dataframes.Query(df,
+                                 columns,
+                                 metrics,
+                                 bounds,
+                                 public_partitions=None)
+
+        # Act
+        budget = dataframes.Budget(1, 1e-10)  # large budget to get small noise
```
sushkoy, line 423: Copy paste error?

sushkoy (approved):

PR #510, opened 2023-11-27
Fix for Histogram lower computations

dvadym:


monsieurmuffin (approved):
nit: it might be helpful to mention in the description what exactly has been fixed.

PR #512, opened 2023-12-01
Sum per partition histogram

dvadym:
Computing histogram for sum per partition (i.e. of non-dp sum per partition). Its computation similar to `linf_sum_contributions_histogram`: i.e. the range between min/max split on 10^4 bins and count/sum/max for each bin is computed.

### DIFF
--- a/tests/dataset_histograms/computing_histograms_test.py
+++ b/tests/dataset_histograms/computing_histograms_test.py
@@ -313,12 +313,13 @@ class ComputingHistogramsTest(parameterized.TestCase):
 
     @parameterized.product(
         (
-            dict(testcase_name='empty', input=[], expected=[]),
+            dict(testcase_name='empty', input=lambda: [], expected=lambda: []),
             dict(
                 testcase_name='small_histogram',
-                input=[((1, 1), 0.5), ((1, 2), 1.5), ((2, 1), -2.5),
-                       ((1, 1), 0.5)],  # ((privacy_id, partition), value)
-                expected=[
+                input=lambda: [((1, 1), 0.5), ((1, 2), 1.5), (
+                    (2, 1), -2.5), (
+                        (1, 1), 0.5)],  # ((privacy_id, partition), value)
+                expected=lambda: [
                     # step is (1.5 - (-2.5)) / 1e4 = 0.0004,
                     # ((2, 1), -2.5)
                     hist.FrequencyBin(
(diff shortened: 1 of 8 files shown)


tests/dataset_histograms/computing_histograms_test.py
```diff
@@ -563,28 +570,164 @@ def test_compute_partitions_privacy_id_count_histogram(
                          histogram.name)
         self.assertListEqual(expected, histogram.bins)
 
+    @parameterized.product(
+        (
+            dict(testcase_name='empty', input=lambda: [], expected=lambda: []),
```
RamSaw, line 575: nit: let's add histogram, not very clear what is empty

tests/dataset_histograms/computing_histograms_test.py
```diff
@@ -563,28 +570,164 @@ def test_compute_partitions_privacy_id_count_histogram(
...
+            dict(
+                testcase_name='small_histogram',
+                input=lambda: [((1, 1), 0.5), ((1, 2), 1.5), (
+                    (2, 1), -2.5), (
+                        (1, 1), 0.5)],  # ((privacy_id, partition), value)
+                expected=lambda: [
+                    hist.FrequencyBin(
+                        lower=-1.5, upper=-1.4997, count=1, sum=-1.5, max=-1.5),
```
RamSaw, line 583: nit: can be add a comment what the step in bucketing is to be able to verify that lowers and uppers are correct

tests/dataset_histograms/computing_histograms_test.py
```diff
@@ -563,28 +570,164 @@ def test_compute_partitions_privacy_id_count_histogram(
...
+                    hist.FrequencyBin(lower=1.4996999999999998,
+                                      upper=1.5,
+                                      count=1,
+                                      sum=1.5,
+                                      max=1.5)
+                ]),
+            dict(
+                testcase_name='Different privacy ids, 1 equal contribution',
```
RamSaw, line 591: nit: and different pks

tests/dataset_histograms/computing_histograms_test.py
RamSaw: ah, I see, the step is below, maybe move it higher, but idk, current code is ok as well

tests/dataset_histograms/computing_histograms_test.py
```diff
@@ -563,28 +570,164 @@ def test_compute_partitions_privacy_id_count_histogram(
...
+                # step = (1 - (-1)) / 1e4 = 0.0002,
+                # therefore last lower is 1 - 0.0002 = 0.9998.
+                expected=lambda: [
+                    hist.FrequencyBin(lower=-1.0,
+                                      upper=-0.9998,
+                                      count=10,
+                                      sum=-10.0,
+                                      max=-1.0),
+                    hist.FrequencyBin(lower=0.0,
+                                      upper=0.00019999999999997797,
```
RamSaw, line 676: hm, so this value we have here due to imprecise math, should then try to assert with an interval around 0.0002? But I guess then assertion will be more complicated because we won't be able to assert just like assert(actual == expected).

RamSaw (approved):
LGTM, only nits :)

PR #514, opened 2024-05-16
Add pre_thresholding to PrivatePCollection API

dvadym:


examples/movie_view_ratings/run_on_beam.py
```diff
@@ -32,6 +32,8 @@
 FLAGS = flags.FLAGS
 flags.DEFINE_string('input_file', None, 'The file with the movie view data')
 flags.DEFINE_string('output_file', None, 'Output file')
+flags.DEFINE_integer('pre_threshold', None,
+                     'Pre threshold which is used in the DP aggregation')
```
monsieurmuffin, line 36: nit: it might be worth mentioning to which aggregations pre-threshold is currently applied.

examples/movie_view_ratings/run_on_beam.py
```diff
@@ -60,21 +62,25 @@ def main(unused_argv):
...
+            # .. with minimal rating of "1"
+            min_value=1,
+            # .. and maximum rating of "5"
+            max_value=5,
+            # The aggregation key: we're grouping data by movies
+            partition_extractor=lambda mv: mv.movie_id,
+            # The value we're aggregating: we're summing up ratings
+            value_extractor=lambda mv: mv.rating)
+        if FLAGS.pre_threshold:
+            params.pre_threshold = FLAGS.pre_threshold
```
monsieurmuffin, line 80: I would have a preference for defining it at the same level as the other parameters rather than as a flag, because it's not very clear why we specify parameters in different places.

monsieurmuffin (approved):

PR #535, opened 2025-03-24
Mark numpy version as 1.23.5 to fix import issue on example Codelab

alexandrefreire:
## Description
Specify numpy version as 1.23.5 to fix import issue

If I don't specify this version, when trying to run in Google Codelab I get the following error:

ValueError: numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject

## Affected Dependencies
numpy

## How has this been tested?
- Running the codelab
- instructions to reproduce: trying to run the Codelab without the change

## Checklist
- [ X ] I have followed the [Contribution Guidelines](https://github.com/OpenMined/.github/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://github.com/OpenMined/.github/blob/master/CODE_OF_CONDUCT.md)
- [ X ] I have commented my code following the [OpenMined Styleguide](https://github.com/OpenMined/.github/blob/master/STYLEGUIDE.md)
- [ X ] I have labeled this PR with the relevant [Type labels](https://github.com/OpenMined/.github/labels?q=Type%3A)
- [ n/a ] My changes are covered by tests

### DIFF
--- a/examples/codelab/codelab_PipelineDP.ipynb
+++ b/examples/codelab/codelab_PipelineDP.ipynb
@@ -35,7 +35,7 @@
       "outputs": [],
       "source": [
         "!pip install numpy==1.23.5\n",
-        "!pip install pipeline-dp apache-beam=2.63\n",
+        "!pip install pipeline-dp apache-beam==2.63\n",
         "from IPython.display import clear_output\n",
         "\n",
         "clear_output()"
--- a/examples/codelab/codelab_PipelineDP.ipynb
+++ b/examples/codelab/codelab_PipelineDP.ipynb
@@ -35,7 +35,7 @@
       "outputs": [],
       "source": [
         "!pip install numpy==1.23.5\n",
-        "!pip install pipeline-dp apache-beam\n",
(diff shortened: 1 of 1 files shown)


examples/codelab/codelab_PipelineDP.ipynb
```diff
@@ -34,6 +34,7 @@
       },
       "outputs": [],
       "source": [
+        "!pip install numpy==1.23.5\n",
         "!pip install pipeline-dp apache-beam\n",
```
dvadym, line 38: It seems that the problem is that apache-beam expects np < 2.0, but at some point, apache-beam might require np == 2.0,  let's also fix the version of apache-beam, could you please add

`apache-beam==2.63`

dvadym (changes requested):
Thanks for the fix! I left a small comments.

examples/codelab/codelab_PipelineDP.ipynb
alexandrefreire: yes of course, thanks

alexandrefreire:
also fixed apache-beam version

PR #537, opened 2025-03-27
Implement ReiterableLazyIterable and use it in to_multi_transformable_collection

dvadym:
It's better to have to_multi_transformable_collection to be lazy, since all API is assumed to be lazy. This PR implements that.

### DIFF
--- a/pipeline_dp/pipeline_backend.py
+++ b/pipeline_dp/pipeline_backend.py
@@ -17,8 +17,8 @@ import functools
 import multiprocessing as mp
 import random
 import numpy as np
-from collections.abc import Iterable
-from typing import Callable
+from collections.abc import Iterable, Iterator
+from typing import Callable, List
 
 import abc
 import pipeline_dp.combiners as dp_combiners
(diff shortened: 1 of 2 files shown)


pipeline_dp/pipeline_backend.py
```diff
@@ -474,11 +474,39 @@ def to_list(self, col, stage_name: str):
...
+    """
+
+    def __init__(self, iterable: Iterable):
+        """Initializes the ReiterableLazyIterable.
+
+        Args:
+            iterable: Iterable to make reiterable
+        """
+        self._iterable = iterable
+        self._cache: List = None
```
RamSaw, line 491: why not to initialize it to `[]` immediately?

tests/pipeline_backend_test.py
```diff
@@ -588,7 +593,7 @@ def assert_laziness(operator, *args):
         assert_laziness(self.backend.sum_per_key)
         assert_laziness(self.backend.flat_map, str)
         assert_laziness(self.backend.sample_fixed_per_key, int)
-        assert_laziness(self.backend.filter_by_key, list)
```
RamSaw, line 591: why so?

pipeline_dp/pipeline_backend.py
dvadym: self._cache = [] wouldn't work, because this list will shared between instances

pipeline_dp/pipeline_backend.py
```diff
@@ -474,11 +474,39 @@ def to_list(self, col, stage_name: str):
...
+            self._first_run_complete = True
+        else:
+            yield from self._cache
+
+
 class LocalBackend(PipelineBackend):
     """Local Pipeline adapter."""
 
     def to_multi_transformable_collection(self, col):
-        return list(col)
```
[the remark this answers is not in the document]
dvadym: yes, for laziness. Comment added

tests/pipeline_backend_test.py
dvadym: we need somthing that it's converted to set.

pipeline_dp/pipeline_functions.py
```diff
@@ -146,10 +146,6 @@ def filter_by_key_with_sharding(backend: pipeline_backend.PipelineBackend, col,
             lambda p: tuple((p, i) for i in range(sharding_factor)),
             f"Shard partitions into {sharding_factor} keys",
         )
-        # to_multi_transformable_collection is no-op for not LocalMode. For
-        # local mode it is transform iterable to list, which is neded because
-        # filter_by_key requires list.
-        keys_to_keep = backend.to_multi_transformable_collection(keys_to_keep)
```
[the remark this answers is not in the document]
dvadym: we moved this logic inside LocalBackend.filter_by_key

dvadym (commented):
Thanks for review!

PR #539, opened 2025-03-28
Introduce Exponential mechanism to Budget Accounting

dvadym:
Atm dp_accounting doesn't support yet Exponential mechanism natively, so for PLD accounting PLD from eps-DP mechanism (which is actually eps-randomzed response) is used. eps-Randomized response is dominating any eps-DP so it's safe from privacy perspective, but it results in slightly higher noise.

### DIFF
--- a/tests/budget_accounting_test.py
+++ b/tests/budget_accounting_test.py
@@ -110,21 +110,25 @@ class NaiveBudgetAccountantTest(parameterized.TestCase):
         self.assertEqual(budget1.eps, 1.0 / (1.0 + 0.5))
         self.assertEqual(budget2.eps, 0.5 / (1.0 + 0.5))
 
-    def test_count(self):
+    def test_weights_count(self):
         budget_accountant = NaiveBudgetAccountant(total_epsilon=1,
                                                   total_delta=1e-6)
         budget1 = budget_accountant.request_budget(
             mechanism_type=MechanismType.LAPLACE, weight=4)
         budget2 = budget_accountant.request_budget(
             mechanism_type=MechanismType.GAUSSIAN, weight=3, count=2)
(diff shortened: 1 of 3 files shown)


tests/budget_accounting_test.py
```diff
@@ -110,22 +110,26 @@ def test_budget_scopes_no_parentscope(self):
...
         budget2 = budget_accountant.request_budget(
             mechanism_type=MechanismType.GAUSSIAN, weight=3, count=2)
+        budget3 = budget_accountant.request_budget(
+            mechanism_type=MechanismType.EXPONENTIAL, weight=2, count=5)
         budget_accountant.compute_budgets()
 
-        self.assertEqual(budget1.eps, 0.4)
-        self.assertEqual(budget1.delta,
-                         0)  # Delta should be 0 if mechanism is Laplace.
+        self.assertEqual(budget1.eps, 0.2)
```
RamSaw, line 124: maybe write a formula, not trivial

tests/budget_accounting_test.py
```diff
@@ -467,7 +471,26 @@ class ComputeBudgetTestCase:
                         weight=8,
                         sensitivity=3),
                 ],
-                expected_pipeline_noise_std=40.02)
```
RamSaw, line 470: should we write/point to the formulas in the class definition?

RamSaw (approved):
How does it work? Shouldn't change any mechanism specific code?

tests/budget_accounting_test.py
dvadym: Yeah, we should write documentation on PLD accountant. In this case, there is no simple formula anyway :)

dvadym (commented):
Thanks for review!

PR #541, opened 2025-04-04
Improve backend map and flat_map side input

dvadym:
This PR contains:

1. Adding `flat_map_with_side_int`
2. Changing assumption on side-input, previously the side input can be a collection with any number of elements, now, it must be a Singleton (I.e. collection with 1 element). As a result it simplifies the API, now `fn` takes an object, not a list (usually 1 element list). 
3. Docstring improvements
4. Some fixes


pipeline_dp/pipeline_backend.py
```diff
@@ -246,14 +288,23 @@ def map_with_side_inputs(self,
                              side_input_cols,
                              stage_name: str = None):
         side_inputs = [
-            beam.pvalue.AsList(side_input_col)
+            beam.pvalue.AsSingleton(side_input_col)
```
RamSaw, line 291: is it more efficient in terms of speed? It makes more sense to use `AsList` because it use more memory-efficient.

pipeline_dp/pipeline_backend.py
```diff
@@ -496,10 +552,70 @@ def __iter__(self) -> Iterator:
...
+        """
+        if self._singleton:
+            return self._singleton
+        it = iter(self._iterable)
+        self._singleton = next(it)
+        try:
+            next(it)
+        except StopIteration:
+            return self._singleton
+        raise ValueError("")
```
RamSaw, line 618: should we write something here?

pipeline_dp/pipeline_backend.py
```diff
@@ -496,10 +552,70 @@ def __iter__(self) -> Iterator:
...
+                    f"{len(iterable_or_list)} elements.")
+            self._singleton = iterable_or_list[0]
         else:
-            yield from self._cache
+            if not isinstance(iterable_or_list, Iterable):
+                raise TypeError(f"Input must be a list or an Iterable, but got "
+                                f"{type(iterable_or_list).__name__}.")
+            self._iterable = iterable_or_list
+
+    def singleton(self):
```
RamSaw, line 600: should we make it a property?

pipeline_dp/pipeline_backend.py
```diff
@@ -534,6 +665,7 @@ def group_by_key_generator():
             d = collections.defaultdict(list)
             for key, value in col:
                 d[key].append(value)
+            print("***", d)
```
RamSaw, line 668: should we remove it?

tests/pipeline_backend_test.py
```diff
@@ -417,6 +423,43 @@ def tearDownClass(cls):
         cls.sc.stop()
 
 
+class TestLazySingleton(unittest.TestCase):
```
RamSaw, line 426: should we maybe test it through public API? LazySingleton is implementaiton details

tests/pipeline_backend_test.py
```diff
@@ -417,6 +423,43 @@ def tearDownClass(cls):
...
+    def test_init_with_valid_iterable_generator(self):
+        data = (x for x in [10])  # lazy generator
+        ls = LazySingleton(data)
+        self.assertIsNone(ls._singleton)  # Should be lazy
+        self.assertIsInstance(ls._iterable, Iterable)
+        self.assertEqual(ls.singleton(), 10)
+        self.assertEqual(ls.singleton(), 10)
+
+    def test_init_with_empty_list_raises_value_error(self):
+        with self.assertRaisesRegex(ValueError, "exactly one element.*found 0"):
```
RamSaw, line 444: what `*` means here?

tests/pipeline_backend_test.py
```diff
@@ -417,6 +423,43 @@ def tearDownClass(cls):
         cls.sc.stop()
 
 
+class TestLazySingleton(unittest.TestCase):
+
+    def test_with_list(self):
+        ls = LazySingleton([10])
+        self.assertEqual(ls._singleton, 10)
```
RamSaw, line 430: these are implementations details, we should assert on them I think

tests/pipeline_backend_test.py
```diff
@@ -417,6 +423,43 @@ def tearDownClass(cls):
...
+
+    def test_init_with_multi_element_list_raises_value_error(self):
+        with self.assertRaisesRegex(ValueError, "exactly one element.*found 3"):
+            LazySingleton([1, 2, 3])
+
+    def test_init_with_non_iterable_raises_type_error(self):
+        with self.assertRaisesRegex(TypeError,
+                                    "must be a list or an Iterable.*got int"):
+            LazySingleton(123)
+        with self.assertRaisesRegex(
```
RamSaw, line 455: I think one such assert is enough

just want to avoid accumulating test maintenance burden

RamSaw (approved):
Let's also add a description for the PR and explain what is the improvement (key code changes) and why it is an improvement.

pipeline_dp/pipeline_backend.py
dvadym: I don't think there are any difference in efficient. This is simplification, in all our cases we have 1 element collection (and we always can get 1-element collection with `to_list`) and I found it's confusing for new people, that `fn` takes a list of 1-element

pipeline_dp/pipeline_backend.py
dvadym: I'd say it's better a method, because it has non-trivial logic

tests/pipeline_backend_test.py
dvadym: result experssion, any number of element

tests/pipeline_backend_test.py
dvadym: Yeah, it's an implementation detail, but pretty non-trivial, so I'd prefer to test it.

dvadym (commented):
Thanks for review!

PR #546, opened 2025-04-28
Add contribution_bounds_already_enforced to select_partition

dvadym:
This PR introduces a flat `contribution_bounds_already_enforced` to `DPEngine.select_partitions` (it's already available in `DPEngine.aggregate`). This flag can be used when the contribution bounding is performed before the call of  `select_partitions`.


pipeline_dp/aggregate_params.py
```diff
@@ -409,11 +409,24 @@ class SelectPartitionsParams:
             partition selection.
         partition_selection_strategy: which strategy to use for private
          partition selection.
+        pre_threshold: the minimum amount of privacy units which are required
+         for keeping a partition in private partition selection. Note that this
+         is in addition to a differentially private partition selection, so a
+         partition with pre_threshold privacy units isn't necessarily kept. It
```
RamSaw, line 415: with at least pre_threshold

tests/dp_engine_test.py
```diff
@@ -902,6 +902,58 @@ def test_select_partitions(self):
...
+
+        col = list(col)
+
+        # Assert
+        if max_partitions_contributed == 1:
+            self.assertEqual(["pk-many-contribs", "pk-medium-contribs"],
+                             sorted(col))
+        elif max_partitions_contributed == 100:
+            self.assertEqual(["pk-many-contribs"], col)
+        else:
```
RamSaw, line 954: looks like a code smell. Doesn't case with `max_partitions_contributed == 100` cover all lines and more than `max_partitions_contributed == 1`? Can we keep only this one?

RamSaw (approved):

dvadym (commented):
Thanks for review!

PR #547, opened 2025-04-28
Extend add_noise to support numpy arrays

dvadym:
`DPEngine.add_noise`  adds noise for a collection of `(partition_key, value)`. In contrast to `DPEngine.aggregate`, `add_noise` doesn't perform aggregation. 

This PR extends `add_noise` on supporting Numpy arrays

### DIFF
--- a/pipeline_dp/aggregate_params.py
+++ b/pipeline_dp/aggregate_params.py
@@ -116,6 +116,12 @@ class MechanismType(Enum):
         raise ValueError(f"MechanismType {self.value} can not be converted to "
                          f"PartitionSelectionStrategy")
 
+    def is_thresholding_mechanism(self):
+        return self.value in [
+            MechanismType.LAPLACE_THRESHOLDING.value,
+            MechanismType.GAUSSIAN_THRESHOLDING.value
+        ]
+
 
 def noise_to_thresholding(noise_kind: NoiseKind) -> MechanismType:
     if noise_kind == NoiseKind.LAPLACE:
--- a/pipeline_dp/aggregate_params.py
+++ b/pipeline_dp/aggregate_params.py
@@ -116,12 +116,6 @@ class MechanismType(Enum):
         raise ValueError(f"MechanismType {self.value} can not be converted to "
                          f"PartitionSelectionStrategy")
 
-    def is_thresholding_mechanism(self):
-        return self.value in [
-            MechanismType.LAPLACE_THRESHOLDING.value,
-            MechanismType.GAUSSIAN_THRESHOLDING.value
-        ]
-
 
 def noise_to_thresholding(noise_kind: NoiseKind) -> MechanismType:
     if noise_kind == NoiseKind.LAPLACE:
--- a/pipeline_dp/aggregate_params.py
+++ b/pipeline_dp/aggregate_params.py
@@ -656,18 +656,44 @@ class AddDPNoiseParams:
     sensitivity by contribution bounding and relies on the caller to ensure the
     provided data satisfies the provided bound.
 
-   Attributes:
+    The noise in DP is scaled with sensitivity, where sensitivity for query q is
+     sensitivity = \max_{D, D' neighbouring} ||q(D)-q(D')||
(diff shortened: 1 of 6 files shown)


pipeline_dp/aggregate_params.py
```diff
@@ -650,18 +650,44 @@ class AddDPNoiseParams:
...
+    (partition_key, value). Where partition_key is any identifier, value is
+    a scalar. q values can be represented as a vector indexed by partition_key.
+    One of the convenient ways to define l1/l2 sensitivity is through l0/linf
+    sensitivities.
+    Namely
+
+    l1_sensitivity = l0_sensitivity*linf_sensitivity
+    l2_sensitivity = sqrt(l0_sensitivity)*linf_sensitivity
+
+    For applying the Laplace mechanism or l1_sensitivity or
```
RamSaw, line 669: remove first "or"

or use "either"

pipeline_dp/aggregate_params.py
```diff
@@ -650,18 +650,44 @@ class AddDPNoiseParams:
...
+    l0_sensitivity, linf_sensitivity must be specified.
```
RamSaw, line 670: optional: maybe put l0 linf into brackets () to highlight that both of them have to specified

pipeline_dp/aggregate_params.py
```diff
@@ -650,18 +650,44 @@ class AddDPNoiseParams:
...
+    
+    For applying the Gaussian mechanism or l2_sensitivity or
```
RamSaw, line 672: ditto

pipeline_dp/dp_computations.py
```diff
@@ -505,7 +513,13 @@ def create_from_std_deviation(cls, normalized_stddev: float,
...
 
-    def add_noise(self, value: Union[int, float]) -> float:
+    def add_noise(
+            self, value: Union[int, float,
+                               np.ndarray]) -> Union[float, np.ndarray]:
+        if isinstance(value, np.ndarray):
+            add_noise_vectorized = np.vectorize(
+                lambda x: self._mechanism.add_noise(1.0 * x))
+            return add_noise_vectorized(value)
         return self._mechanism.add_noise(1.0 * value)
```
RamSaw, line 523: this is duplicate can code, does it make sense to define it in the parent abstract code?

also at line 523 nit: (same above for laplace) we can reuse lambda from numpy case and call it here

tests/dp_computations_test.py
```diff
@@ -544,6 +559,25 @@ def test_gaussian_mechanism_distribution(self, epsilon, delta,
         res = stats.ks_1samp(noised_values, expected_cdf)
         self.assertGreater(res.pvalue, 1e-4)
 
+    def test_gaussian_mechanism_distribution_nd_array(self):
```
RamSaw, line 562: I wonder whether we can parameterize (probaly not by mechanism but by type of the input)

tests/dp_computations_test.py
```diff
@@ -544,6 +559,25 @@ def test_gaussian_mechanism_distribution(self, epsilon, delta,
...
+        epsilon = 0.1
+        delta = 1e-5
+        l2_sensitivity = 0.55
+        value = 2000
```
RamSaw, line 566: why do we multiply by value? in Laplace we don't do this

pipeline_dp/aggregate_params.py
```diff
@@ -650,18 +650,44 @@ class AddDPNoiseParams:
...
        budget_weight: Relative weight of the privacy budget allocated to this
          aggregation.
    """
     noise_kind: NoiseKind
-    l0_sensitivity: int
-    linf_sensitivity: float
+    l0_sensitivity: Optional[int] = None
+    linf_sensitivity: Optional[float] = None
+    l1_sensitivity: Optional[float] = None
+    l2_sensitivity: Optional[float] = None
```
RamSaw, line 690: where do we use these new values in this PR? should be a separate PR with code where we use them?

better to split I think but if we keep then let's at least update PR description

RamSaw (approved):

PR #549, opened 2025-05-06
Support PLD for Laplace/Gaussian thresholding

dvadym:
This PR implements support of Laplace/Gaussian thresholding in `PLDBudgetAccountant`. It's implemented the following way.

PLD implementation for dp_accounting doesn't directly support thresholding, we need to track indpendently threshold delta (e.g. Gaussian Thresholding = (Gaussian Mechanism, threshold delta). `0.25*total_delta` is taken of  threshold_deltas of thresholding mechanisms (0.25 is rather arbitrary). (this implemented in budget_accountant.py)

Another complication is that the output of PLD computations for Thresholding is stddev of the noise and threshold_delta. PyDP supports only `(epsilon, delta)` for partition selection, so we need to choose  proper `(epsilon, delta)` (implementation in partition_selection.py)

### DIFF
--- a/pipeline_dp/budget_accounting.py
+++ b/pipeline_dp/budget_accounting.py
@@ -46,10 +46,11 @@ class MechanismSpec:
     (_eps, _delta) are parameters of (eps, delta)-differential privacy
     """
     mechanism_type: agg_params.MechanismType
-    _noise_standard_deviation: float = None
-    _eps: float = None
-    _delta: float = None
-    _count: int = 1
+    _noise_standard_deviation: Optional[float] = None
+    _eps: Optional[float] = None
+    _delta: Optional[float] = None
+    _count: Optional[int] = 1
+    _thresholding_delta: Optional[float] = None
 
     @property
     def noise_standard_deviation(self):
@@ -102,6 +103,13 @@ class MechanismSpec:
 
     def set_noise_standard_deviation(self, stddev: float):
         self._noise_standard_deviation = stddev
+
+    def set_thresholding_delta(self, delta):
+        self._thresholding_delta = delta
+
+    @property
+    def thresholding_delta(self) -> float:
+        return self._thresholding_delta
 
     def use_delta(self) -> bool:
         return self.mechanism_type != agg_params.MechanismType.LAPLACE
@@ -534,6 +542,8 @@ class PLDBudgetAccountant(BudgetAccountant):
             minimum_noise_std = self._find_minimum_noise_std()
 
         self.minimum_noise_std = minimum_noise_std
+        thresholding_delta_per_mechanism = self._get_thresholding_delta(
+        ) / self._count_thresholding_mechanisms()
         for mechanism in self._mechanisms:
             mechanism_noise_std = mechanism.sensitivity * minimum_noise_std / mechanism.weight
             mechanism.mechanism_spec._noise_standard_deviation = mechanism_noise_std
@@ -541,6 +551,10 @@ class PLDBudgetAccountant(BudgetAccountant):
                 epsilon_0 = math.sqrt(2) / mechanism_noise_std
                 delta_0 = epsilon_0 / self._total_epsilon * self._total_delta
                 mechanism.mechanism_spec.set_eps_delta(epsilon_0, delta_0)
+            elif mechanism.mechanism_spec.mechanism_type.is_thresholding_mechanism(
+            ):
+                mechanism.set_thresholding_delta(
+                    thresholding_delta_per_mechanism)
 
     def _find_minimum_noise_std(self) -> float:
         """Finds the minimum noise which satisfies the total budget.
@@ -552,13 +566,14 @@ class PLDBudgetAccountant(BudgetAccountant):
         Returns:
             The noise value adjusted for the given epsilon.
         """
+        delta = self._total_delta - self._get_thresholding_delta()
         threshold = 1e-4
         maximum_noise_std = self._calculate_max_noise_std()
         low, high = 0, maximum_noise_std
         while low + threshold < high:
             mid = (high - low) / 2 + low
             pld = self._compose_distributions(mid)
-            pld_epsilon = pld.get_epsilon_for_delta(self._total_delta)
+            pld_epsilon = pld.get_epsilon_for_delta(delta)
             if pld_epsilon <= self._total_epsilon:
                 high = mid
             elif pld_epsilon > self._total_epsilon:
--- a/pipeline_dp/budget_accounting.py
+++ b/pipeline_dp/budget_accounting.py
@@ -52,7 +52,7 @@ class MechanismSpec:
     _noise_standard_deviation: Optional[float] = None
     _eps: Optional[float] = None
     _delta: Optional[float] = None
-    _count: Optional[int] = 1
+    _count: int = 1
     _thresholding_delta: Optional[float] = None
 
     @property
--- a/pipeline_dp/budget_accounting.py
+++ b/pipeline_dp/budget_accounting.py
@@ -577,46 +577,33 @@ class PLDBudgetAccountant(BudgetAccountant):
         return max_noise_std
 
     def _compose_distributions(
-            self, noise_standard_deviation: float
-    ) -> 'pldlib.PrivacyLossDistribution':
+            self, noise_stddev: float) -> 'pldlib.PrivacyLossDistribution':
         """Uses the Privacy Loss Distribution library to compose distributions.
 
         Args:
-            noise_standard_deviation: The noise of the distributions to construct.
+            noise_stddev: The base noise standard deviation of the distributions
+              to construct.
 
         Returns:
-            A PrivacyLossDistribution object for the pipeline.
+            A PrivacyLossDistribution which corresponds to self._mechanisms and
+            the base noise 'noise_stddev'.
         """
         composed, pld = None, None
 
-        for mechanism_spec_internal in self._mechanisms:
-            if mechanism_spec_internal.mechanism_spec.mechanism_type == agg_params.MechanismType.LAPLACE:
-                # The Laplace distribution parameter = std/sqrt(2).
-                pld = pldlib.from_laplace_mechanism(
-                    mechanism_spec_internal.sensitivity *
-                    noise_standard_deviation / math.sqrt(2) /
-                    mechanism_spec_internal.weight,
-                    value_discretization_interval=self._pld_discretization)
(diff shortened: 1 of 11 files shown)


RamSaw:
Reading the description of the PR, it seems like we do a lot of workarounds because the libraries and code we use downstream has not very convenient APIs for our task. Should we also consider changing APIs of these libraries (e.g. PyDP)? We have access to them, could be a better long-term solution

dvadym:
There are 2 things:

1. Support of thresholding in dp_accounting. It's actually not clear whether it should be. It can be too specific for this library. It would require discussions.
2. Creating Partition Selection objects from noise stddev and thresholding delta. It's probably good things to have. The main reason for using this is using some advanced composition tools (PLD, RDP etc). But I think we should make a bigger design decission - how to make support of PLD/RDP in C++ building block library, not just partition selection objects. 

I agree that it would be nice to have, but it would require significant discussions and design works. Which is at least O(months). My proposal is to submit as is, and to create tracking bugs. WDYT?

pipeline_dp/budget_accounting.py
```diff
@@ -458,12 +473,12 @@ def __init__(self,
                               "NaiveBudgetAccountant instead of "
                               "PLDBudgetAccountant")
 
-        self.minimum_noise_std = None
+        self.base_noise_std = None
```
RamSaw, line 476: what's the difference between minimum and base? should we add a comment highlighting that it is base and not minimum?

pipeline_dp/dp_computations.py
```diff
@@ -116,6 +118,47 @@ def compute_sigma(eps: float, delta: float, l2_sensitivity: float) -> float:
...
+        raise ValueError(f"sigma must be > 0, but {sigma=}")
+    if delta < 0 or delta > 1:
+        raise ValueError(f"delta must be in [0, 1], but {delta=}")
+    # For a fixed sigma a function delta(epsilon) is decreasing. Solve the
+    # equation gaussian_delta(sigma, epsilon) = delta with binary search.
+    f = functools.partial(gaussian_delta, sigma)
+    if f(0) >= delta:
+        L = 0
+        R = 1
+        while f(R) > delta:
```
RamSaw, line 144: `>=` to preserve the invariant that f(L) >= delta and f(R) < R

pipeline_dp/dp_computations.py
```diff
@@ -116,6 +118,47 @@ def compute_sigma(eps: float, delta: float, l2_sensitivity: float) -> float:
...
+            R *= 2
+    else:
+        R = 0
+        L = -1
+        while f(L) < delta:
+            L *= 2
+
+    while R - L > 1e-10:
```
RamSaw, line 152: not sure whether it is useful but we can also consider using `scipy.optimize.brentq`, `scipy.optimize.bisect` or `scipy.optimize.root_scalar`

pipeline_dp/dp_engine.py
```diff
@@ -649,3 +657,20 @@ def _check_data_extractors(data_extractors: pipeline_dp.DataExtractors):
         raise ValueError("data_extractors must be set to a DataExtractors")
     if not isinstance(data_extractors, pipeline_dp.DataExtractors):
         raise TypeError("data_extractors must be set to a DataExtractors")
+
+
+def _create_partition_selector(strategy: pipeline_dp.PartitionSelectionStrategy,
+                               budget: budget_accounting.MechanismSpec,
+                               max_partitions: int, pre_threshold: int):
+    if budget.standard_deviation_is_set:
+        assert strategy.is_thresholding
```
RamSaw, line 666: can it happen and user of a library should address it? Then we should throw a more user-friendly message

pipeline_dp/dp_engine.py
```diff
@@ -649,3 +657,20 @@ def _check_data_extractors(data_extractors: pipeline_dp.DataExtractors):
...
+        if strategy == pipeline_dp.PartitionSelectionStrategy.LAPLACE_THRESHOLDING:
+            return partition_selection.create_laplace_thresholding(
+                budget.noise_standard_deviation, budget.thresholding_delta,
+                max_partitions, pre_threshold)
+        return partition_selection.create_gaussian_thresholding(
```
RamSaw, line 671: can `strategy` be at this point not GAUSSIAN_THRESHOLDING? is it WAI? should we have explicit and complete `if elif else` switch?

pipeline_dp/dp_engine.py
```diff
@@ -649,3 +657,20 @@ def _check_data_extractors(data_extractors: pipeline_dp.DataExtractors):
...
+            budget.noise_standard_deviation, budget.thresholding_delta,
+            max_partitions, pre_threshold)
+
+    return partition_selection.create_partition_selection_strategy(
```
RamSaw, line 675: should we rename this method? its name is fitting to thresholding mechanisms as well

pipeline_dp/partition_selection.py
```diff
@@ -42,3 +43,83 @@ def create_partition_selection_strategy(
...
+    """Creates PyDP Gaussian partition selection object.
+
+    Gaussian thresholding with parameters sigma and thresholding_delta is the
+    partition selection mechanism:
+      if num_privacy_units + N(0, sigma^2) >= threshold:
+        return "release"
+      else:
+        return "not-release"
+    Where `threshold` is chosen, such that the probability of releasing of a
+    partition with 1 privacy unit is less equal to thresholding_delta, i.e.
```
RamSaw, line 62: nit: less or equal

tests/budget_accounting_test.py
```diff
@@ -529,6 +522,51 @@ class ComputeBudgetTestCase:
...
+                        delta=1e-2,
+                        msg=f"failed test {case.name} expected mechanism noise "
+                        f"{expected_mechanism_noise_std} "
+                        f"got {actual_mechanism.noise_standard_deviation}")
+
+    def test_compute_budgets_gaussian_thresholding(self):
+        accountant = PLDBudgetAccountant(total_epsilon=1.0, total_delta=1e-8)
+        thresholding_budget = accountant.request_budget(
+            MechanismType.GAUSSIAN_THRESHOLDING)
+        guasisan_budget = accountant.request_budget(MechanismType.GAUSSIAN)
```
RamSaw, line 538: nit: gaussian

tests/budget_accounting_test.py
```diff
@@ -529,6 +522,51 @@ class ComputeBudgetTestCase:
...
+        accountant.compute_budgets()
+
+        # Gaussian mechanism thresholding and Gaussian mechanism have to have the
+        # same sigma, since Gaussian mechanism thresholding is modeled with
+        # Gaussian.
+        self.assertAlmostEqual(thresholding_budget.noise_standard_deviation,
```
RamSaw, line 544: why not to assert that `thresholding_budget.noise_standard_deviation == guasisan_budget.noise_standard_deviation`?

tests/budget_accounting_test.py
```diff
@@ -529,6 +522,51 @@ class ComputeBudgetTestCase:
...
+                               7.284667,
+                               delta=1e-5)
+        self.assertAlmostEqual(guasisan_budget.noise_standard_deviation,
+                               7.284667,
+                               delta=1e-5)
+        self.assertEqual(thresholding_budget.thresholding_delta, 1e-8 / 4)
```
RamSaw, line 550: we should explain how we get this value? same above if we keep `7.284667` equality check

tests/budget_accounting_test.py
```diff
@@ -529,6 +522,51 @@ class ComputeBudgetTestCase:
...
+
+    def test_compute_budgets_laplace_thresholding(self):
+        accountant = PLDBudgetAccountant(total_epsilon=1.0, total_delta=1e-8)
+        thresholding_budget = accountant.request_budget(
+            MechanismType.LAPLACE_THRESHOLDING)
+        guasisan_budget = accountant.request_budget(MechanismType.LAPLACE,
```
RamSaw, line 556: nit: gaussian

tests/dp_engine_test.py
```diff
@@ -1189,7 +1189,7 @@ def run_e2e_private_partition_selection_large_budget(self, col, backend):
         sys.version_info.major == 3 and sys.version_info.minor <= 8,
         "dp_accounting library only support python >=3.9")
     @parameterized.parameters(False, True)
-    def test_run_e2e_count_public_partition_local(self, pld_accounting):
+    def test_run_e2e_count_public_partition_local(self, pld_accounting=True):
```
RamSaw, line 1192: why do we set it default to True if it is parameterized test?

tests/partition_selection_test.py
```diff
@@ -67,6 +67,27 @@ def test_truncated_pre_thresholding(self, mock_method):
         mock_method.assert_called_once_with("gaussian", eps, delta,
                                             max_partitions, pre_threshold)
 
+    @patch("pydp.algorithms.partition_selection.create_partition_strategy")
+    def test_truncated_laplace_thresholding_with_sigma(self, mock_method):
+        sigma, delta, max_partitions_contributed, pre_threshold = 1.5, 1e-5, 2, 3
+        partition_selection.create_laplace_thresholding(
+            sigma, delta, max_partitions_contributed, pre_threshold)
+        mock_method.assert_called_once_with("laplace",
+                                            math.sqrt(2) / sigma, delta,
```
RamSaw, line 76: I would use named arguments here, e.g. `epsilon=math.sqrt(2) / sigma`.

tests/partition_selection_test.py
```diff
@@ -67,6 +67,27 @@ def test_truncated_pre_thresholding(self, mock_method):
...
+                                            max_partitions_contributed,
+                                            pre_threshold)
+
+    @patch("pydp.algorithms.partition_selection.create_partition_strategy")
+    def test_truncated_gaussian_thresholding_with_sigma(self, mock_method):
+        sigma, delta, max_partitions_contributed, pre_threshold = 1.5, 1e-5, 5, 10
+        partition_selection.create_gaussian_thresholding(
+            sigma, delta, max_partitions_contributed, pre_threshold)
+        expected_epsilon = 2.7533813795016613
```
RamSaw, line 85: is it useful to hardcode such a precise number? but I guess we have no choice if we assert it with mockito...maybe not assert on that? or add a comment how we calculated that.

also, it would be prefereable to test properties of the object returned by `create_gaussian_thresholding` but I guess it is not possible, or?

RamSaw (approved):
Yes, agree that given the size of the added code it is ok to have such an intermediate solution

dvadym (commented):
Thanks for review!

PR #553, opened 2025-05-13
Introduce filter_with_side_input

dvadym:
Side input is a singleton collection which is passed as an argument to a mapper function. Previously map and flat_map with side input were already added.

### DIFF
--- a/tests/pipeline_backend_test.py
+++ b/tests/pipeline_backend_test.py
@@ -81,14 +81,16 @@ class BeamBackendTest(parameterized.TestCase):
     def test_filter_with_side_inputs(self):
         with test_pipeline.TestPipeline() as p:
             data = p | "Create data" >> beam.Create([1, 2, 3, 4, 5, 6])
-            side_input = p | "Create side_input" >> beam.Create([[2, 4]])
+            side_input1 = p | "Create side_input1" >> beam.Create([[2, 4]])
+            side_input2 = p | "Create side_input2" >> beam.Create([[5]])
 
-            def filter_fn(x, side_input):
-                return x in side_input
+            def filter_fn(x, side_input1, side_input2):
+                return x in side_input1 or x in side_input2
 
             result = self.backend.filter_with_side_inputs(
-                data, filter_fn, [side_input], "Filter")
-            beam_util.assert_that(result, beam_util.equal_to([2, 4]))
+                data, filter_fn, [side_input1, side_input2], "Filter")
+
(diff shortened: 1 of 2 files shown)


tests/pipeline_backend_test.py
```diff
@@ -78,6 +78,18 @@ def add_fn(elems, s1, s2):
...
+        with test_pipeline.TestPipeline() as p:
+            data = p | "Create data" >> beam.Create([1, 2, 3, 4, 5, 6])
+            side_input = p | "Create side_input" >> beam.Create([[2, 4]])
+
+            def filter_fn(x, side_input):
+                return x in side_input
+
+            result = self.backend.filter_with_side_inputs(
+                data, filter_fn, [side_input], "Filter")
+            beam_util.assert_that(result, beam_util.equal_to([2, 4]))
```
RamSaw, line 91: nit: add empty line betweem the,

tests/pipeline_backend_test.py
```diff
@@ -78,6 +78,18 @@ def add_fn(elems, s1, s2):
...
 
+    def test_filter_with_side_inputs(self):
+        with test_pipeline.TestPipeline() as p:
+            data = p | "Create data" >> beam.Create([1, 2, 3, 4, 5, 6])
+            side_input = p | "Create side_input" >> beam.Create([[2, 4]])
+
+            def filter_fn(x, side_input):
+                return x in side_input
+
+            result = self.backend.filter_with_side_inputs(
```
RamSaw, line 89: I would test with 2 side inputs if possible

tests/pipeline_backend_test.py
```diff
@@ -561,6 +573,17 @@ def test_filter(self):
...
 
+    def test_filter_with_side_inputs(self):
+        data = [1, 2, 3, 4, 5, 6]
+        side_input = [[2, 4]]
+
+        def filter_fn(x, side_input):
+            return x in side_input
+
+        result = list(
+            self.backend.filter_with_side_inputs(data, filter_fn, [side_input]))
```
RamSaw, line 584: `[side_input]` = `[[[2, 4]]]`, list of list of list. Is it what we want? Not list of list?

if yes, then maybe add some comment in the pipeline_backend.py because I find it confusing.

RamSaw (approved):

tests/pipeline_backend_test.py
dvadym: Good point. I've added docstrings

dvadym (commented):
Thanks for review!

PR #554, opened 2025-05-13
Add an option to add noise standard deviation to the ouput of add_dp_noise

dvadym:
It's similar to [PR](https://github.com/OpenMined/PipelineDP/pull/550), but for another function  `DPEngine.add_dp_noise`.


RamSaw (approved):

dvadym (commented):
Thanks for review!

PR #557, opened 2025-05-26
Make LocalBackend methods output reiterable

dvadym:
Old behaviour:

```
  backend = pipeline_dp.LocalBackend(output_reiterable=False)
  output = backend.map([1, 2, 3], lambda x: x, "Map")
  print(list(output)) # , [1, 2, 3])
  print(list(output)) # [], only one iteration
```

New (default) behavior
```
  backend = pipeline_dp.LocalBackend(output_reiterable=False)
  output = backend.map([1, 2, 3], lambda x: x, "Map")
  print(list(output)) # , [1, 2, 3])
  print(list(output)) # [1,2,3] any number of iterations
```

### DIFF
--- a/pipeline_dp/pipeline_backend.py
+++ b/pipeline_dp/pipeline_backend.py
@@ -663,7 +663,7 @@ class LocalBackend(PipelineBackend):
         return ReiterableLazyIterable(col)
 
     def map(self, col, fn, stage_name: typing.Optional[str] = None):
-        return map(fn, col)
+        return ReiterableLazyIterable(map(fn, col))
 
     def map_with_side_inputs(self,
                              col,
@@ -676,10 +676,10 @@ class LocalBackend(PipelineBackend):
             side_input_values = [s.singleton() for s in side_inputs_singletons]
             return fn(x, *side_input_values)
 
-        return map(map_fn, col)
+        return ReiterableLazyIterable(map(map_fn, col))
 
     def flat_map(self, col, fn, stage_name: str = None):
-        return (x for el in col for x in fn(el))
+        return ReiterableLazyIterable((x for el in col for x in fn(el)))
 
     def flat_map_with_side_inputs(self, col, fn, side_input_cols,
                                   stage_name: str):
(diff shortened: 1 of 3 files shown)


pipeline_dp/pipeline_backend.py
```diff
@@ -659,11 +659,28 @@ def singleton(self):
...
 
+    def __init__(self, output_reiterable: bool = True):
+        """
+        Args:
+          output_reiterable: if true, output of all methods are reiterable,
+           i.e. it's possible to run iteration any number of time. It is
+           guaranteed that the iteration output is the same on each iteration.
+           Setting it to False, make things faster (~16%), and require ~2 times
+           less memory (no need to cache each output) but can lead to
+           errors.
```
RamSaw, line 670: I think a little bit misleading, it cannot lead to errors if implemented correctly. Better "but it is more error-prone from coding perspective because you might try to iterate it more than once and get empty output".

pipeline_dp/pipeline_backend.py
```diff
@@ -659,11 +659,28 @@ def singleton(self):
 class LocalBackend(PipelineBackend):
     """Local Pipeline adapter."""
 
+    def __init__(self, output_reiterable: bool = True):
```
RamSaw, line 662: Do we need this flag in general? Should it be set to false by default for backwards compatibility (afaik it is part of public API)? 16% is significant and two times more memory as well, but I think if local is used only for testing then it is still ok.
One big concern: currently LocalBackend is created by the user but we use it internally therefore we will have to handle non-reiterable cases anyway and code with that in mind, so introducing this flag but not enforcing it to True does not help, or?

tests/pipeline_backend_test.py
```diff
@@ -743,6 +744,18 @@ def test_distinct(self):
         output = set(self.backend.distinct(input, "distinct"))
         self.assertSetEqual({1, 2, 3, 4, 5}, output)
 
+    def test_output_reiterable_false(self):
+        backend = pipeline_dp.LocalBackend(output_reiterable=False)
```
RamSaw, line 748: nit: let's separate arrange/act/assert with empty lines

RamSaw (changes requested):

pipeline_dp/pipeline_backend.py
dvadym: Agree. I've dropped the flag and make it Reiterable always

pipeline_dp/pipeline_backend.py
```diff
@@ -774,25 +782,27 @@ def merge_accumulators(accumulators):
...
         def generator():
             for v in set(col):
                 yield v
 
-        return generator()
+        return ReiterableLazyIterable(generator())
 
     def to_list(self, col, stage_name: str):
-        return (list(col) for _ in range(1))
+        return ReiterableLazyIterable((list(col) for _ in range(1)))
```
RamSaw, line 805: does it make sense to do it here? we wrap a tuple that is always reiterable

RamSaw (approved):

Issue #552, opened 2025-05-09
VectorSum: add noise efficiently


RamSaw:
## Description
Currently we provide [additive_vector_noise_params](https://github.com/OpenMined/PipelineDP/blob/6af3163fbea4810e2bbe6e2b7dd1b1ab1a584a62/pipeline_dp/combiners.py#L835) to noise addition function. In these params we divide eps and delta by vector_size. It is not necessary to do, we can keep the total eps and delta per operation.

Then in add_noise_vector we use [_add_random_noise](https://github.com/OpenMined/PipelineDP/blob/6af3163fbea4810e2bbe6e2b7dd1b1ab1a584a62/pipeline_dp/dp_computations.py#L220) function that adds noise given l0 (aka max_partitions_contributed) and l_inf (aka max_contributions_per_partition) sensitivies. However we can add noise more efficiently using [vector_max_norm](https://github.com/OpenMined/PipelineDP/blob/6af3163fbea4810e2bbe6e2b7dd1b1ab1a584a62/pipeline_dp/combiners.py#L181). We can do the same way as it is [done in PipelineDP4j](https://github.com/google/differential-privacy/blob/a7cc26ad91f74756fbe39bab44af6d655b37cc61/pipelinedp4j/main/com/google/privacy/differentialprivacy/pipelinedp4j/core/Combiners.kt#L516), i.e. if vector norm kind is L1 or L2 then we just use it as sensitivity for the whole vector. If norm kind is L_INF, then we multiply it by vector_size if it is Laplace noise or sqrt(vector_size) if it is Gaussian noise.

To summarize current implementation treats each component of the vector as a standalone value, i.e. we kind of perform SUM metrics `vector_size` times independently, but it is not efficient, implementation in PipelineDP4j adds less noise with the same guarantees.

lb034582341:
WIP https://github.com/OpenMined/PipelineDP/pull/559

PR #559, opened 2025-06-10
Fix issue #552: noise computation w.r.t. vector size

lb034582341:
## Description

Problem description in #552 

Currently we provide [additive_vector_noise_params](https://github.com/OpenMined/PipelineDP/blob/6af3163fbea4810e2bbe6e2b7dd1b1ab1a584a62/pipeline_dp/combiners.py#L835) to noise addition function. In these params we divide eps and delta by vector_size. It is not necessary to do, we can keep the total eps and delta per operation.

Then in add_noise_vector we use [_add_random_noise](https://github.com/OpenMined/PipelineDP/blob/6af3163fbea4810e2bbe6e2b7dd1b1ab1a584a62/pipeline_dp/dp_computations.py#L220) function that adds noise given l0 (aka max_partitions_contributed) and l_inf (aka max_contributions_per_partition) sensitivies. However we can add noise more efficiently using [vector_max_norm](https://github.com/OpenMined/PipelineDP/blob/6af3163fbea4810e2bbe6e2b7dd1b1ab1a584a62/pipeline_dp/combiners.py#L181). We can do the same way as it is [done in PipelineDP4j](https://github.com/google/differential-privacy/blob/a7cc26ad91f74756fbe39bab44af6d655b37cc61/pipelinedp4j/main/com/google/privacy/differentialprivacy/pipelinedp4j/core/Combiners.kt#L516), i.e. if vector norm kind is L1 or L2 then we just use it as sensitivity for the whole vector. If norm kind is L_INF, then we multiply it by vector_size if it is Laplace noise or sqrt(vector_size) if it is Gaussian noise.

To summarize current implementation treats each component of the vector as a standalone value, i.e. we kind of perform SUM metrics vector_size times independently, but it is not efficient, implementation in PipelineDP4j adds less noise with the same guarantees.

This PR fixes this issue.

## Affected Dependencies
none

## How has this been tested?
- local tests

## Checklist

### DIFF
--- a/tests/combiners_test.py
+++ b/tests/combiners_test.py
@@ -1052,7 +1052,7 @@ class VectorSumCombinerTest(parameterized.TestCase):
         m = combiner.compute_metrics(np.array([0] * vector_size))
 
         for value in m['vector_sum']:
-            self.assertAlmostEqual(value, 0, delta=1e-4)
+            self.assertAlmostEqual(value, 0, delta=1e-1)
 
     def test_compute_metrics_with_noise(self):
         combiner = self._create_combiner(no_noise=False, vector_size=2)


uv.lock
RamSaw, line 1: should it be reverted? not relevant to this change

RamSaw (changes requested):

RamSaw (changes requested):

PR #560, opened 2025-06-12
Types and other small linter errors

lb034582341:
## Description

Spring-cleanup commit. Initially the goal was to add type annotations. On the way, I also went over small linter errors, including typos. The work is not finished, but this PR is already too big and messy, the follow-ups will be shorter and cleaner.

The goal is to eventually have all types annotated (generics will require python 3.12), and ~no linter error. 

This PR has zero functionality change.

## Affected Dependencies
- none

## How has this been tested?
- make test

## Checklist

### DIFF
--- a/analysis/contribution_bounders.py
+++ b/analysis/contribution_bounders.py
@@ -14,6 +14,8 @@
 """ContributionBounder for utility analysis."""
 
 import numpy as np
+
+import pipeline_dp
 from pipeline_dp import contribution_bounders
 from pipeline_dp import sampling_utils
 from typing import Iterable, Tuple, Union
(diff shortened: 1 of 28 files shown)


RamSaw (changes requested):
Thanks! I reviewed first 14 files, let's discuss before I review the rest. For my better understanding, did you do some changes manually or all of them were done by some automatic tooling? Changes like adding `the` to nouns, converting some variable to list, etc.
I would try to avoid doing any code changes and try to limit the changes to the typing and docstring changes. If we need to do further changes in code, then what about creating a separate PR for more thorough review?

Also, one thing I would like to double check: whether this formatting is aligned with our internal linters. I am pretty sure yes, but just double check. For that it would be useful to know, if you have used any tools for formatting and if yes, then which ones? Sharing exact commands in the PR description would the best.

PR #609, opened 2026-05-13
Speed-up quantiles

dvadym:


pipeline_dp/combiners.py
```diff
@@ -626,6 +615,90 @@ def mechanism_spec(self) -> budget_accounting.MechanismSpec:
...
+        elif self.elements is not None and other.tree is not None:
+            self.create_tree()
+            self.tree.merge(
+                pydp._pydp.bytes_to_summary(other.tree.serialize().to_bytes()))
+        else:
+            self.elements.extend(other.elements)
+            if len(self.elements) >= self.MAX_ELEMENTS_IN_LIST:
+                self.create_tree()
+
+    def __getstate__(self):
```
RamSaw, line 671: why do we need this functions? Should we document it?

RamSaw (approved):

